PR #10, opened 2019-02-11
Expose additional csv parsing options

vishalbollu:
---
Checklist:
- [X] Add license header to each new file
- [X] Test end to end manually (e.g. rebuild registry/operator and run `cx refresh` in an example folder)
- [X] Run automated tests (`./build/test.sh`)
- [X] Update documentation
- [X] Update examples and `cx init`
- [x] Cherry-pick bugfixes into release branches


pkg/workloads/lib/util.py
```diff
@@ -88,6 +88,16 @@ def pluralize(num, singular, plural):
         return str(num) + " " + plural
 
 
+def snake_to_camel(input, sep="_", lower=True):
+    output = ""
+    for idx, word in enumerate(input.split(sep)):
+        if not (idx == 0 and lower):
```
deliahu: It might be clearer if we switch the if:

```
if idx == 0 and lower:
    output += word
else:
    output += word[0].upper() + word[1:]
```

pkg/workloads/lib/util.py
```diff
@@ -88,6 +88,16 @@ def pluralize(num, singular, plural):
...
+            output += word[0].upper() + word[1:]
+        else:
+            output += word
```
deliahu: Do you think it makes sense to `lower()` the first letter in this case? Not sure what behavior you intended, just a thought

PR #12, opened 2019-02-12
Add ability to sample dataset

vishalbollu:
---
Checklist:
- [x] Run automated tests (`./build/test.sh`)
- [x] Update documentation
- [x] Cherry-pick bugfixes into release branches

### DIFF
--- a/pkg/workloads/spark_job/spark_job.py
+++ b/pkg/workloads/spark_job/spark_job.py
@@ -108,25 +108,27 @@ def validate_dataset(ctx, raw_df, cols_to_validate):
         raise UserException("raw column validations failed")
 
 
+def subset_dataset(full_dataset_size, ingest_df, subset_config):
+    max_rows = full_dataset_size
+    if subset_config.get("limit") is not None:
+        max_rows = min(subset_config["limit"], full_dataset_size)
+        fraction = float(subset_config["limit"] / full_dataset_size)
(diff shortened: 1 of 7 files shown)


pkg/workloads/spark_job/spark_job.py
```diff
@@ -108,27 +108,29 @@ def validate_dataset(ctx, raw_df, cols_to_validate):
...
+    if subset_config.get("limit") is not None:
+        max_rows = min(subset_config["limit"], full_dataset_size)
+        fraction = float(subset_config["limit"] / full_dataset_size)
+    elif subset_config.get("fraction") is not None:
+        max_rows = round(full_dataset_size * subset_config["fraction"])
+        fraction = subset_config["fraction"]
+    if max_rows == full_dataset_size:
+        return ingest_df
+    if subset_config["shuffle"]:
+        ingest_df = ingest_df.sample(fraction=fraction, seed=subset_config["seed"])
```
deliahu: As discussed, research if using `fraction=1.0` reduces efficiency. If it does hurt, we should either bump the fraction a little to avoid undersampling, or add the word "roughly" in the docs. If it does work, then the `logger.info` below doesn't need "at most" :)

vishalbollu: Added an extra buffer to sampling but I think we should still have at most to cover randomness

PR #24, opened 2019-02-27
Setup telemetry

vishalbollu:
---
Checklist:
- [x] Run automated tests (`make test`)
- [x] Update documentation


docs/operator/config.md
```diff
@@ -23,6 +23,9 @@ export CORTEX_REGION="us-west-2"
 # The name of the Kubernetes namespace Cortex will use
 export CORTEX_NAMESPACE="cortex"
 
+# Enable usage reporting
+export CORTEX_ENABLE_USAGE_REPORTING="true"
```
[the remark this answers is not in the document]
vishalbollu: makes sense, done

cortex-installer.sh
```diff
@@ -1925,6 +1928,24 @@ function ask_sudo() {
   fi
 }
 
+function prompt_for_usage_stats() {
+  if [ "$CORTEX_ENABLE_TELEMETRY" != "true" ] && [ "$CORTEX_ENABLE_TELEMETRY" != "false" ]; then
+    while true
+    do
+      echo
+      read -p "Would you like to help improve Cortex by anonymously sending errors and usage stats to the dev team [Y/n] " -n 1 -r
```
deliahu: Do you know why the convention is for "Y" to be capital and "n" lower case in the prompt? We accept upper or lower case, and don't have a default, so if it's one of those reasons maybe it should be "[y/n]". We can also update them in a separate commit.

cortex-installer.sh
```diff
@@ -1925,6 +1928,24 @@ function ask_sudo() {
...
+      echo
+      if [[ $REPLY =~ ^[Yy]$ ]]; then
+        export CORTEX_ENABLE_TELEMETRY=true
+        break
+      elif [[ $REPLY =~ ^[Nn]$ ]]; then
+        export CORTEX_ENABLE_TELEMETRY=false
+        break
+      fi
```
deliahu: Maybe print `Please enter "y" or "n"` before showing the prompt again?

docs/operator/config.md
```diff
@@ -23,6 +23,9 @@ export CORTEX_REGION="us-west-2"
 # The name of the Kubernetes namespace Cortex will use
 export CORTEX_NAMESPACE="cortex"
 
+# Flag to enable collecting errors and usage stats. If flag is not set to either true or false, user will be prompted.
```
deliahu: I think double quotes around "true" and "false" would make it even more clear

pkg/operator/errors/errors.go
```diff
@@ -0,0 +1,39 @@
...
+
+func Exit(items ...interface{}) {
+	err := errors.ConsolidateErrItems(items...)
+	telemetry.ReportErrorBlocking(err)
+	errors.Exit(err)
+}
+
+func Panic(items ...interface{}) {
+	err := errors.ConsolidateErrItems(items...)
+	telemetry.ReportError(err)
```
deliahu: Is this non-blocking because the `panic` will be caught? If so, we just need to remember to not use `Panic()` in `init()`, just `Exit()`

pkg/operator/operator.go
```diff
@@ -122,32 +125,31 @@ func startCron() {
 
 func runCron() {
 	defer errors.Recover("cron failed")
```
deliahu: If a panic happens, will the error be sent to telemetry? Do we need to have a recover in the operator package, so we can send the error? If so, maybe we don't need `operator.Panic()` since otherwise the error would be reported twice. Same question applies to the recover middleware in the http server.

cortex-installer.sh
```diff
@@ -1933,15 +1933,16 @@ function prompt_for_usage_stats() {
     while true
     do
       echo
-      read -p "Would you like to help improve Cortex by anonymously sending errors and usage stats to the dev team [Y/n] " -n 1 -r
+      read -p "Would you like to help improve Cortex by anonymously sending errors and usage stats to the dev team [y/n] " -n 1 -r
       echo
-      if [[ $REPLY =~ ^[Yy]$ ]]; then
```
deliahu: We might as well still accept capital letters, since it's safe enough to assume "Y" means yes and "N" means no

pkg/operator/errors/errors.go
```diff
@@ -0,0 +1,38 @@
...
+func ReportAndExit(items ...interface{}) {
+	err := errors.ConsolidateErrItems(items...)
+	telemetry.ReportErrorBlocking(err)
+	errors.Exit(err)
+}
+
+func ReportAndRecover(strs ...string) error {
+	if errInterface := recover(); errInterface != nil {
+		err := errors.CastRecoverError(errInterface, strs...)
+		telemetry.ReportErrorBlocking(err)
```
deliahu: Can this be asynchronous?

pkg/operator/telemetry/telemetry.go
```diff
@@ -0,0 +1,128 @@
...
+		Timestamp:  time.Now(),
+		Version:    consts.CortexVersion,
+		OperatorID: operatorID,
+		Event:      name,
+	}
+
+	byteArray, _ := json.Marshal(usageEvent)
+	resp, err := httpClient.Post(telemetryURL+eventPath, "application/json", bytes.NewReader(byteArray))
+	if err != nil {
+		fmt.Println(err.Error())
```
deliahu: Should we `errors.PrintError(err)` instead?

pkg/operator/telemetry/telemetry.go
```diff
@@ -0,0 +1,128 @@
...
+		Timestamp:  time.Now(),
+		Version:    consts.CortexVersion,
+		OperatorID: operatorID,
+		Title:      err.Error(),
+		Stacktrace: fmt.Sprintf("%+v", err),
+	}
+	byteArray, _ := json.Marshal(errorEvent)
+	resp, err := httpClient.Post(telemetryURL+errorPath, "application/json", bytes.NewReader(byteArray))
+	if err != nil {
+		fmt.Println(err.Error())
```
deliahu: Should we `errors.PrintError(err)` instead?

pkg/operator/endpoints/shared.go
```diff
@@ -57,6 +59,7 @@ func RespondIfError(w http.ResponseWriter, err error, strs ...string) bool {
 func RecoverAndRespond(w http.ResponseWriter, strs ...string) {
 	if errInterface := recover(); errInterface != nil {
 		err := errors.CastRecoverError(errInterface, strs...)
+		telemetry.ReportError(err)
```
deliahu: How do you feel about `errors.PrintError(err)` here, and removing the print from `RespondErrorCode()`? The concern is that we do return (instead of panic on) some errors that should be printed (even though most of the errors we return are user errors and shouldn't be printed), so maybe we should make that change when we revisit our internal/user error handling strategy.

cortex-installer.sh
```diff
@@ -1925,6 +1928,25 @@ function ask_sudo() {
   fi
 }
 
+function prompt_for_usage_stats() {
+  if [ "$CORTEX_ENABLE_TELEMETRY" != "true" ] && [ "$CORTEX_ENABLE_TELEMETRY" != "false" ]; then
+    while true
+    do
+      echo
+      read -p "Would you like to help improve Cortex by anonymously sending error reports and usage stats to the dev team [y/n] " -n 1 -r
```
deliahu: I just checked ubuntu, and they use "[Y/n]" and accept upper or lower case, so let's just do that throughout (I think that is already the behavior in the rest of `cortex-installer.sh`

pkg/operator/errors/errors.go
vishalbollu: Okay will make it async and add Cron to the function title

docs/operator/config.md
```diff
@@ -23,6 +23,9 @@ export CORTEX_REGION="us-west-2"
 # The name of the Kubernetes namespace Cortex will use
 export CORTEX_NAMESPACE="cortex"
 
+# Flag to enable collecting error reports and usage stats. If flag is not set to either "true" or "false", user will be prompted.
```
1vn: is the user reading this? if so may it should be "you will be prompted"

pkg/lib/errors/errors.go
```diff
@@ -97,11 +97,7 @@ func FirstError(errs ...error) error {
 	return nil
 }
 
-func Exit(items ...interface{}) {
-	if len(items) == 0 {
-		items = append(items, "empty exit")
-	}
-
+func ConsolidateErrItems(items ...interface{}) error {
```
1vn: `Merge`?

pkg/operator/errors/errors.go
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+package errors
+
+import (
+	"github.com/cortexlabs/cortex/pkg/lib/errors"
+	"github.com/cortexlabs/cortex/pkg/operator/telemetry"
+)
+
+func ReportAndExit(items ...interface{}) {
```
1vn: i don't feel like this package is necessary.  From what I am reading, `ReportAndExit` is adding a layer of abstraction on top of e.g.
```
telemetry.ReportErrorBlocking(errors.Wrap(err, "kubernetes"))
errors.Exit(err)
```
IMO, `CronReportAndRecover` is specific enough to be inlined in operator.go and unexported

pkg/operator/telemetry/telemetry.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/cortexlabs/cortex/pkg/lib/errors"
+	"github.com/cortexlabs/cortex/pkg/operator/aws"
+	cc "github.com/cortexlabs/cortex/pkg/operator/cortexconfig"
+)
+
+const defaultTelemetryURL = "https://telemetry.cortexlabs.com"
+const eventPath = "/events"
+const errorPath = "/errors"
+
+var telemetryURL string
```
1vn: I think this and `httpClient` should be fields in a `type Client struct`package variable you export

pkg/operator/telemetry/telemetry.go
```diff
@@ -0,0 +1,129 @@
...
+var httpClient http.Client
+
+func init() {
```
1vn: IMO we should move away from this init() pattern and make this an explicit exported `Init`function called by operator

PR #30, opened 2019-03-02
Prevent users from installing conflicting packages

vishalbollu:
---
Checklist:
- [x] Run automated tests (`make test`)
- [x] Update documentation


pkg/workloads/lib/package.py
```diff
@@ -35,6 +36,20 @@ def get_build_order(python_packages):
...
 
 
+def get_restricted_packages():
+    cortex_packages = {"pyspark": "2.4.0", "tensorflow": "1.12.0"}
+    req_files = glob.glob("/src/**/requirements.txt", recursive=True)
+    for req_file in req_files:
+        with open(req_file) as f:
+            lines = [line.strip() for line in f.readlines()]
+            for line in lines:
+                print(line)
```
deliahu: Did you intend to leave this print?

pkg/workloads/tf_api/requirements.txt
```diff
@@ -2,15 +2,3 @@ flask==1.0.2
 Flask-API==1.1
```
deliahu: Do you think we should be explicit and call this file "requirements.cortex.txt" or something similar?

deliahu: Do you think we should add a comment, saying that you must use `name==version` syntax with nothing else on that line (e.g. no in-line comments, no ` --hash=sha256:...`, no `>=`, `<=`, `<`, etc?

pkg/workloads/lib/package.py
deliahu: We might as well support whole-line comments, e.g. continue if `line.startswith("#")`

pkg/workloads/lib/package.py
```diff
@@ -50,17 +65,31 @@ def build_packages(python_packages, bucket):
...
-                os.path.join(WHEELHOUSE_PATH, package_name), requirement
-            ).split()
+            "pip3 wheel -w {} {}".format(package_wheel_path, requirement).split()
         )
         if completed_process.returncode != 0:
             raise UserException("creating wheels", package_name)
 
+        for wheelname in os.listdir(package_wheel_path):
+            name_split = wheelname.split("-")
+            dist_name = name_split[0]
```
deliahu: nit: this might be a little cleaner if it's `dist_name, version = name_split[0], name_split[1]`, or if there is only one `-`, you can do `dist_name, version = wheelname.split("-")`

pkg/workloads/lib/package.py
```diff
@@ -94,6 +123,10 @@ def build(args):
     try:
         build_packages(python_packages, ctx.bucket)
         util.log_job_finished(ctx.workload_id)
+    except CortexException as e:
+        e.wrap("error")
+        logger.exception(e)
+        ctx.upload_resource_status_failed(*python_packages_list)
     except Exception as e:
         logger.exception(e)
```
deliahu: is there a reason to not `e.wrap("error")` in this case and consolidate the error handling? Also do the other workload top-level python files need a similar change?

pkg/workloads/tf_train/requirements.oob.txt
```diff
@@ -0,0 +1,11 @@
+numpy==1.15.4
```
deliahu: Do you think it would make sense to have a single shared `requirements.oob.txt` file? Are there differences in what we provide OOB in the different workloads?

pkg/workloads/tf_train/requirements.oob.txt
```diff
@@ -0,0 +1,11 @@
...
+pandas==0.23.4
+scipy==1.2.0
+sympy==1.3
+statsmodels==0.9.0
+python-dateutil==2.7.5
+six==1.11.0
+wrapt==1.11.0
+requests==2.21.0
```
deliahu: These bottom three look strange, are they required by Cortex but listed here so the user can change the versions? Maybe we should add a comment here explaining

pkg/workloads/tf_api/requirements.txt
vishalbollu: will do

PR #40, opened 2019-03-22
OOM (Out of memory) status

1vn:
---
Checklist:
- [x] Update documentation

### DIFF
--- a/pkg/api/resource/status.go
+++ b/pkg/api/resource/status.go
@@ -110,7 +110,7 @@ const (
 	StatusAPIGroupParentKilled
 	StatusAPIGroupUpdateSkipped
 
-	StatusDataOOM
+	StatusDataKilledOOM
 )
 
 var statusCodes = []string{
@@ -142,7 +142,7 @@ var statusCodes = []string{
 	"status_data_oom",
 }
 
-var _ = [1]int{}[int(StatusDataOOM)-(len(statusCodes)-1)] // Ensure list length matches
+var _ = [1]int{}[int(StatusDataKilledOOM)-(len(statusCodes)-1)] // Ensure list length matches
 
 var statusCodeMessages = []string{
 	"unknown", // StatusUnknown
@@ -170,10 +170,10 @@ var statusCodeMessages = []string{
 	"upstream termination", // StatusAPIGroupParentKilled
 	"update skipped",       // StatusAPIGroupUpdateSkipped
 
-	"killed (oom)", // StatusDataOOM
+	"terminated (out of mem)", // StatusDataOOM
 }
 
-var _ = [1]int{}[int(StatusDataOOM)-(len(statusCodeMessages)-1)] // Ensure list length matches
+var _ = [1]int{}[int(StatusDataKilledOOM)-(len(statusCodeMessages)-1)] // Ensure list length matches
 
 // StatusDataRunning aliases
 const (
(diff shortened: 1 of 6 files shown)


pkg/api/resource/status.go
```diff
@@ -109,6 +109,8 @@ const (
 	StatusAPIGroupParentFailed
 	StatusAPIGroupParentKilled
 	StatusAPIGroupUpdateSkipped
+
+	StatusDataOOM
```
deliahu: This should be up with the other data statuses

pkg/api/resource/status.go
```diff
@@ -165,9 +169,11 @@ var statusCodeMessages = []string{
 	"upstream error",       // StatusAPIGroupParentFailed
 	"upstream termination", // StatusAPIGroupParentKilled
 	"update skipped",       // StatusAPIGroupUpdateSkipped
+
+	"killed (oom)", // StatusDataOOM
```
deliahu: How about “terminated (out of memory)”? Does that fit on the cx get line without widening the table? If it’s too long, maybe “terminated (OOM)”, but I worry some people might not know what OOM is

pkg/api/resource/status.go
deliahu: Maybe call it DataStatusKilledOOM to make it clear that it is a “killed” state?

pkg/operator/k8s/pod.go
```diff
@@ -40,6 +40,7 @@ const (
 	PodStatusFailed      = "Failed"
 	PodStatusKilled      = "Killed"
 	PodStatusUnknown     = "Unknown"
+	PodStatusOOM         = "Out of Memory"
```
deliahu: PodStatusKilledOOM?

pkg/operator/k8s/pod.go
```diff
@@ -40,6 +40,7 @@ const (
 	PodStatusFailed      = "Failed"
 	PodStatusKilled      = "Killed"
 	PodStatusUnknown     = "Unknown"
```
deliahu: I do you think Unknown should be last in the list?

pkg/operator/k8s/pod.go
```diff
@@ -108,11 +109,17 @@ func GetPodStatus(pod *corev1.Pod) string {
 		for _, containerStatus := range pod.Status.ContainerStatuses {
 			if containerStatus.LastTerminationState.Terminated != nil {
 				exitCode := containerStatus.LastTerminationState.Terminated.ExitCode
+				if exitCode == 137 {
```
deliahu: Do you think we want these two blocks in the “running” case too, since it can return “killed”?

pkg/api/resource/status.go
deliahu: Should we add StatusDataParentKilledOOM and StatusAPIGroupParentKilledOOM? Another option is to reuse StatusDataParentKilled and StatusAPIGroupParentKilled if you don’t think it’s necessary to differentiate these two on the downstream resources.

1vn: Best practice with enums is appending to the end only, to not break previous states

pkg/operator/k8s/pod.go
1vn: I think it could be convenient if we refactor this into enums

pkg/operator/k8s/pod.go
1vn: I'm not sure if it would be an OOM case, haven't been able to reproduce to know for sure.

pkg/operator/workloads/api_status.go
```diff
@@ -194,7 +194,7 @@ func updateAPIStatusCodeByParents(apiStatus *resource.APIStatus, dataStatuses ma
 	parentSkipped := false
 	for dependency := range allDependencies {
 		switch dataStatuses[dependency].Code {
-		case resource.StatusDataKilled:
+		case resource.StatusDataKilled, resource.StatusDataKilledOOM:
```
deliahu: This change should also be made in `updateDataStatusCodeByParents()`

PR #70, opened 2019-04-19
Refactor aws/k8s to lib, api to operator, cortex_config to config 

1vn:
---
Checklist:
- [x] Update documentation


pkg/lib/telemetry/telemetry.go
```diff
@@ -25,38 +25,35 @@ import (
...
+	EnableTelemetry bool
+	OperatorID      string
 }
 
-func getDefaultClient() *Client {
-	if defaultClient != nil {
-		return defaultClient
-	}
-
+func New(telemetryURL, hashedAccountID string) *Client {
```
deliahu: `New(telemetryURL string, hashedAccountID string)`?

pkg/operator/argo/argo.go
```diff
@@ -50,9 +50,9 @@ var runningStates = []string{
 	string(awfv1.NodeRunning),
 }
 
-func init() {
-	wfcs := wfclientset.NewForConfigOrDie(k8s.Config)
-	workflowClient = wfcs.ArgoprojV1alpha1().Workflows(cc.Namespace)
+func Init() {
```
deliahu: Is there anything Cortex-specific about this package? What do you think about passing in `namespace` and `restConfig` in `Init()` (is that all that's needed?), using it to populate a client, returning the client, and using the client downstream as a public variable in `config`, like with do with k8s and aws?

Now that I typed it out it sounds like it might be overkill, but I'll leave the comment here in case you want to consider it.

A lighter-weight option that still might be an improvement could be to pass in `namespace` and `restConfig` and save it in a package variable, and don't import the `config` package. The advantage would be a more explicit list of what's required to init argo, rather than relying on `config` to already be populated correctly. The disadvantage would be if we want changes in `config` to automatically populate to `argo`, which I could see. This would be less of an issue if we used a client as a global variable in `config`, since we would have just one place to update it.

The same discussion can be had regarding the `spark` package (but maybe not since `spark` probably has more dependencies on cortex-specific things).

What do you think? We can discuss it live.

pkg/operator/config/config.go
```diff
@@ -0,0 +1,77 @@
...
+		Namespace:           env.GetStr("NAMESPACE"),
+		OperatorImage:       env.GetStr("IMAGE_OPERATOR"),
+		SparkImage:          env.GetStr("IMAGE_SPARK"),
+		TFTrainImage:        env.GetStr("IMAGE_TF_TRAIN"),
+		TFServeImage:        env.GetStr("IMAGE_TF_SERVE"),
+		TFAPIImage:          env.GetStr("IMAGE_TF_API"),
+		PythonPackagerImage: env.GetStr("IMAGE_PYTHON_PACKAGER"),
+		TFTrainImageGPU:     env.GetStr("IMAGE_TF_TRAIN_GPU"),
+		TFServeImageGPU:     env.GetStr("IMAGE_TF_SERVE_GPU"),
+		TelemetryURL:        configreader.MustStringFromEnv("CONST_TELEMETRY_URL", &configreader.StringValidation{Required: false, Default: consts.TelemetryURL}),
```
deliahu: nit: `required: false` is implied. But it might still be worth leaving for clarity, up to you.

pkg/operator/config/config.go
```diff
@@ -0,0 +1,77 @@
...
+	TFAPIImage          string `json:"tf_api_image"`
+	PythonPackagerImage string `json:"python_packager_image"`
+	TFTrainImageGPU     string `json:"tf_train_image_gpu"`
+	TFServeImageGPU     string `json:"tf_serve_image_gpu"`
+	TelemetryURL        string `json:"telemetry_url"`
+	EnableTelemetry     bool   `json:"enable_telemetry"`
+	OperatorInCluster   bool   `json:"operator_in_cluster"`
+}
+
+func NewFromEnv() *CortexConfig {
```
deliahu: `NewFromEnv()` should populate `Cortex.EnableTelemetry` too, which will be used when initializing telemetry

pkg/operator/operator.go
```diff
@@ -44,10 +47,41 @@ const (
...
-var markedWorkflows = strset.New()
+var (
+	awsClient       *aws.Client
+	telemtryClient  *telemetry.Client
+	markedWorkflows = strset.New()
+)
 
 func main() {
-	telemetry.ReportEvent("operator.init")
+	conf := config.NewFromEnv()
```
deliahu: Maybe it should be called `config.Init()`, and not return anything, since in the rest of the codebase we use the `config` package directly?

pkg/operator/operator.go
```diff
@@ -44,10 +47,41 @@ const (
...
+	config.AWS = aws.New(conf.Region, conf.Bucket)
```
deliahu: What do you think about initializing `AWS`, `Telemetry`, and `Kubernetes` in `config.Init()` as well?

deliahu: Actually, now I'm wondering if we should use the lower-case `init()` just for `config`, to ensure it's always properly initialized? Also maybe we should use it for `argo` and `spark` since those are singletons which downstream packages (like `workflows`) rely on being initialized? we can discuss live

pkg/operator/argo/argo.go
```diff
@@ -50,7 +50,7 @@ var runningStates = []string{
 	string(awfv1.NodeRunning),
 }
 
-func Init() {
+func Init(restConfig *rest.Config, namespace string) {
```
deliahu: These args don't seem to be used. Let's discuss what we want to do here

pkg/operator/config/config.go
```diff
@@ -53,25 +58,65 @@ type CortexConfig struct {
...
+
+	if err := spark.Init(); err != nil {
+		return err
+	}
+
+	if err := workloads.Init(); err != nil {
+		return err
+	}
+
 	return Cortex
```
deliahu: Should it be `return nil`? Did it compile like this?

pkg/operator/operator.go
```diff
@@ -51,7 +54,30 @@ var (
 )
 
 func main() {
-	if err := config.Init(); err != nil {
+	config.Init()
+
+	config.AWS = aws.New(config.Cortex.Region, config.Cortex.Bucket)
```
deliahu: I think we could move lines 59 - 68 in `config.Init()`. `context.Init()` can still be here, but it would be nice for `config.Init()` to be self contained

pkg/operator/workloads/data_job.go
```diff
@@ -51,11 +55,122 @@ func dataJobSpec(
 	if shouldIngest {
 		args = append(args, "--ingest")
 	}
-	spec := spark.Spec(workloadID, ctx, workloadTypeData, sparkCompute, args...)
+	spec := Spec(workloadID, ctx, workloadTypeData, sparkCompute, args...)
 	argo.EnableGC(spec)
 	return spec
 }
 
+func Spec(workloadID string, ctx *context.Context, workloadType string, sparkCompute *userconfig.SparkCompute, args ...string) *sparkop.SparkApplication {
```
deliahu: nit: maybe this should be called `sparkSpec()` since it's in the `workloads` package now, and can be private?

PR #89, opened 2019-04-29
Refactor away api/strings

### LINKED ISSUE #62: Split strings package
### Description

Strings package shouldn't have anything cortex-specific, and should be moved to lib

`pkg/lib/interfaces/interfaces.go` would use `PrimType` from whichever package it moves too and would use `ErrorInvalidPrimitiveType` from that package, and we could delete `pkg/lib/interfaces/errors.go`

1vn:
Closes #62 

lib/math -> lib/slices: functions here were more slice-like and less "math"
lib/interfaces -> cr: interfaces was too dependent on config usecase using the prim type constants api/strings …-> lib/strings

---
Checklist:
- [ ] Test end to end manually (e.g. build/push all images, restart local operator, and run `cx refresh` in an example folder)
- [ ] Update documentation
- [ ] Update examples and `cx init`


pkg/lib/configreader/interface_test.go
```diff
@@ -14,15 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package interfaces_test
+package configreader
```
deliahu: We currently have the convention of using the external package for tests, but I'm on board for switching to internal packages going forward since it simplifies some things, and [Dave Cheney is ok with it](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_prefer_internal_tests_to_external_tests) :)

pkg/lib/configreader/types.go
```diff
@@ -16,6 +16,77 @@ limitations under the License.
...
+var (
+	typeStrRegex         = regexp.MustCompile(`"(INT|FLOAT|STRING|BOOL)(_COLUMN)?(\|(INT|FLOAT|STRING|BOOL)(_COLUMN)?)*"`)
+	singleDataTypeRegexp = regexp.MustCompile(`^\w*\w$`)
+)
+
+func EnvVar(envVarName string) string {
+	return fmt.Sprintf("environment variable \"%s\"", envVarName)
+}
+
+func DataTypeStrsOr(dataTypes []interface{}) string {
```
deliahu: Would it be easy to move the `DataType*` functions to `userconfig`? I'd rather keep Cortex related code out of `configreader` if possible

pkg/operator/api/userconfig/transformed_columns.go
```diff
@@ -19,8 +19,7 @@ package userconfig
 import (
 	"sort"
 
-	cr "github.com/cortexlabs/cortex/pkg/lib/configreader"
```
deliahu: There are probably a lot of other instances of `cr "github.com/cortexlabs/cortex/pkg/lib/configreader"`, do you want to switch them all? It can be a separate PR after this one

pkg/lib/debug/debug.go
```diff
@@ -45,9 +45,9 @@ func Sppg(obj interface{}) string {
 }
 
 func Ppj(obj interface{}) {
-	str, err := json.MarshalJSONStr(obj)
+	b, err := json.Marshal(obj)
```
deliahu: Can you use `json.MarshalIndent(obj, "", "  ")`?

PR #93, opened 2019-05-01
Refactor resource status

### LINKED ISSUE #75: Merge resource status types
### Description

`Status*` consts in `resource/status.go` should be shared across resource types

1vn:
fixes #75 

---
Checklist:
- [x] Update documentation


cli/cmd/predict.go
```diff
@@ -103,7 +102,7 @@ var predictCmd = &cobra.Command{
 			for _, prediction := range predictResponse.ClassificationPredictions {
 				if prediction.PredictedClassReversed != nil {
 					json, _ := json.Marshal(prediction.PredictedClassReversed)
-					fmt.Println(s.TrimPrefixAndSuffix(string(json), "\""))
+					fmt.Println(libstrings.TrimPrefixAndSuffix(string(json), "\""))
```
deliahu: nit: Can you use `s.*` for now since that's what we use elsewhere? We can discuss if we want to change to `libstrings`, and if so, we should change them all at once.

pkg/operator/api/resource/status.go
```diff
@@ -92,25 +92,23 @@ const (
...
 	// Additional API group statuses (i.e. aggregated API status)
-	StatusAPIGroupPendingUpdate
-	StatusAPIGroupParentFailed
-	StatusAPIGroupParentKilled
-	StatusAPIGroupUpdateSkipped
+	StatusPendingUpdate
+	StatusUpdateSkipped
 
-	StatusDataKilledOOM
+	StatusKilledOOM
```
deliahu: If we're keeping the comments (e.g. `// Data statuses`), then we should move this to the `// Shared statuses` section. If you don't want to change the order, we can remove the comments or add an inline comment for this one (e.g. `// Shared status`)

PR #100, opened 2019-05-07
Make schema validation optional

1vn:
---
Checklist:
- [x] Update documentation


pkg/lib/strings/operations.go
```diff
@@ -128,3 +128,7 @@ func StrsSentence(strs []string, lastJoinWord string) string {
 		return strings.Join(strs[:lastIndex], ", ") + ", " + lastJoinWord + " " + strs[lastIndex]
 	}
 }
+
+func PathToName(path string) string {
```
deliahu: Safest might be to use the hash of the implementation file, what do you think?

pkg/operator/context/aggregators.go
```diff
@@ -23,31 +23,55 @@ import (
...
 			return nil, err
 		}
 		userAggregators[aggregator.Name] = aggregator
 	}
 
+	for _, aggregateConfig := range config.Aggregates {
+		if aggregateConfig.AggregatorPath == nil {
+			continue
+		}
+
```
deliahu: I think we can also short circuit here if `userAggregators[*aggregateConfig.AggregatorPath] != nil`. Same for transformers.

pkg/operator/context/aggregators.go
```diff
@@ -23,31 +23,55 @@ import (
...
+			},
+			Path: *aggregateConfig.AggregatorPath,
+		}
+		aggregator, err := newAggregator(*anonAggregatorConfig, impl, nil, pythonPackages, true)
+		if err != nil {
+			return nil, err
+		}
+		userAggregators[*aggregateConfig.AggregatorPath] = aggregator
+	}
+
```
deliahu: If we set the `Aggregator` field of the `aggregate` to the autogenerated name, and used `userAggregators[s.PathToName(*aggregateConfig.AggregatorPath)]` instead of `userAggregators[*aggregateConfig.AggregatorPath]` would it make things easier down the line? E.g. I think we could remove the logic in `autogenerator.go` and lines 162-171 in this file. (same applies for transformers)

pkg/workloads/spark_job/spark_util.py
```diff
@@ -460,6 +487,37 @@ def validate_transformer(column_name, df, ctx, spark):
     trans_impl, _ = ctx.get_transformer_impl(column_name)
 
     if hasattr(trans_impl, "transform_python"):
+        sample_df = df.collect()
+        sample = sample_df[0]
+        inputs = ctx.create_column_inputs_map(sample, column_name)
+        _, impl_args = extract_inputs(column_name, ctx)
+        transformedSample = trans_impl.transform_python(inputs, impl_args)
+        rowType = type(transformedSample)
+        isList = rowType == list
```
deliahu: Is `isList = (rowType == list)` valid syntax (just for clarity)?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -460,6 +487,37 @@ def validate_transformer(column_name, df, ctx, spark):
...
+
+        for row in sample_df:
+            inputs = ctx.create_column_inputs_map(row, column_name)
+            transformedSample = trans_impl.transform_python(inputs, impl_args)
+            if rowType != type(transformedSample):
```
deliahu: Will this catch thee case if they are both lists of different element types?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -460,6 +487,37 @@ def validate_transformer(column_name, df, ctx, spark):
     trans_impl, _ = ctx.get_transformer_impl(column_name)
 
     if hasattr(trans_impl, "transform_python"):
+        sample_df = df.collect()
```
deliahu: @vishalbollu should review this section

pkg/workloads/spark_job/spark_util.py
```diff
@@ -460,6 +487,37 @@ def validate_transformer(column_name, df, ctx, spark):
...
+            )
+
+
+        typeConversionDict = PYTHON_TYPE_TO_CORTEX_TYPE
+        if isList:
+            rowType = type(transformedSample[0])
+            typeConversionDict = PYTHON_TYPE_TO_CORTEX_LIST_TYPE
+
+        # for downstream operations on this job
+        ctx.transformed_columns[column_name]["type"] = typeConversionDict[rowType]
```
deliahu: Here we set the field in the context, which is inconsistent with how we access these inferred types in other cases. It might be a good "rule" to never modify the context. We can cache the metadata in memory when calling `update_metadata()` (see my comment about this in `context.py`), so later in this job when we call `ctx.get_inferred_colum_type(column_name)` it will already have it cached. Would that work? Which downstream operations use `transformed_columns[column_name]["type"]`, and can we replace calls to it with `ctx.get_inferred_colum_type(column_name)`? I guess this means we should never access `transformed_columns[column_name]["type"]` in Python, since it might be `unknown`.

pkg/workloads/spark_job/spark_util.py
deliahu: Can we only run this type inference code if the user didn't provide the type?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -520,12 +582,13 @@ def validate_transformer(column_name, df, ctx, spark):
...
 
             # perform the necessary upcast/downcast for the column e.g INT -> LONG or DOUBLE -> FLOAT
-            transform_spark_df = transform_spark_df.withColumn(
-                column_name,
-                F.col(column_name).cast(
-                    CORTEX_TYPE_TO_SPARK_TYPE[ctx.transformed_columns[column_name]["type"]]
-                ),
-            )
+            if not skip_validation:
+                transform_spark_df = transform_spark_df.withColumn(
```
deliahu: Why do we not need to run this cast if we're skipping validation?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -520,12 +582,13 @@ def validate_transformer(column_name, df, ctx, spark):
...
+                    column_name,
+                    F.col(column_name).cast(
+                        CORTEX_TYPE_TO_SPARK_TYPE[ctx.transformed_columns[column_name]["type"]]
```
deliahu: `ctx.transformed_columns[column_name]["type"]` should probably not be used, and instead the helper should be used (see comment above)

pkg/workloads/spark_job/spark_util.py
```diff
@@ -579,6 +642,18 @@ def transform_column(column_name, df, ctx, spark):
...
+        if skip_validation:
+            df = execute_transform_spark(column_name, df, ctx, spark)
+            column_type = df.select(column_name).schema[0].dataType
+            # for downstream operations on other jobs
+            ctx.update_metadata(
+                {"type": SPARK_TYPE_TO_CORTEX_TYPE[column_type]}, "transformed_columns", column_name
+            )
+            return df.withColumn(column_name, F.col(column_name).cast(column_type))
+
         return execute_transform_spark(column_name, df, ctx, spark).withColumn(
```
deliahu: Why shouldn't we cast if `skip_validation` is `true`?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -462,65 +465,63 @@ def _transform(*values):
...
-                "type inference failed, mixed data types in dataframe.",
-            )
-
-
-        typeConversionDict = PYTHON_TYPE_TO_CORTEX_TYPE
-        if isList:
-            rowType = type(transformedSample[0])
-            typeConversionDict = PYTHON_TYPE_TO_CORTEX_LIST_TYPE
+            rowType = type(transformedSample)
+            isList = rowType == list
```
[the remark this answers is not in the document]
deliahu: Never mind, I now see that Black auto-formats it away

pkg/workloads/spark_job/spark_util.py
```diff
@@ -435,34 +465,63 @@ def _transform(*values):
...
 
     trans_impl, _ = ctx.get_transformer_impl(column_name)
 
     if hasattr(trans_impl, "transform_python"):
+        if transformed_column.get("transformer_path", None):
+            sample_df = test_df.collect()
+            sample = sample_df[0]
+            inputs = ctx.create_column_inputs_map(sample, column_name)
+            _, impl_args = extract_inputs(column_name, ctx)
+            transformedSample = trans_impl.transform_python(inputs, impl_args)
```
vishalbollu: camel case in python

pkg/workloads/spark_job/spark_util.py
```diff
@@ -435,34 +465,63 @@ def _transform(*values):
...
+                inputs = ctx.create_column_inputs_map(row, column_name)
+                transformedSample = trans_impl.transform_python(inputs, impl_args)
+                if rowType != type(transformedSample):
+                    raise UserRuntimeException(
+                        "transformed column " + column_name,
+                        "type inference failed, mixed data types in dataframe.",
+                    )
+
+            typeConversionDict = PYTHON_TYPE_TO_CORTEX_TYPE
+            if isList:
```
vishalbollu: I would refactor out the code that infers the cortex type of the python object to a function. That way we can thoroughly unit test that function

examples/fraud/resources/weight_column.yaml
```diff
@@ -5,9 +5,18 @@
     columns:
       col: class
 
+- kind: transformer
```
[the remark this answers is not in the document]
1vn: testing, should keep some examples of A's

pkg/workloads/lib/tf_lib.py
```diff
@@ -69,7 +69,8 @@ def get_column_tf_types(model_name, ctx, training=True):
         ]
 
         for column_name in model["training_columns"]:
-            column_types[column_name] = CORTEX_TYPE_TO_TF_TYPE[ctx.columns[column_name]["type"]]
+            columnType = ctx.get_inferred_column_type(column_name)
```
vishalbollu: snake_case

pkg/workloads/spark_job/spark_util.py
```diff
@@ -435,34 +465,67 @@ def _transform(*values):
...
-    column_data_type_str = ctx.transformed_columns[column_name]["type"]
+    column_data_type_str = ctx.get_inferred_column_type(column_name)
     transform_udf = F.udf(transform_python_func, CORTEX_TYPE_TO_SPARK_TYPE[column_data_type_str])
     return df.withColumn(column_name, transform_udf(*required_columns_sorted))
 
 
-def validate_transformer(column_name, df, ctx, spark):
+def infer_type(obj):
+    objType = type(obj)
+    typeConversionDict = PYTHON_TYPE_TO_CORTEX_TYPE
```
vishalbollu: snake_case

pkg/workloads/lib/context.py
```diff
@@ -469,6 +470,56 @@ def upload_resource_status_end(self, exit_code, *resources):
...
+        if use_cache and self.ctx[context_key][context_item].get("metadata", None):
+            return self.ctx[context_key][context_item]["metadata"]
+
+        metadata_uri = self.ctx[context_key][context_item]["metadata_key"]
+        metadata = self.storage.get_json(metadata_uri, allow_missing=True)
+        self.ctx[context_key][context_item]["metadata"] = metadata
+        return metadata
+
+    def get_inferred_column_type(self, column_name):
+        columnType = self.columns[column_name].get("type", "unknown")
```
deliahu: camel case

pkg/operator/api/userconfig/config.go
```diff
@@ -200,19 +200,40 @@ func (config *Config) Validate(envName string) error {
...
+	// Check local aggregators exist or a path to one is defined
 	aggregatorNames := config.Aggregators.Names()
 	for _, aggregate := range config.Aggregates {
-		if !strings.Contains(aggregate.Aggregator, ".") && !slices.HasString(aggregatorNames, aggregate.Aggregator) {
+		if aggregate.AggregatorPath == nil && aggregate.Aggregator == "" {
+			return errors.Wrap(ErrorSpecifyOnlyOneMissing("aggregator", "aggregator_path"), Identify(aggregate))
+		}
+
+		if aggregate.AggregatorPath != nil && aggregate.Aggregator != "" {
+			return ErrorMultipleAggregatorSpecified(aggregate)
```
deliahu: Can you use `ErrorSpecifyOnlyOne` here?

pkg/operator/api/userconfig/config.go
```diff
@@ -200,19 +200,40 @@ func (config *Config) Validate(envName string) error {
...
+	// Check local transformers exist or a path to one is defined
 	transformerNames := config.Transformers.Names()
 	for _, transformedColumn := range config.TransformedColumns {
-		if !strings.Contains(transformedColumn.Transformer, ".") && !slices.HasString(transformerNames, transformedColumn.Transformer) {
+		if transformedColumn.TransformerPath == nil && transformedColumn.Transformer == "" {
+			return errors.Wrap(ErrorSpecifyOnlyOneMissing("transformer", "transformer_path"), Identify(transformedColumn))
+		}
+
+		if transformedColumn.TransformerPath != nil && transformedColumn.Transformer != "" {
+			return ErrorMultipleTransformerSpecified(transformedColumn)
```
deliahu: Can you use `ErrorSpecifyOnlyOne` here?

pkg/operator/api/userconfig/config.go
```diff
@@ -200,19 +200,40 @@ func (config *Config) Validate(envName string) error {
...
+		}
+
+		if transformedColumn.Transformer != "" &&
+			!strings.Contains(transformedColumn.Transformer, ".") &&
+			!slices.HasString(transformerNames, transformedColumn.Transformer) {
 			return errors.Wrap(ErrorUndefinedResource(transformedColumn.Transformer, resource.TransformerType), Identify(transformedColumn), TransformerKey)
+
```
deliahu: nit: extra whitespace

pkg/operator/api/userconfig/errors.go
```diff
@@ -58,6 +58,9 @@ const (
 	ErrK8sQuantityMustBeInt
 	ErrRegressionTargetType
 	ErrClassificationTargetType
+	ErrMultipleAggregatorSpecified
```
deliahu: This can be deleted (see above)

pkg/operator/api/userconfig/errors.go
```diff
@@ -58,6 +58,9 @@ const (
...
+	ErrMultipleTransformerSpecified
```
deliahu: This can be deleted (see above)

pkg/workloads/lib/context.py
```diff
@@ -103,6 +103,8 @@ def __init__(self, **kwargs):
             self.raw_columns, self.transformed_columns  # self.aggregates
         )
 
+        self.ctx["columns"] = self.columns
```
deliahu: This won't be needed after the metadata change

pkg/workloads/spark_job/spark_util.py
```diff
@@ -435,34 +465,71 @@ def _transform(*values):
...
+
+    return type_conversion_dict[obj_type]
+
+
+def validate_transformer(column_name, test_df, ctx, spark):
     transformed_column = ctx.transformed_columns[column_name]
 
     trans_impl, _ = ctx.get_transformer_impl(column_name)
 
     if hasattr(trans_impl, "transform_python"):
```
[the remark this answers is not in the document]
deliahu: I just spoke with Vishal, and he reminded me that it is possible to not define `transform_python()` if you are only using the transformation for training columns, and not for inference. Do you mind putting a comment explaining that on the `transform_python` line of `TRANSFORMER_IMPL_VALIDATION` in `context.py`?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -639,27 +644,41 @@ def validate_transformer(column_name, test_df, ctx, spark):
...
 
 def transform_column(column_name, df, ctx, spark):
     if not ctx.is_transformed_column(column_name):
         return df
     if column_name in df.columns:
         return df
-    transformed_column = ctx.transformed_columns[column_name]
 
+    transformed_column = ctx.transformed_columns[column_name]
+    transformer = ctx.transformers[transformed_column["transformer"]]
```
deliahu: This doesn't seem to be used

pkg/workloads/spark_job/spark_util.py
```diff
@@ -530,9 +525,13 @@ def validate_transformer(column_name, test_df, ctx, spark):
                         + inferred_python_type,
                     )
 
+            ctx.write_metadata(
+                    transformed_column["id"], transformed_column["metadata_key"], {"type": inferred_python_type}
+                )
+
         try:
```
deliahu: I feel like maybe the whole `if hasattr(trans_impl, "transform_python"):` section should be within the `try`? @vishalbollu what do you think?

vishalbollu: yes I agree, then we can add more context to the potential exception in transform_python

pkg/workloads/spark_job/spark_util.py
```diff
@@ -569,22 +650,33 @@ def validate_transformer(column_name, df, ctx, spark):
                         "{} != {}".format(ts_row, tp_row),
```
deliahu: Should lines 630 - 651 be indented left by one level?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -569,22 +650,33 @@ def validate_transformer(column_name, df, ctx, spark):
...
                     )
 
+    if transformer["output_type"] == "unknown":
```
deliahu: I think we should move this to the top of the section above, since it's better to check types before checking values (otherwise the value check will fail and the error message might be confusing). E.g. delete lines 653 - 665, and add this at line 631 after `if hasattr(trans_impl, "transform_spark") and hasattr(trans_impl, "transform_python"):`:

```
if inferred_spark_type != inferred_python_type:
    raise UserRuntimeException(
        "transformed column " + column_name,
        "type inference failed, transform_spark and transform_python had differing types.",
        "transform_python: " + inferred_python_type,
        "transform_spark: " + inferred_spark_type,
    )
```

(they'll either both be `None` or not, so I think the `if` check is sufficient)

PR #105, opened 2019-05-22
Refactor away metadata_key

### LINKED ISSUE #104: Refactor resource metadata files to use shared metadata root directory
### Description

Metadata files should be stored in a shared metadata root directory, rather than a path scoped to resources like it is now.

### Motivation

Consuming metadata will no longer require a `metadata_key`, and the storage path can be derived only from the resource ID

1vn:
Closes #104 

---
Checklist:
- [x] Update documentation


pkg/operator/context/aggregates.go
```diff
@@ -86,15 +86,13 @@ func getAggregates(
 			id,
 		)
 		aggregateKey := aggregateRootKey + ".msgpack"
```
deliahu: Maybe you don't need `aggregateRootKey` now?

PR #111, opened 2019-05-28
Make raw column kinds optional

### LINKED ISSUE #103: Make raw columns optional
### Description

If the user doesn't specify any raw features, infer them from the environment's schema.

### Motivation

This is a continuation of making all schemas optional to reduce the amount of required config.

### Notes

* Still check that all defined raw columns are ingested
* Still ensure schemas from all environments match
* Should we ingest columns that aren't used anywhere in the pipeline?

1vn:
closes #103 

---
Checklist:
- [ ] Test end to end manually (e.g. build/push all images, restart local operator, and run `cx refresh` in an example folder)
- [ ] Update documentation
- [ ] Update examples and `cx init`


pkg/workloads/spark_job/spark_util.py
```diff
@@ -246,26 +268,27 @@ def ingest(ctx, spark):
...
     csv_config = {
         util.snake_to_camel(param_name): val
         for param_name, val in data_config.get("csv_config", {}).items()
         if val is not None
     }
 
-    df = spark.read.csv(
-        data_config["path"], schema=StructType(schema_fields), mode="FAILFAST", **csv_config
-    )
+    df = spark.read.csv(data_config["path"], inferSchema=True, **csv_config)
```
deliahu: Is it possible to only infer schema on certain columns? That seems like it would be safer to me (e.g if you have a string column which happens to have values that look like ints). @vishalbollu any thoughts here?

As you're working on this you might also want to take the chance to support underscores for skipping and leaving some of the columns off the end (e.g. if you had columns named 1, 2, 3, 4, 5 and you cared about 1, 2, and 4, you could ingest them with `[1, 2, _, 4]`)

Also, can you add a test that confirms we use the correct types, e.g. something like this:

```python
schema = [a_str, b_float]
csv_str = "1,1"
```

pkg/workloads/spark_job/spark_util.py
```diff
@@ -223,8 +223,30 @@ def ingest(ctx, spark):
 
     for raw_column_name in ctx.raw_columns.keys():
         raw_column = ctx.raw_columns[raw_column_name]
-        expected_types = CORTEX_TYPE_TO_ACCEPTABLE_SPARK_TYPES[raw_column["type"]]
         actual_type = input_type_map[raw_column_name]
+
+        if actual_type not in SPARK_TYPE_TO_CORTEX_TYPE.keys():
```
deliahu: when would this `if` be `true`?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -223,8 +223,30 @@ def ingest(ctx, spark):
...
+            actual_type = StringType()
+
+        column_type = raw_column["type"]
+        if column_type == consts.COLUMN_TYPE_INFERRED:
+            TEST_DF_SIZE = 10
+            sample_df = df.select(raw_column_name).limit(TEST_DF_SIZE).collect()
```
deliahu: This is being executed for each raw column, but should only have to be executed once

pkg/workloads/spark_job/spark_util.py
```diff
@@ -223,8 +223,30 @@ def ingest(ctx, spark):
...
+            sample = sample_df[0][raw_column_name]
+            inferred_type = infer_type(sample)
+
+            for row in sample_df:
```
deliahu: Do we need to actually check the data? My guess is since it's a dataframe, all elements would be the same type. So can we just use the type of the column in the dataframe schema?

pkg/workloads/spark_job/spark_util.py
1vn: if it somehow inferred a spark type we did not support, like timestamp

pkg/workloads/spark_job/spark_util.py
1vn: thats a good point, we should delete the check for transformed_column as well

pkg/workloads/spark_job/spark_util.py
vishalbollu: It isn't possible to partially infer columns. Better option is to infer all the columns and cast to the known types.

pkg/workloads/spark_job/spark_util.py
```diff
@@ -223,8 +223,20 @@ def ingest(ctx, spark):
```
vishalbollu: Spark dataframe provides the schema for you, you can use that to create a map:

```
inferred_col_type_map = {(c.name: c.dataType) for c in df.schema}
```

The inferred_col_type_map that maps column name to spark type. You can use this to convert to cortex type. Instead of calling infer_type on the first row of every column.

pkg/workloads/spark_job/spark_util.py
```diff
@@ -246,26 +258,40 @@ def ingest(ctx, spark):
...
+    for column_name in ctx.raw_columns:
+        column_type = ctx.raw_columns[column_name]["type"]
+        if column_type == consts.COLUMN_TYPE_INFERRED:
+            casted_cols.append(F.col(column_name).alias(column_name))
+        else:
+            casted_cols.append(
+                F.col(column_name).cast(CORTEX_TYPE_TO_SPARK_TYPE[column_type]).alias(column_name)
+            )
+
+    df = df.select(*casted_cols)
```
vishalbollu: you can `return df.select(*casted_cols)`

pkg/workloads/spark_job/spark_util.py
```diff
@@ -246,26 +258,40 @@ def ingest(ctx, spark):
...
+            + " but got "
+            + str(set(df.columns))
+        )
+
+    # for columns with types, cast it
+    casted_cols = []
+    for column_name in ctx.raw_columns:
+        column_type = ctx.raw_columns[column_name]["type"]
+        if column_type == consts.COLUMN_TYPE_INFERRED:
+            casted_cols.append(F.col(column_name).alias(column_name))
```
vishalbollu: alias is not needed here since no transformation has been applied to the column
`casted_cols.append(F.col(column_name))`

pkg/workloads/spark_job/spark_util.py
```diff
@@ -220,53 +229,61 @@ def ingest(ctx, spark):
...
 
-    df = spark.read.csv(
-        data_config["path"], schema=StructType(schema_fields), mode="FAILFAST", **csv_config
-    )
-    return df.select(*ctx.raw_columns.keys())
+    df = spark.read.csv(data_config["path"], inferSchema=True, mode="FAILFAST", **csv_config)
+    renamed_cols = [F.col(c).alias(data_config["schema"][idx]) for idx, c in enumerate(df.columns)]
+    df = df.select(*renamed_cols)
+
+    if set(data_config["schema"]) != set(df.columns):
```
deliahu: Does this check make sense, since the columns are being renamed on line 274? Maybe we should just check the number of ingested columns matches the number of columns in the schema, and complain if not equal?

Try hitting this error as a user would and let me know what you think.

pkg/workloads/spark_job/spark_util.py
```diff
@@ -220,53 +229,61 @@ def ingest(ctx, spark):
...
-                "raw column " + raw_column_name,
-                "type mismatch",
-                "expected {} but found {}".format(
-                    " or ".join(str(x) for x in expected_types), actual_type
-                ),
-            )
-        target_type = CORTEX_TYPE_TO_SPARK_TYPE[raw_column["type"]]
+        expected_cortex_type = raw_column["type"]
+        if expected_cortex_type == consts.COLUMN_TYPE_INFERRED:
+            column_type = SPARK_TYPE_TO_CORTEX_TYPE[inferred_col_type_map[raw_column_name]]
```
deliahu: Does spark read numbers as floats or doubles, and longs or ints? If not, do we need to cast int -> float and double -> long? (since we expect only float and long in our ingested dataset)

Also, is this where we end up if the user is ingesting parquet and has a column of type int (which we need to cast to float), and will that be casted?

Also, what happens if it gets inferred as a column type that we didn't enumerate (e.g. date), or if the user is ingesting a parquet which has column types that we didn't expect (also e.g. date)? In that case, should we cast to string? Or error?

Something like we have on lines 257-259 could address these. Can you also add tests to confirm these three cases?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -220,53 +229,61 @@ def ingest(ctx, spark):
         df = read_parquet(ctx, spark)
 
     input_type_map = {f.name: f.dataType for f in df.schema}
-
-    for raw_column_name in ctx.raw_columns.keys():
+    inferred_col_type_map = {c.name: c.dataType for c in df.schema}
```
deliahu: `input_type_map` is the same as `inferred_col_type_map`. Probably `input_type_map` makes sense, since for parquet we don't infer it

pkg/workloads/spark_job/spark_util.py
```diff
@@ -220,53 +229,61 @@ def ingest(ctx, spark):
...
-            )
-        target_type = CORTEX_TYPE_TO_SPARK_TYPE[raw_column["type"]]
+        expected_cortex_type = raw_column["type"]
+        if expected_cortex_type == consts.COLUMN_TYPE_INFERRED:
+            column_type = SPARK_TYPE_TO_CORTEX_TYPE[inferred_col_type_map[raw_column_name]]
+            ctx.write_metadata(raw_column["id"], {"type": column_type})
+        else:
+            actual_spark_type = input_type_map[raw_column_name]
+            if actual_spark_type not in SPARK_TYPE_TO_CORTEX_TYPE:
+                actual_spark_type = StringType()
```
deliahu: Is this correct? If we get an unexpected type, we need to cast it to string, but setting `actual_spark_type` to `StringType()` means that we won't run the cast (lines 257-259). Can you add a test for this?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -48,6 +48,15 @@
     consts.COLUMN_TYPE_STRING_LIST: [ArrayType(StringType(), True)],
 }
 
+CORTEX_TYPE_TO_CASTABLE_SPARK_TYPES = {
+    consts.COLUMN_TYPE_INT: [IntegerType(), LongType()],
+    consts.COLUMN_TYPE_INT_LIST: [ArrayType(IntegerType(), True), ArrayType(LongType(), True)],
+    consts.COLUMN_TYPE_FLOAT: [FloatType(), DoubleType()],
+    consts.COLUMN_TYPE_FLOAT_LIST: [ArrayType(FloatType(), True), ArrayType(DoubleType(), True)],
+    consts.COLUMN_TYPE_STRING: [StringType(), IntegerType(), LongType(), FloatType(), DoubleType()],
```
deliahu: Should this be a catch-all? Can we cast any type to string if the user asked for it?

pkg/workloads/spark_job/spark_util.py
1vn: Spark would assign default names (i.e. `c1,c2,c3`) when `inferSchema` is True, so we rename it back to the column names the user provided. This check catches the case when we're missing columns. I'll be refactor it to be more clear.

pkg/workloads/spark_job/spark_util.py
1vn: Yes.

pkg/workloads/spark_job/spark_job.py
```diff
@@ -150,6 +151,16 @@ def ingest_raw_dataset(spark, ctx, cols_to_validate, should_ingest):
             logger.info("Ingesting {} data from {}".format(ctx.app["name"], data_config["path"]))
             ingest_df = spark_util.ingest(ctx, spark)
 
+            input_type_map = {f.name: f.dataType for f in ingest_df.schema}
```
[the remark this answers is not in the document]
1vn: I moved it here so that the `ingest` can be unit tested without the need to account for metadata upload logic (which I'd have to mock to test properly)

pkg/workloads/spark_job/spark_util.py
```diff
@@ -220,53 +229,61 @@ def ingest(ctx, spark):
...
         for param_name, val in data_config.get("csv_config", {}).items()
         if val is not None
     }
 
-    df = spark.read.csv(
-        data_config["path"], schema=StructType(schema_fields), mode="FAILFAST", **csv_config
-    )
-    return df.select(*ctx.raw_columns.keys())
+    df = spark.read.csv(data_config["path"], inferSchema=True, mode="FAILFAST", **csv_config)
+    renamed_cols = [F.col(c).alias(data_config["schema"][idx]) for idx, c in enumerate(df.columns)]
```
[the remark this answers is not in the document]
1vn: count may be equal but elements in the set could be different

pkg/workloads/spark_job/spark_util.py
```diff
@@ -223,40 +245,60 @@ def value_check_data(ctx, df, raw_columns=None):
...
+            if actual_spark_type not in SPARK_TYPE_TO_CORTEX_TYPE:
+                df = df.withColumn(raw_column_name, F.col(raw_column_name).cast(StringType()))
+            else:
+                actual_cortex_type = SPARK_TYPE_TO_CORTEX_TYPE[actual_spark_type]
+                expected_spark_type = CORTEX_TYPE_TO_SPARK_TYPE[actual_cortex_type]
+                if actual_spark_type != expected_spark_type:
+                    df = df.withColumn(
+                        raw_column_name, F.col(raw_column_name).cast(expected_spark_type)
+                    )
+            continue
```
deliahu: Is this `continue` necessary?

pkg/workloads/spark_job/spark_util.py
```diff
@@ -49,12 +49,34 @@
...
+        consts.COLUMN_TYPE_FLOAT_LIST: [
+            ArrayType(FloatType(), True),
+            ArrayType(DoubleType(), True),
+        ],
+        consts.COLUMN_TYPE_STRING: [
+            StringType(),
+            IntegerType(),
+            LongType(),
+            FloatType(),
+            DoubleType(),
```
deliahu: Can you add the array types here? I have no idea how it would be read, but in theory if it could, this cast would be supported.

pkg/workloads/spark_job/spark_util.py
```diff
@@ -49,12 +49,34 @@
...
+        ],
+        consts.COLUMN_TYPE_STRING_LIST: [ArrayType(StringType(), True)],
```
deliahu: Can you add the array types of the other primitives here? In theory if it could be read, these casts would be supported.

pkg/workloads/spark_job/spark_util.py
```diff
@@ -223,40 +245,60 @@ def value_check_data(ctx, df, raw_columns=None):
...
+                    )
+            else:
+                try:
+                    df = df.withColumn(
+                        raw_column_name, F.col(raw_column_name).cast(expected_spark_type)
+                    )
+                except Exception as e:
+                    raise UserException(
+                        "tried casting " + raw_column_name,
+                        "to from ingested type " + actual_spark_type,
```
deliahu: nit: "to from..." -> "to...

PR #155, opened 2019-06-11
Add region field to input data

### LINKED ISSUE #115: Support bucket regions for data ingestion
### Description

Currently we default to the same region that your cluster is in. That should be the default behavior, but we should allow ingesting from buckets in other regions. We should also update all of our examples to specify the us-west-2, so that the examples can run in any region.

### Notes

We either need to add the region to the bucket URL, or add a config field for region

vishalbollu:
Resolves #115 
---
Checklist:
- [x] Update documentation

### DIFF
--- a/pkg/lib/aws/s3.go
+++ b/pkg/lib/aws/s3.go
@@ -43,6 +43,8 @@ func init() {
 	resolver := endpoints.DefaultResolver()
 	partitions := resolver.(endpoints.EnumPartitions).Partitions()
 
+	S3Regions = strset.New()
+
 	for _, p := range partitions {
 		if p.ID() == endpoints.AwsPartitionID || p.ID() == endpoints.AwsCnPartitionID {
 			for id := range p.Regions() {
--- a/pkg/lib/aws/s3.go
+++ b/pkg/lib/aws/s3.go
@@ -31,22 +31,22 @@ import (
 	"github.com/cortexlabs/cortex/pkg/lib/json"
 	"github.com/cortexlabs/cortex/pkg/lib/msgpack"
 	"github.com/cortexlabs/cortex/pkg/lib/parallel"
+	"github.com/cortexlabs/cortex/pkg/lib/sets/strset"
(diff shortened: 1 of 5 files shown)


pkg/lib/aws/s3.go
```diff
@@ -21,38 +21,44 @@ import (
...
 	"github.com/cortexlabs/cortex/pkg/lib/json"
+	"github.com/cortexlabs/cortex/pkg/lib/msgpack"
 	"github.com/cortexlabs/cortex/pkg/lib/parallel"
 	s "github.com/cortexlabs/cortex/pkg/lib/strings"
 )
 
-func (c *Client) S3Path(key string) string {
-	return "s3://" + filepath.Join(c.Bucket, key)
-}
+var DefaultS3Region string
```
deliahu: does `const DefaultS3Region string = endpoints.UsWest2RegionID` work here, and remove it from the `init()`?

pkg/lib/aws/s3.go
```diff
@@ -21,38 +21,44 @@ import (
...
+var S3Regions []string
```
deliahu: maybe use `strset`?

pkg/operator/workloads/data_job.go
```diff
@@ -183,10 +184,13 @@ func dataWorkloadSpecs(ctx *context.Context) ([]*WorkloadSpec, error) {
 
 	shouldIngest := !rawFileExists
 	if shouldIngest {
-		externalDataPath := ctx.Environment.Data.GetExternalPath()
-		externalDataExists, err := config.AWS.IsS3aPrefixExternal(externalDataPath)
+		externalData := ctx.Environment.Data.GetExternalData()
+		externalDataExists, err := aws.IsS3aPrefixExternal(externalData.Path, externalData.Region)
 		if err != nil || !externalDataExists {
```
deliahu: I think you meant to remove the `|| !externalDataExists`?

Issue #177, opened 2019-06-24
Show API schema for external models


deliahu:
### Description

`cx get <api_name>` should show the input JSON schema for external models (just like with pipeline models).

PR #240, opened 2019-07-19
Trim cortex get and add descriptive logging around request handlers

### LINKED ISSUE #207: Improve logging around request handlers
### Description

Log output of pre_inference request handler and input to post_inference request handler

### Motivation

Help users debug their request handlers by printing out their output.

### Additional Context

- output of pre_inference and input to post_inference may be a numpy which is not JSON seralizable by default

vishalbollu:
Closes #207 #177 
---
Checklist:
- [X] Run `make test` and `make lint`
- [X] Test end to end manually (e.g. build/push all images, restart operator, and run `cx refresh`)
- [X] Update examples
- [ ] Merge to master

### DIFF
--- a/cli/cmd/get.go
+++ b/cli/cmd/get.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"fmt"
+	"net/http"
 	"sort"
 	"strings"
 	"time"
@@ -45,6 +46,7 @@ func init() {
 	addEnvFlag(getCmd)
 	addWatchFlag(getCmd)
 	addSummaryFlag(getCmd)
+	addVerboseFlag(getCmd)
 	// addResourceTypesToHelp(getCmd)
 }
 
@@ -92,7 +94,7 @@ func runGet(cmd *cobra.Command, args []string) (string, error) {
 			return "", err
 		}
 
-		return resourceByNameStr(resourceNameOrType, resourcesRes)
+		return resourceByNameStr(resourceNameOrType, resourcesRes, flagVerbose)
 
 	case 2:
 		userResourceType := args[0]
@@ -101,7 +103,7 @@ func runGet(cmd *cobra.Command, args []string) (string, error) {
 		if err != nil {
 			return "", resource.ErrorInvalidType(userResourceType)
 		}
-		return resourceByNameAndTypeStr(resourceName, resourceType, resourcesRes)
+		return resourceByNameAndTypeStr(resourceName, resourceType, resourcesRes, flagVerbose)
 	}
 
 	return "", errors.New("too many args") // unexpected
@@ -127,7 +129,7 @@ func getResourcesResponse() (*schema.GetResourcesResponse, error) {
 	return &resourcesRes, nil
 }
 
-func resourceByNameStr(resourceName string, resourcesRes *schema.GetResourcesResponse) (string, error) {
+func resourceByNameStr(resourceName string, resourcesRes *schema.GetResourcesResponse, flagVerbose bool) (string, error) {
 	rs, err := resourcesRes.Context.VisibleResourceByName(resourceName)
 	if err != nil {
 		return "", err
@@ -146,7 +148,7 @@ func resourceByNameStr(resourceName string, resourcesRes *schema.GetResourcesRes
 	case resource.ModelType:
 		return describeModel(resourceName, resourcesRes)
 	case resource.APIType:
-		return describeAPI(resourceName, resourcesRes)
+		return describeAPI(resourceName, resourcesRes, flagVerbose)
 	default:
 		return "", resource.ErrorInvalidType(resourceType.String())
 	}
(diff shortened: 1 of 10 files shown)


dev/registry.sh
```diff
@@ -148,7 +148,7 @@ elif [ "$cmd" = "update" ]; then
     build_and_push $ROOT/images/nvidia nvidia latest
     build_and_push $ROOT/images/metrics-server metrics-server latest
   fi
-
+  build_and_push $ROOT/images/spark spark latest
```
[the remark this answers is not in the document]
deliahu: Also if you put this here, you might also want to bring over tf-train too

pkg/operator/workloads/logs.go
```diff
@@ -441,20 +439,6 @@ func extractFromCortexLog(match string, loglevel string, logStr string) (*string
 		return formatHeader3(cutStr), false
 	}
 
-	matches := jsonPrefixRegex.FindStringSubmatch(cutStr)
```
deliahu: Was this removed to save space on `cx logs`?

cli/cmd/get.go
```diff
@@ -480,14 +489,73 @@ func describeAPI(name string, resourcesRes *schema.GetResourcesResponse) (string
...
+		t := table.Table{
+			Headers: []table.Header{
+				{Title: "FEATURE", MaxWidth: 32},
+				{Title: "TYPE", MaxWidth: 10},
+				{Title: "SHAPE", MaxWidth: 20},
+			},
+			Rows: rows,
+		}
+
+		out += table.MustFormat(t)
```
deliahu: It it worth also showing what the curl would look like, like we do for e2e? Since you printed the table we wouldn't need to include the shape information in the sample curl.

cli/cmd/get.go
```diff
@@ -480,14 +489,73 @@ func describeAPI(name string, resourcesRes *schema.GetResourcesResponse) (string
...
 		samplesPlaceholderStr := `{ "samples": [ { ` + strings.Join(samplePlaceholderFields, ", ") + " } ] }"
 		out += "Payload:  " + samplesPlaceholderStr + "\n"
-	}
-	if api != nil {
-		out += titleStr("Configuration") + api.UserConfigStr()
+	} else {
+		out += titleStr("Model Input Signature")
+
+		if groupStatus.Available() == 0 {
+			out += "Model "
```
deliahu: Is this intended? Maybe this check should happen before `out += titleStr("Model Input Signature")`?

cli/cmd/get.go
```diff
@@ -480,14 +489,73 @@ func describeAPI(name string, resourcesRes *schema.GetResourcesResponse) (string
...
+			return out, nil
+		}
+
+		modelInput, err := getModelInput(urls.Join(apiEndpoint, "signature"))
+		if err != nil {
+			return "", errors.Wrap(err, "")
```
deliahu: We might want to silently error here, since this might happen - right now I think "ready" is a little bit of a misnomer since the pod might be ready, but the server might not be running yet (although I haven't confirmed that yet, it's on my todo list)

pkg/workloads/cortex/onnx_serve/api.py
```diff
@@ -29,25 +29,29 @@
...
+    "tensor(uint32)": "uint32",
     "tensor(int8)": "int8",
     "tensor(uint8)": "uint8",
     "tensor(int16)": "int16",
     "tensor(uint16)": "uint16",
     "tensor(int64)": "int64",
     "tensor(uint64)": "uint64",
     "tensor(bool)": "bool",
-    "tensor(string)": "string",
+    "tensor(string)": "object",  # variable string lengths, numpy string types are fixed-length
```
deliahu: I don't totally understand your comment, but would "object or string" make sense?

pkg/workloads/cortex/lib/util.py
```diff
@@ -48,9 +51,19 @@ def indent_str(text, indent):
     return indent * " " + text.replace("\n", "\n" + indent * " ")
 
 
+def json_tricks_dump(obj, **kwargs):
```
deliahu: Maybe this should just use `pp_str_flat()` in other files and leave this as a helper internal to `util.py`?

cli/cmd/get.go
```diff
@@ -445,22 +446,30 @@ func describeAPI(name string, resourcesRes *schema.GetResourcesResponse) (string
 		staleComputeStr = fmt.Sprintf(" (%s %s previous compute)", s.Int32(groupStatus.ReadyStaleCompute), hasStr)
 	}
 
-	out := titleStr("Summary")
-	out += fmt.Sprintf("Status:  %s\n", groupStatus.Message())
+	apiEndpoint := urls.Join(resourcesRes.APIsBaseURL, anyAPIStatus.Path)
+
+	out := "\nURL:      " + apiEndpoint + "\n"
+	out += fmt.Sprintf("cURL:     curl -k -X POST -H \"Content-Type: application/json\" %s -d @<json_file_path>\n", apiEndpoint)
```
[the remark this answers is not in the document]
vishalbollu: yea it is required when passing a file path to -d flag instead of json string

pkg/operator/workloads/logs.go
vishalbollu: Yes, with the addition of printing the output of pre_inference and post_inference, a single prediction request was producing many log lines, even for something as small as iris

cli/cmd/get.go
vishalbollu: The sample curl I am imagining for say xgboost iris would be something like this:

```
{
  input: [int]
}
```

That doesn't look like new information that isn't already described in the table. Additionally, it can not just be copied and pasted. We can discuss this more.

PR #323, opened 2019-08-14
Support model directories on S3

### LINKED ISSUE #215: Support specifying an s3 directory for TensorFlow model paths
### Description

As an alternative to requiring the user to zip the TensorFlow export directory, allow the user to specify an s3 prefix which contains the exported model.

1vn:
Closes #215 

---
Checklist:
- [ ] Update examples
- [ ] Merge to master


pkg/lib/models/models.go
```diff
@@ -0,0 +1,66 @@
...
+
+// IsValidS3Directory checks that the path contains a valid S3 directory for Tensorflow models
+// Must contain the following structure:
+// - 1523423423/ (version prefix, usually a timestamp)
+// 		- saved_model.pb
+//		- variables/
+//			- variables.index
+//			- variables.data-00000-of-00001 (there are a variable number of these files)
+func IsValidS3Directory(path string) bool {
+	listOut, err := aws.ListObjectsExternal(path)
```
deliahu: Just to be safe, I think we should do `path = s.EnsureSuffix(path, "/")` before calling `aws.ListObjectsExternal()` since there could be an edge case where file names collide. E.g. if you have `path/to/1`, `path/to/2`, and `path/today`, I think `ListObjectsExternal("path/to")` will return all three files (but not if we add the trailing `/`)

deliahu: On second thought, I think it would be simpler if we checked `IsS3FileExternal()` for each of the desired files, rather than listing. To save re-creating the s3 client each time, maybe `IsS3FileExternal()` can take in `keys ...string`). `ListObjectsExternal()` could then be deleted.

pkg/workloads/cortex/lib/storage/s3.py
```diff
@@ -249,3 +249,7 @@ def get_json_external(self, s3_path, num_retries=0, retry_delay_sec=2):
         if obj is None:
             return None
         return json.loads(obj.decode("utf-8"))
+
+    def list_objects(self, s3_path):
```
deliahu: To maintain the interface, can you implement this function in `storage/local.py` too? Also, for the implementations to match, maybe just a list of `obj["Key"]` should be returned in this case, and the file paths in the local case

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -442,6 +442,24 @@ def get_signature(app_name, api_name):
     return jsonify(response)
 
 
+def download_dir_external(ctx, s3_path, local_path):
+    util.mkdir_p(local_path)
+    bucket_name, prefix = ctx.storage.deconstruct_s3_path(s3_path)
+    objects = ctx.storage.list_objects(s3_path)["Contents"]
+    version = prefix.split("/")[-1]
+    for obj in objects:
+        local_key = obj["Key"].lstrip(prefix[: -len(version)])
```
deliahu: Can this be simplified?

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -442,6 +442,24 @@ def get_signature(app_name, api_name):
...
+        if not os.path.exists(os.path.dirname(local_key)):
+            util.mkdir_p(os.path.join(local_path, os.path.dirname(local_key)))
+
+        if obj["Key"][-1] == "/":
```
deliahu: Is this check necessary? I wouldn't expect directories to be returned by `list_objects()`.

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -480,13 +498,11 @@ def start(args):
         if api.get("request_handler") is not None:
             local_cache["request_handler"] = ctx.get_request_handler_impl(api["name"])
 
-        if not util.is_resource_ref(api["model"]):
-            if not os.path.isdir(args.model_dir):
-                ctx.storage.download_and_unzip_external(api["model"], args.model_dir)
+        if args.only_download:
```
deliahu: I think the previous logic around this should be preserved, but not entirely sure. It seems like e2e would be broken with this logic (since it still uses the zip, but its init container would not unzip). We can discuss

pkg/operator/api/userconfig/apis.go
```diff
@@ -118,24 +119,27 @@ func (apis APIs) Validate() error {
...
+	switch {
+	case strings.HasSuffix(api.Model, ".onnx"):
+		api.ModelFormat = ONNXModelFormat
 		if ok, err := aws.IsS3PathFileExternal(api.Model); err != nil || !ok {
 			return errors.Wrap(ErrorExternalNotFound(api.Model), Identify(api), ModelKey)
 		}
+	case models.IsValidS3Directory(api.Model):
+		api.ModelFormat = TensorFlowModelFormat
+	default:
+		return errors.Wrap(ErrorUnableToInferModelFormat(), Identify(api))
```
deliahu: The message in `ErrorUnableToInferModelFormat` needs to be updated. Can you describe something like you wrote in you comment in `models/models.go`? I can see users hitting this, so we should be as clear as possible in this error message, it's ok if it ends up being a bit long.

pkg/workloads/cortex/tf_api/api.py
1vn: Yes, list objects would include top level dir

pkg/lib/models/models.go
1vn: need to check for `containsVariableDataFile`

pkg/workloads/cortex/lib/storage/s3.py
```diff
@@ -250,6 +250,10 @@ def get_json_external(self, s3_path, num_retries=0, retry_delay_sec=2):
...
         return json.loads(obj.decode("utf-8"))
 
-    def list_objects(self, s3_path):
-        bucket, prefix = self.deconstruct_s3_path(s3_path)
-        return self.s3.list_objects(Bucket=bucket, Prefix=prefix)
+    def list_objects(self, path):
+        bucket, prefix = self.deconstruct_s3_path(path)
+        return [
+            obj["Key"]
+            for obj in self.s3.list_objects(Bucket=bucket, Prefix=prefix)["Contents"]
```
deliahu: This should be [list_objects_v2](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.list_objects_v2)

pkg/workloads/cortex/lib/storage/s3.py
```diff
@@ -250,6 +250,10 @@ def get_json_external(self, s3_path, num_retries=0, retry_delay_sec=2):
...
+            if obj["Key"][:1] != "/"
```
deliahu: nit: `endswith()` would be more clear

pkg/workloads/cortex/lib/storage/local.py
```diff
@@ -141,3 +141,6 @@ def download_and_unzip(self, key, local_dir):
         local_zip = os.path.join(local_dir, "zip.zip")
         self.download_file(key, local_zip)
         util.extract_zip(local_zip, delete_zip_file=True)
+
+    def list_objects(self, path):
+        return os.listdir(path)
```
deliahu: This isn't exactly the same behavior as the s3 one, since it should return file paths of all files at or below the directory. I think `os.walk()` can help.

pkg/workloads/cortex/lib/storage/s3.py
```diff
@@ -250,6 +250,10 @@ def get_json_external(self, s3_path, num_retries=0, retry_delay_sec=2):
             return None
         return json.loads(obj.decode("utf-8"))
 
-    def list_objects(self, s3_path):
-        bucket, prefix = self.deconstruct_s3_path(s3_path)
-        return self.s3.list_objects(Bucket=bucket, Prefix=prefix)
+    def list_objects(self, path):
```
deliahu: Do we need to handle continuation tokens?

deliahu: EDIT: I remember dealing with this once, I think we can use `search()` in this file for `list_objects()` (which correctly handles continuation tokens). The only issue is that assumes the internal bucket, so maybe you can generalize that?

pkg/lib/models/models.go
```diff
@@ -34,33 +30,17 @@ import (
 //			- variables.index
 //			- variables.data-00000-of-00001 (there are a variable number of these files)
 func IsValidS3Directory(path string) bool {
```
deliahu: What do you think about moving this function into the `userconfig` package, and later we can move it out if other packages need it? It could go either way, up to you

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -445,18 +445,15 @@ def get_signature(app_name, api_name):
 def download_dir_external(ctx, s3_path, local_path):
     util.mkdir_p(local_path)
     bucket_name, prefix = ctx.storage.deconstruct_s3_path(s3_path)
-    objects = ctx.storage.list_objects(s3_path)["Contents"]
+    objects = ctx.storage.list_objects(s3_path)
     version = prefix.split("/")[-1]
     for obj in objects:
-        local_key = obj["Key"].lstrip(prefix[: -len(version)])
+        local_key = obj[len(prefix) - len(version) :]
```
deliahu: I think `list_objects()` should return paths relative to the dir that was passed in. Can you update the implementation to do that (and confirm the local implementation does the same)?

pkg/workloads/cortex/lib/storage/s3.py
deliahu: EDIT2: maybe we can instantiate a new `storage` client in the appropriate bucket in `api.py` and then we wouldn't have to modify `search()`

pkg/workloads/cortex/lib/storage/local.py
```diff
@@ -141,3 +141,13 @@ def download_and_unzip(self, key, local_dir):
         local_zip = os.path.join(local_dir, "zip.zip")
         self.download_file(key, local_zip)
         util.extract_zip(local_zip, delete_zip_file=True)
+
+    def list_objects(self, path):
+        objects = []
+        for root, dirs, files in os.walk(path):
+            for name in files:
+                objects.append(os.path.join(root, name))
+            for name in dirs:
```
deliahu: I don't think the S3 version of `list_objects()` returns directories, just objects. Can you confirm, and if so, only return files here?

pkg/workloads/cortex/lib/storage/local.py
```diff
@@ -141,3 +141,13 @@ def download_and_unzip(self, key, local_dir):
         local_zip = os.path.join(local_dir, "zip.zip")
         self.download_file(key, local_zip)
         util.extract_zip(local_zip, delete_zip_file=True)
+
+    def list_objects(self, path):
+        objects = []
+        for root, dirs, files in os.walk(path):
+            for name in files:
+                objects.append(os.path.join(root, name))
```
deliahu: Does this return the path relative to the `path` argument? If not, maybe it should to match the s3 implementation

pkg/workloads/cortex/lib/storage/s3.py
```diff
@@ -249,3 +249,9 @@ def get_json_external(self, s3_path, num_retries=0, retry_delay_sec=2):
         if obj is None:
             return None
         return json.loads(obj.decode("utf-8"))
+
+    def list_objects(self, path):
+        _, prefix = self.deconstruct_s3_path(path)
+        return [
+            obj[len(prefix) + 1 :] for obj in self.search(prefix=prefix) if not obj.endswith("/")
```
deliahu: You might be able to delete the `obj.endswith("/")`, can you check?

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -493,23 +510,22 @@ def start(args):
...
-                ctx.storage.download_and_unzip_external(api["model"], args.model_dir)
+        if not os.path.isdir(args.model_dir):
+            if util.is_resource_ref(api["model"]):
+                model_name = util.get_resource_ref(api["model"])
+                model = ctx.models[model_name]
+                ctx.storage.download_and_unzip(model["key"], args.model_dir)
+            else:
+                download_dir_external(ctx, api["model"], args.model_dir)
 
             if args.only_download:
```
deliahu: This might as well be indented left one level, in case we pass in args.only_download but the file is already present

pkg/workloads/cortex/lib/storage/s3.py
```diff
@@ -249,3 +249,9 @@ def get_json_external(self, s3_path, num_retries=0, retry_delay_sec=2):
         if obj is None:
             return None
         return json.loads(obj.decode("utf-8"))
+
+    def list_objects(self, path):
```
deliahu: I think list_object should take in the path without the leading `s3://<bucket_name>`, since the client already knows the bucket, what do you think?

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -441,6 +442,22 @@ def get_signature(app_name, api_name):
     return jsonify(response)
 
 
+def download_dir_external(ctx, s3_path, local_path):
+    util.mkdir_p(local_path)
+    bucket_name, prefix = ctx.storage.deconstruct_s3_path(s3_path)
+    storage_client = S3(bucket_name, client_config={})
+    objects = storage_client.list_objects(s3_path)
+    version = prefix.split("/")[-1]
```
deliahu: We should ensure `prefix` doesn't end with a `/` before calling `split()`

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -442,8 +442,8 @@ def download_dir_external(ctx, s3_path, local_path):
     util.mkdir_p(local_path)
     bucket_name, prefix = ctx.storage.deconstruct_s3_path(s3_path)
     storage_client = S3(bucket_name, client_config={})
-    objects = storage_client.list_objects(s3_path)
-    version = prefix.split("/")[-1]
+    objects = [obj[len(prefix) + 1 :] for obj in storage_client.search(prefix=prefix)]
```
deliahu: I think this trim will also not work correctly if the user includes the trailing "/". It might be safest to add the "/" to `prefix` (if it's not there already) before calling `search`, since then you'll also avoid the situation of matching prefixes which are not in the correct directory (e.g. if the user types in `1/2/3`, I think it would match a file with path `1/2/33/1` in addition to `1/2/3/1`.

PR #460, opened 2019-09-13
Improve signature def detection

### LINKED ISSUE #451: Help users automatically detect signature key
### Description

- Look for common signature def names, `predict`, `serving_default`
- Look for signature def with method name `tensorflow/serving/predict`
- If there are multiple prompt users to configure signature_key
- Print entire signature, including signature key

vishalbollu:
Closes #451 

---
Checklist:
- [ ] Update examples
- [ ] Merge to master


pkg/operator/api/userconfig/apis.go
```diff
@@ -115,7 +114,7 @@ var apiValidation = &cr.StructValidation{
 					{
 						StructField: "SignatureKey",
 						StringValidation: &cr.StringValidation{
-							Default: consts.DefaultTFServingSignatureKey,
+							Required: true,
```
deliahu: Should this be `required: false`? I guess right now it doesn’t matter since this is the only field in the struct, but later if we had more fields could this one be optional?

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -252,38 +252,57 @@ def predict(deployment_name, api_name):
...
-    ctx = local_cache["ctx"]
-    api = local_cache["api"]
-
     try:
-        metadata = extract_signature(
-            local_cache["metadata"]["signatureDef"],
-            local_cache["api"]["tf_serving"]["signature_key"],
-        )
+        signature = local_cache["signature"]
     except Exception as e:
```
deliahu: Do we still need the `try/except` here? In what case would an exception be thrown?

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -385,14 +404,16 @@ def start(args):
...
-    )
+
+    signature_key = None
+    if api.get("tf_serving") is not None and api["tf_serving"].get("signature_key") is not None:
+        signature_key = api["tf_serving"]["signature_key"]
+
+    key, metadata = extract_signature(local_cache["metadata"]["signatureDef"], signature_key)
+
+    local_cache["signature_key"] = key
+    local_cache["signature"] = metadata
```
deliahu: What is the relation between `local_cache[“signature”]` (which is set to `metadata`) and `local_cache[“metadata”]`, and should we rename one or both of them?

pkg/workloads/cortex/tf_api/api.py
```diff
@@ -274,9 +276,13 @@ def extract_signature(signature_def, signature_key):
             )
     else:
         if signature_def.get(signature_key) is None:
+            possibilities_str = "key '{}'".format(available_keys[0])
```
deliahu: nit: what do you think about doing `"key: '{}'"` here and `"keys: '{}'"` below?

PR #488, opened 2019-09-26
Cortex python client

### LINKED ISSUE #467: Python Client for cx deploy [1.5D]
### Description

Goal: cx deploy in google colab notebook

vishalbollu:
Closes #467 

---
Checklist:
- [ ] Update examples
- [ ] Merge to master


pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -0,0 +1,124 @@
...
+                )
+                resp.raise_for_status()
+            except HTTPError as err:
+                resp = err.response
+                if "error" in resp.json():
+                    raise Exception(resp.json()["error"]) from err
+                raise
+
+            if resp.status_code == 200:
+                return self.get_endpoint(deployment_name, api_name)
```
deliahu: Do you prefer this over the operator responding to the deploy() with the base URL and the context (which would eliminate the second request)? Do we intend users to also call `get_endpoint()`? If not we could merge it for now to clean up the API and avoid the extra request. Let me know if you want to discuss

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -0,0 +1,124 @@
...
+        self.aws_access_key_id = aws_access_key_id
+        self.aws_secret_access_key = aws_secret_access_key
+        self.headers = {
+            "CortexAPIVersion": "master",
+            "Authorization": "CortexAWS {}|{}".format(
+                self.aws_access_key_id, self.aws_secret_access_key
+            ),
+        }
+
+        pathlib.Path(self.workspace).mkdir(parents=True, exist_ok=True)
```
deliahu: Sweet! Do you mind updating the implementation of (or removing) `mkdir_p()`? I think we made that before switching to Python 3. We can also do that in a separate commit if you prefer.

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -0,0 +1,124 @@
...
+import urllib.parse
+
+import dill
+import requests
+from requests.exceptions import HTTPError
+
+
+class Client(object):
+    def __init__(
+        self, aws_access_key_id, aws_secret_access_key, operator_url, workspace="./workspace"
```
deliahu: What do you think about using `$HOME/.cortex/workspace`?

deliahu: What do you think about not making `workspace` configurable?

pkg/workloads/cortex/lib/context.py
```diff
@@ -108,10 +110,19 @@ def download_python_file(self, impl_key, module_name):
 
     def load_module(self, module_prefix, module_name, impl_path):
         full_module_name = "{}_{}".format(module_prefix, module_name)
-        try:
-            impl = imp.load_source(full_module_name, impl_path)
-        except Exception as e:
-            raise UserException("unable to load python file", str(e)) from e
+
+        if impl_path.endswith("pickle"):
```
deliahu: nit: maybe `impl_path.endswith(".pickle")` would be even safer?

pkg/workloads/cortex/lib/requirements.txt
```diff
@@ -4,3 +4,4 @@ numpy==1.17.2
 json_tricks==3.13.2
 requests==2.22.0
 datadog==0.30.0
+dill==0.3.0
```
deliahu: nit: can you add `dill` to `versions.md`?

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -0,0 +1,124 @@
...
+        self.headers = {
+            "CortexAPIVersion": "master",
+            "Authorization": "CortexAWS {}|{}".format(
+                self.aws_access_key_id, self.aws_secret_access_key
+            ),
+        }
+
+        pathlib.Path(self.workspace).mkdir(parents=True, exist_ok=True)
+
+    def deploy(self, deployment_name, api_name, model_path, pre_inference, post_inference):
```
deliahu: Should we support e.g. `model_format`, `tf_signature_key`, `tracker`, `compute`?

deliahu: Should we use the same fields as the names of the fields in yaml, e.g `model` instead of `model_path`?

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -0,0 +1,124 @@
...
+        with open(os.path.join(api_working_dir, "requirements.txt"), "w") as f:
+            f.writelines(reqs.decode())
+
+        with open(os.path.join(api_working_dir, "request_handler.pickle"), "wb") as f:
+            dill.dump(
+                {"pre_inference": pre_inference, "post_inference": post_inference}, f, recurse=True
+            )
+
+        cortex_config = [
+            {"kind": "deployment", "name": deployment_name},
```
deliahu: Maybe it's ok for now, but would this not support deploying e.g. `iris/tensorflow` and then `iris/sklearn`? If we wanted to do that, would we keep track of a map of api name -> config so we could maintain the entire yaml? Let me know if you think we should discuss.

pkg/workloads/cortex/lib/context.py
```diff
@@ -108,10 +110,19 @@ def download_python_file(self, impl_key, module_name):
...
+            impl = imp.new_module(full_module_name)
+
+            with open(impl_path, "rb") as pickle_file:
+                pickled_dict = dill.load(pickle_file)
+                for key in pickled_dict:
+                    setattr(impl, key, pickled_dict[key])
```
deliahu: Should we catch exceptions here like we do in the `else`?

pkg/workloads/cortex/client/setup.py
```diff
@@ -20,7 +20,7 @@
     description="",
     author="Cortex Labs",
     author_email="dev@cortexlabs.com",
-    install_requires=["dill==0.3.0", "requests==2.21.0"],
+    install_requires=["dill==0.3.0", "requests==2.21.0", "msgpack==0.6.1"],
```
deliahu: Can we, and should we, be more flexible with these versions since the user will be installing this in their environment?

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -29,11 +30,17 @@
...
+        
+        Args:
+            aws_access_key_id (string): AWS access key associated with the account that created the Cortex Cluster
+            aws_secret_access_key (string): AWS secrey key associated with the provided AWS access key  
+            operator_url (string): Operator URL of your cortex cluster
+        """
+
         self.operator_url = operator_url
-        self.workspace = workspace
+        self.workspace = str(Path.home() / ".cortex" / "workspace")
```
deliahu: nit: Is there something like `path.Join()` that would be safer to use here? In either case, `".cortex" / "workspace"` can probably be `".cortex/workspace"` instead

PR #563, opened 2019-11-07
Refresh logger after loading user module

vishalbollu:
Closes #<issue ID>

---
Checklist:
- [ ] Update examples
- [ ] Merge to master
- [ ] Check [gitbook](https://www.cortex.dev/v/master/) that docs look good and links function properly

### DIFF
--- a/pkg/workloads/cortex/lib/log.py
+++ b/pkg/workloads/cortex/lib/log.py
@@ -14,6 +14,7 @@
 
 import logging
 import sys
+import time
 
 from cortex.lib import stringify
 import datetime as dt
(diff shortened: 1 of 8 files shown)


pkg/workloads/cortex/lib/log.py
```diff
@@ -28,23 +29,39 @@ def formatTime(self, record, datefmt):
...
+    logger.propagate = False
+    logger.addHandler(handler)
+    logger.setLevel(logging.DEBUG)
+    return logger
+
+
+def refresh_logger():
+    global current_logger
+    if current_logger is not None:
+        current_logger.disabled = True
```
deliahu: Does this get garbage collected? (since if the user has an `init()` function this will get called on every prediction request)

pkg/workloads/cortex/lib/log.py
```diff
@@ -28,23 +29,39 @@ def formatTime(self, record, datefmt):
...
+    current_logger = register_logger("{}-cortex".format(int(time.time() * 1000000)))
+
+
+def cx_logger():
```
deliahu: Is it possible / is it a good idea / is it worth it to implement each of the logger methods to wrap `current_logger`? E.g. if we had something like

```python
def info(str):
    return current_logger.info(str)
```

then could we do `import cx_logger` and use `cx_logger.info()`? We can discuss

pkg/workloads/cortex/lib/log.py
vishalbollu: Init() doesn't called on every prediction, only once on startup

deliahu: Yep, sorry for my confusion!

PR #580, opened 2019-11-13
Add answer generator example

ospillinger:


examples/pytorch/answer-generator/generator.py
```diff
@@ -0,0 +1,39 @@
...
+
+import torch
+import torch.nn.functional as F
+
+
+def generate(model, conditioned_tokens, device):
+    generated_tokens = []
+    while True:
+        result = recalc(model, conditioned_tokens, generated_tokens, device)
+        if result == 50256:
```
vishalbollu: assign 50256 to a variable so that it can be used in the predictor.py

examples/pytorch/answer-generator/predictor.py
```diff
@@ -0,0 +1,29 @@
...
+def init(model_path, metadata):
+    wget.download(
+        "https://convaisharables.blob.core.windows.net/lsp/multiref/medium_ft.pkl", "medium_ft.pkl"
+    )
+
+    weights = torch.load("medium_ft.pkl")
+    weights["lm_head.weight"] = weights["lm_head.decoder.weight"]
+    weights.pop("lm_head.decoder.weight", None)
+
+    model.load_state_dict(weights)
```
vishalbollu: `model.load_state_dict({'lm_head.weight': weights['lm_head.decoder.weight']})` defining a new dictionary is preferred over  popping a key from a dictionary

examples/pytorch/answer-generator/requirements.txt
```diff
@@ -0,0 +1,3 @@
+torch==1.3.1
```
vishalbollu: Is it necessary to specify torch version? we decided to not have versions for packages that we provide out of box in predictor container right?

PR #604, opened 2019-11-27
Wait for resources to be deleted on cluster update

deliahu:
Checklist:

- [ ] Test end to end manually (e.g. build/push all images, restart operator, and run `cortex deploy --refresh`)
- [ ] Merge to master

### DIFF
--- a/manager/install.sh
+++ b/manager/install.sh
@@ -122,21 +122,53 @@ function main() {
   echo "✓ configured autoscaling"
 
   kubectl -n=cortex delete --ignore-not-found=true daemonset fluentd >/dev/null 2>&1  # Pods in DaemonSets cannot be modified
+  if [ "$(kubectl -n=cortex get pods -l app=fluentd -o json | jq -j '.items | length')" -ne "0" ]; then
+    echo -n "￮ configuring logging "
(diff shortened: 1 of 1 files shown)


manager/install.sh
```diff
@@ -122,21 +122,53 @@ function main() {
...
+  if [ "$(kubectl -n=cortex get pods -l app=fluentd -o json | jq -j '.items | length')" -ne "0" ]; then
+    echo -n "￮ configuring logging "
+    until [ "$(kubectl -n=cortex get pods -l app=fluentd -o json | jq -j '.items | length')" -eq "0" ]; do
+      echo -n "."
+      sleep 2
+    done
+    echo
+  fi
   envsubst < manifests/fluentd.yaml | kubectl apply -f - >/dev/null
   echo "✓ configured logging"
```
vishalbollu: Instead of having:
```
￮ configuring logging .......
✓ configured logging
```

This may make more sense
```
￮ configuring logging ....... ✓
```

So we can have one status change per line. We can remove the outer if

```
    echo -n "￮ configuring logging "
    until [ "$(kubectl -n=cortex get pods -l app=fluentd -o json | jq -j '.items | length')" -eq "0" ]; do
      echo -n "."
      sleep 2
    done
    envsubst < manifests/fluentd.yaml | kubectl apply -f - >/dev/null
    echo -n " ✓"
    echo # for new line
```

PR #705, opened 2019-12-24
fixed misplaced comma in a jupyter notebook file in an example

Ynng:
there's a single misplaced comma in a jupyter notebook file preventing it from being opened and I moved it back... that's it.


ospillinger:
Thank you Kevin!

PR #777, opened 2020-01-21
Bucket region

deliahu:
checklist:


pkg/types/clusterconfig/clusterconfig.go
```diff
@@ -406,12 +407,16 @@ func (cc *Config) Validate(awsClient *aws.Client) error {
 		return ErrorMinInstancesGreaterThanMax(*cc.MinInstances, *cc.MaxInstances)
 	}
 
+	bucketRegion, err := aws.GetBucketRegion(*cc.Bucket)
```
vishalbollu: I think there should be an err != nil check here

PR #792, opened 2020-02-02
Ensure previous logs are never shown after showing newer ones

deliahu:


pkg/operator/operator/logs.go
```diff
@@ -167,6 +168,11 @@ func streamFromCloudWatch(apiName string, podCheckCancel chan struct{}, socket *
 
 			endTime := libtime.ToMillis(time.Now())
 
+			if lastLogTime.Before(lastLogTimeQuery) {
```
vishalbollu: I think this if statement can be removed if this line was deleted https://github.com/cortexlabs/cortex/pull/792/files#diff-0751bf49752315b9e28b82c491299692R162 and the poll period was subtracted here https://github.com/cortexlabs/cortex/pull/792/files#diff-0751bf49752315b9e28b82c491299692R162

PR #800, opened 2020-02-09
Support .cortexignore file to exclude files/directories from cortex project zip

### LINKED ISSUE #723: Support .cortexignore file to exclude files/directories from cortex project zip
`cortex deploy` zips the directory containing the `cortex.yaml` or `CONFIG_FILE`. Implement a .cortexignore to allow users to exclude files and directories from being added to the zip file.

Use `.gitignore` and/or `.dockerignore` for inspiration e.g.

Make sure that the .cortexignore file is not included in the files being zipped.

wingkwong:
closes #723 

---

checklist:

### DIFF
--- a/cli/cmd/deploy.go
+++ b/cli/cmd/deploy.go
@@ -110,7 +110,7 @@ func deploy(configPath string, force bool) {
 
 	cortexIgnorePath := path.Join(projectRoot, ".cortexignore")
 	cortexIgnore, err := os.Open(cortexIgnorePath)
-	excludes := make([]string, 0)
+	var excludes []string
 	if err != nil {
 		if !os.IsNotExist(err) {
 			exit.Error(err)
--- a/cli/cmd/deploy.go
+++ b/cli/cmd/deploy.go
@@ -103,9 +103,6 @@ func deploy(configPath string, force bool) {
 		files.IgnoreHiddenFolders,
 		files.IgnorePythonGeneratedFiles,
 	}
-	if !_flagDeployYes {
-		ignoreFns = append(ignoreFns, files.PromptForFilesAboveSize(_warningFileBytes, "do you want to upload %s (%s)?"))
-	}
 
 	cortexIgnorePath := path.Join(projectRoot, ".cortexignore")
 	if files.IsFile(cortexIgnorePath) {
@@ -114,6 +111,10 @@ func deploy(configPath string, force bool) {
 			exit.Error(err)
 		}
 		ignoreFns = append(ignoreFns, cortexIgnore)
+	}
+
+	if !_flagDeployYes {
+		ignoreFns = append(ignoreFns, files.PromptForFilesAboveSize(_warningFileBytes, "do you want to upload %s (%s)?"))
 	}
 
 	projectPaths, err := files.ListDirRecursive(projectRoot, false, ignoreFns...)
--- a/cli/cmd/deploy.go
+++ b/cli/cmd/deploy.go
@@ -19,6 +19,8 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -106,7 +108,21 @@ func deploy(configPath string, force bool) {
 		ignoreFns = append(ignoreFns, files.PromptForFilesAboveSize(_warningFileBytes, "do you want to upload %s (%s)?"))
 	}
 
-	projectPaths, err := files.ListDirRecursive(projectRoot, false, ignoreFns...)
+	cortexIgnorePath := path.Join(projectRoot, ".cortexignore")
+	cortexIgnore, err := os.Open(cortexIgnorePath)
+	excludes := make([]string, 0)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			exit.Error(err)
+		}
+	} else {
+		excludes, err = files.ReadAllIgnorePatterns(cortexIgnore)
+		if err != nil {
+			exit.Error(err)
+		}
+	}
+
+	projectPaths, err := files.ListDirRecursive(projectRoot, false, excludes, ignoreFns...)
 	if err != nil {
 		exit.Error(err)
 	}
--- a/pkg/lib/files/files.go
+++ b/pkg/lib/files/files.go
@@ -632,7 +632,7 @@ func ReadAllIgnorePatterns(reader io.Reader) ([]string, error) {
 		excludes = append(excludes, pattern)
 	}
 	if err := scanner.Err(); err != nil {
-		return nil, fmt.Errorf("Error reading .dockerignore: %v", err)
+		return nil, errors.Wrap(err, "error reading .cortexignore")
 	}
 	return excludes, nil
 }
--- a/pkg/lib/files/files.go
+++ b/pkg/lib/files/files.go
@@ -509,7 +509,7 @@ func ListDirRecursive(dir string, relative bool, excludes []string, ignoreFns ..
 				path = path[len(dir)+1:]
 			}
 
-			if strings.Contains(path, ".cortexignore") { 
+			if strings.Contains(path, ".cortexignore") {
 				return nil
 			}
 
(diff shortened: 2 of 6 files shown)


pkg/lib/files/files_test.go
```diff
@@ -239,12 +259,89 @@ func TestListDirRecursive(t *testing.T) {
...
+	expected = []string{
+		filepath.Join("2.py"),
+	}
+	require.NoError(t, err)
+	require.ElementsMatch(t, expected, filesListRecursive)
+}
+
+func TestReadAllIgnorePatterns(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "cortexignore-test")
+	if err != nil {
```
deliahu: `require.NoError(t, err)` (here and below) might be a little cleaner, but not a big deal either way, so feel free to make the change or not based on your preference

pkg/lib/files/files_test.go
```diff
@@ -239,12 +259,89 @@ func TestListDirRecursive(t *testing.T) {
...
+
+	di, err = ReadAllIgnorePatterns(diFd)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(di) != 7 {
+		t.Fatalf("Expected 5 entries, got %v", len(di))
+	}
+	if di[0] != "test1" {
```
deliahu: You can use `require.Equal(t, di[0], "test1")` here and below if you'd like

pkg/lib/zip/zip.go
```diff
@@ -193,7 +193,8 @@ func addFileToZip(fileInput *FileInput, zipInput *Input, archive *zip.Writer, ad
 }
 
 func addDirToZip(dirInput *DirInput, zipInput *Input, archive *zip.Writer, addedPaths strset.Set) error {
-	paths, err := files.ListDirRecursive(dirInput.Source, true, dirInput.IgnoreFns...)
+	emptyExcludes := make([]string, 0)
```
deliahu: Will this also work if it's `var emptyExcludes []string` (i.e. if it's `nil`)? It seems to me like it will, just wanted to double check. Also using `nil` will allocate slightly less memory.

pkg/lib/files/files.go
```diff
@@ -506,7 +508,29 @@ func ListDirRecursive(dir string, relative bool, ignoreFns ...IgnoreFn) ([]strin
...
-			fileList = append(fileList, path)
+
+			if strings.Contains(path, ".cortexignore") {
+				return nil
+			}
+
+			if len(excludes) > 0 {
+				ignore := false
+				for _, exclude := range excludes {
+					match, _ := filepath.Match(exclude, path)
```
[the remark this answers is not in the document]
wingkwong: Agree.

pkg/lib/files/files_test.go
wingkwong: Agree. It looks cleaner.

pkg/lib/files/files_test.go
wingkwong: Agree.

pkg/lib/zip/zip.go
wingkwong: Yep. Just tested.

pkg/lib/files/files.go
```diff
@@ -558,3 +582,57 @@ func ReadReqFile(r *http.Request, fileName string) ([]byte, error) {
 	}
 	return fileBytes, nil
 }
+
+// ReadAll reads a .cortexignore file and returns the list of file patterns
+// to ignore. Note this will trim whitespace from each line as well
+// as use GO's "clean" func to get the shortest/cleanest path for each.
+func ReadAllIgnorePatterns(reader io.Reader) ([]string, error) {
```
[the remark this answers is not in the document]
wingkwong: I was following .dockerignore design. Your way seems to be simple tho.

wingkwong:
> @wingkwong thank you for your contribution, this is a great feature!
> 
> I made a few comments on the pull request, but nothing major, and most of them are up to you if you want to make the changes I mentioned. I also pushed a few very small changes to your branch.
> 
> I had two questions regarding the behavior of `.cortexignore`:
> 
> 1. As I was running the code, it seemed to me that it didn't work quite as I expected. I was in the `examples/tensorflow/iris-classifier` directory, I created `.cortexignore` with `*.md`, but `README.md` was still included in the project files (I tested this by printing out `projectPaths` in `cli/cmd/deploy.go`).
> 2. I wasn't able to determine by reading the code where a leading `!` is handled (I didn't see mention of it in [filepath.Match](https://golang.org/pkg/path/filepath/#Match)).
> 
> Please let me know your thoughts!

1.  If *.md is in cortexignore.md
https://github.com/wingkwong/cortex/blob/ebfd26301609042f9b3b6c52e2cc3ee00d66813e/pkg/lib/files/files_test.go#L188

and I expect there is a README.md. The test case will be failed 
image

so it should be able to ignore it.

2. the logic comes from 
https://github.com/wingkwong/cortex/blob/b0546616882bc9d17380c551b106a515ca91896e/pkg/lib/files/files.go#L615

deliahu:
Thanks for addressing the comments!

For 1, I realized the root of the issue was that were were matching against the relative paths, but when `ListDirRecursive()` was called with `relative` set to `false` (as it is in `cli/cmd/deploy.go`), then it was not working as expected.

For 2, I did see this logic, however it seems like this is just for parsing the patterns, but the `!` was not being interpreted by the matcher.

It seemed like even according to [Docker's documentation](https://docs.docker.com/engine/reference/builder/#dockerignore-file), they had to do some workarounds on top of Go's `filepath.Match()`. So I searched for a replacement to the matcher and found this one: https://godoc.org/github.com/denormal/go-gitignore. I switch to it in to your PR, and also tried using the `files.IgnoreFn()` pattern, what do you think of it?

wingkwong:
> Thanks for addressing the comments!
> 
> For 1, I realized the root of the issue was that were were matching against the relative paths, but when `ListDirRecursive()` was called with `relative` set to `false` (as it is in `cli/cmd/deploy.go`), then it was not working as expected.
> 
> For 2, I did see this logic, however it seems like this is just for parsing the patterns, but the `!` was not being interpreted by the matcher.
> 
> It seemed like even according to [Docker's documentation](https://docs.docker.com/engine/reference/builder/#dockerignore-file), they had to do some workarounds on top of Go's `filepath.Match()`. So I searched for a replacement to the matcher and found this one: https://godoc.org/github.com/denormal/go-gitignore. I switch to it in to your PR, and also tried using the `files.IgnoreFn()` pattern, what do you think of it?

I like this approach. It looks more cleaner!

deliahu:
Great! I'll send it to @vishalbollu for review, and then we'll merge it in

pkg/lib/files/files.go
```diff
@@ -347,6 +348,23 @@ func IgnoreSpecificFiles(absPaths ...string) IgnoreFn {
 	}
 }
 
+func GitIgnoreFn(gitIgnorePath string) (IgnoreFn, error) {
```
vishalbollu: I think the .cortexignore file should be excluded from the zip because I don't think the APIs should be redeployed if the .cortexignore file changes without any of the project files changing.

deliahu: `.cortexignore` will be excluded from the zip since we also use `files.IgnoreHiddenFiles` (so all files starting with . will be ignored)

cli/cmd/deploy.go
```diff
@@ -106,6 +107,15 @@ func deploy(configPath string, force bool) {
 		ignoreFns = append(ignoreFns, files.PromptForFilesAboveSize(_warningFileBytes, "do you want to upload %s (%s)?"))
 	}
 
+	cortexIgnorePath := path.Join(projectRoot, ".cortexignore")
+	if files.IsFile(cortexIgnorePath) {
+		cortexIgnore, err := files.GitIgnoreFn(cortexIgnorePath)
+		if err != nil {
+			exit.Error(err)
+		}
+		ignoreFns = append(ignoreFns, cortexIgnore)
```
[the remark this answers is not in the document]
deliahu: Great point! I'll check now, I think we will need to reverse the order

deliahu:
@wingkwong I will merge it now, thank you for your contribution!

PR #815, opened 2020-02-14
Custom autoscaler

### LINKED ISSUE #573: Request-based horizontal pod autoscaling
### Description

Currently, the user must tune an API's CPU request for horizontal pod autoscaling to behave as expected. An approach based on concurrent requests per container may be better (similar to what Knative uses).

This would also make autoscaling for GPU workloads behave more as expected

It may make sense to have both request-based and CPU/GPU-based autoscaling active at the same time, i.e. it will scale when either of the thresholds are met, and won't scale back until both metrics have backed off.

deliahu:
closes #573 

---

checklist:

- [ ] update examples
- [ ] update docs and add any new files to `summary.md` (view in [gitbook](https://cortex.dev/v/master) after merging)


pkg/operator/operator/autoscaler.go
```diff
@@ -0,0 +1,263 @@
...
+		},
+	}
+
+	output, err := config.AWS.CloudWatchMetrics().GetMetricData(&metricsDataQuery)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(output.MetricDataResults[0].Timestamps) == 0 {
+		return nil, nil
```
vishalbollu: maybe return error here instead of nil, like metrics are not available yet?

pkg/operator/operator/autoscaler.go
```diff
@@ -0,0 +1,263 @@
...
+	for i, timeStamp := range output.MetricDataResults[0].Timestamps {
+		if endTime.Sub(*timeStamp) < 20*time.Second {
+			timestampCounter = i
+		} else {
+			break
+		}
+	}
+
+	if timestampCounter == -1 {
+		return nil, nil // no metrics were available in the last 2 tick intervals
```
vishalbollu: maybe return err here indicating metrics were not found?

pkg/operator/operator/autoscaler.go
deliahu: That is the message that is shown when nil is returned, is that ok? Or did you intend to explicitly return an error?

pkg/operator/operator/autoscaler.go
deliahu: same comment as above

PR #827, opened 2020-02-25
Update telemetry

deliahu:


pkg/operator/operator/cron.go
```diff
@@ -67,13 +77,45 @@ func operatorTelemetry() error {
...
+		}
+
+		if info, ok := instanceInfos[instanceInfosKey]; ok {
+			info.Count++
+			continue
+		}
+
+		price := aws.InstanceMetadatas[*config.Cluster.Region][instanceType].Price
+		if isSpot {
+			spotPrice, err := config.AWS.SpotInstancePrice(*config.Cluster.Region, instanceType)
```
vishalbollu: For large clusters with a lot of spot instances, this can get pretty expensive and may cause you to get throttled by AWS if you make so many consecutive requests in such a short period of time. We should cache this information so that we only make a request per instance type.

deliahu: Good thought! It should be cached, since if we already have an `instanceInfo` for the instance type, we increment the counter and continue (line 94)

vishalbollu: Ah oops I see it now.

PR #828, opened 2020-02-27
Use eksct as a package instead of an external dependency

fbertola:
Implemented a simple API for using `eksctl` as a library instead of an external dependency. A generator is used on every command for extracting its flags. No type checking is implemented, as `pflags` does not exports its types easily.

It can be useful for testing purposes or for totally replace how `eksctl` is invoked.


deliahu:
@fbertola This is really cool, we always hypothesized this would be possible, but haven't had the chance to try!

As you mentioned, this is an essential component for eliminating the docker container that the cortex CLI uses to manage the cluster, and building it all directly into the CLI. This would be great because then users wouldn't need docker to run the CLI.

There is one component of the CLI docker container which probably needs to be updated before we can move everything out: istio. Right now, we use helm to install istio. This is no longer the recommended way (there is `istioctl`), and there is a chance we will remove istio soon, but we aren't sure yet.

Until we are able to move everything into the CLI, I'm not sure if it would make sense to merge this into master, since for testing purposes we try to use the `eksctl` CLI like we do in the docker container. So it might make sense to merge it into a different branch on Cortex, or keep it in your repo for now until we can remove the CLI's docker container. What are your thoughts?

fbertola:
Yes I agree. The generator I wrote Is general enough to be used with most cobra-based cli, so in theory I can also create a similar API for `istioctl`. I'll have a look; for now I think you can close this PR and I will write a new one once I will update my fork it with istio. Thanks!

deliahu:
Ok sounds good, thanks!

Issue #831, opened 2020-02-29
AttributeError _thread._local error on inference when using Keras + Tensorflow


RobertLucian:
#### Version

On version `0.13.x`.

#### Description

When using the latest version of Keras (`2.3.1`) and Tensorflow as its backend (`2.x`), the `AttributeError: '_thread._local' object has no attribute 'value'` error is encountered because `_SYMBOLIC_SCOPE.value` does not exist within the Tensorflow backend. The error is encountered when making the prediction and not when the model is loaded.

It must be noted that this specifically happens if the `keras.models.load_model` function is used.

It appears to be due to a conflict between keras/tensorflow and flask (the underlying web framework used in cortex). More on that here:
https://github.com/keras-team/keras/issues/13353

##### Alt. 1

**One very good way of circumventing this is to replace the**
```python
from keras.models import load_model # the one getting the error with
```
with
```python
from tensorflow.keras.models import load_model # the one with which no error is thrown
```

##### Alt. 2

Another possible way to circumvent this is to switch to theano backend by adding `KERAS_BACKEND=theano` environment variable inside the `cortex.yaml` configuration file. But this only works for the CPU inferencing. With this backend, to make it work on the GPU too, `pygpu` has to be installed and this dependecy can only be installed with `conda` or by building it from source, which is anyway riddled with bugs. Don't forget the `theano` project is no longer maintained, which means it's only a temporary fix.

#### Steps to reproduce

1. Deploy an API that loads a keras model (built on `2.3.x`) with tensorflow (`2.x`) using the `keras.models.load_model` function.
1. Make a prediction.
1. See the error using `cortex logs <api name>`.

#### Stack traces

(error output from `cortex logs <api name>`)

```bash
Traceback (most recent call last):
  File "/src/cortex/serve/serve.py", line 86, in predict
    output = predictor_impl.predict(payload)
  File "/mnt/project/predictor.py", line 61, in predict
    pred = self.make_prediction(image)
  File "/mnt/project/predictor.py", line 79, in make_prediction
    pred = self.model.predict(processed)
  File "/usr/local/lib/python3.6/dist-packages/keras/engine/training.py", line 1452, in predict
    if self._uses_dynamic_learning_phase():
  File "/usr/local/lib/python3.6/dist-packages/keras/engine/training.py", line 382, in _uses_dynamic_learning_phase
    not isinstance(K.learning_phase(), int))
[...]

PR #834, opened 2020-03-01
Add keras example API for an autoencoder model

RobertLucian:
This PR contains an example project built with Keras.

The model that I have implemented is an autoencoder. In this use case, the autoencoder is used to filter out the noise from text documents. The `cortex.yaml` config file uses the CPU for inference. 

Here's a before-and-after situation:

Imgur

The model has to be uploaded to `s3://cortex-examples/keras/document-denoiser` directory. Its name has to be `model.h5`. The model can be found [here](https://www.dropbox.com/s/mczfkh2g0eb0shm/model.h5?dl=0) (temporarily).

For the loading function of the model I had to resort to importing it this way
```python
from tensorflow.keras.models import load_model
```
Instead of
```python
from keras.models import load_model
```
Because I encountered the error specified in https://github.com/cortexlabs/cortex/issues/831. This workaround is good for both types of hardware deployments: CPU or GPU.

---

checklist:

- [x] update examples
- [ ] update docs and add any new files to `summary.md` (view in [gitbook](https://cortex.dev/v/master) after merging)


deliahu:
This looks really awesome, thank you for contributing this!

I did a quick pass, and everything looks good on first glance. I will do a more thorough review in the next day or two, and I'll upload the the model to the cortex examples bucket and try running the API. I'll keep you posted on how it goes.

examples/keras/document-denoiser/README.md
```diff
@@ -0,0 +1,51 @@
...
+export IMAGE_URL=https://i.imgur.com/JJLfFxB.png
+```
+
+Now let's take a sample image like this one.
+
+![Imgur](https://i.imgur.com/JJLfFxB.png)
+
+Then run the following piped commands
+```bash
+curl "${ENDPOINT}" -X POST -H "Content-Type: application/json" -d '{"url":"'${IMAGE_URL}'"}' |
```
vishalbollu: If we put the URL in a sample.json
```
{
  "url": "https://i.imgur.com/JJLfFxB.png"
}
```
users would be able to get the exact curl command to run from `cortex get document-denoiser`

examples/keras/document-denoiser/README.md
```diff
@@ -0,0 +1,51 @@
...
+sed 's/"//g' |
+base64 -do prediction.png
```
vishalbollu: Although the -o flag is an option in OSX, I don't think it is an option in ubuntu's base64 binary. To be on the safe side I recommend you replace `base64 -d >> predictor.png` with `base64 -d >> predictor.png`

examples/keras/document-denoiser/README.md
RobertLucian: I can do this. The reason I kept it inline was to make it easy to try another example and to prevent the necessity of having another file (`sample.json`). I don't have a strong opinion here. What do you think?

examples/keras/document-denoiser/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+Here's a collage of input texts and predictions.
+
+![Imgur](https://i.imgur.com/M4Mjz2l.jpg)
+
+*Figure 1 - The dirty documents are on the left side and the cleaned ones are on the right*
+
+## Sample Prediction
+
+### Dependencies
```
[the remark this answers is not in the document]
RobertLucian: This one slipped through. Thanks! It's going out.

examples/keras/document-denoiser/README.md
RobertLucian: Hmm, I agree. I'll do the appropriate modification.

examples/keras/document-denoiser/sample_inference.py
```diff
@@ -0,0 +1,75 @@
...
+
+import click, cv2, requests, pickle, base64, json
+import numpy as np
+
+
+def image_to_png_nparray(image):
+    """
+    Convert numpy image to png numpy vector.
+    """
+    is_success, im_buf_arr = cv2.imencode(".png", image)
```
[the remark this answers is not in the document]
vishalbollu: I guess now that the preprocessing code has been moved to the predict function and the API just accepts a URL the chances of this failing looks to be fairly low.

examples/keras/document-denoiser/predictor.py
```diff
@@ -0,0 +1,83 @@
...
+        s3 = boto3.client("s3")
+        s3.download_file(bucket, os.path.join(key, model_name), model_name)
+
+        # load the model
+        self.model = load_model(model_name)
+
+        # resize shape (width, height)
+        self.resize_shape = tuple(config["resize_shape"])
+
+    def predict(self, payload):
```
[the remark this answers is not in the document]
vishalbollu: The new approach looks great!

examples/keras/document-denoiser/README.md
vishalbollu: We can create the sample.json but leave the instructions as is (using the exported variables). We can remain consistent with the other examples while still making it easy for users to try different images using env variables.

examples/keras/document-denoiser/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This example model cleans text documents of anything that isn't text (aka noise): coffee stains, old wear artifacts, etc. You can inspect the notebook that has been used to train the model [here](trainer.ipynb).
+
+Here's a collage of input texts and predictions.
+
+![Imgur](https://i.imgur.com/M4Mjz2l.jpg)
+
+*Figure 1 - The dirty documents are on the left side and the cleaned ones are on the right*
+
+## Sample Prediction
```
vishalbollu: One of these headers can be deleted

PR #862, opened 2020-03-11
Check cluster image versions match CLI version

deliahu:


pkg/lib/configreader/reader.go
```diff
@@ -891,7 +891,7 @@ func ParseYAMLFile(dest interface{}, validation *StructValidation, filePath stri
 
 	errs := Struct(dest, fileInterface, validation)
 	if errors.HasError(errs) {
-		return errors.WrapAll(errs, filePath)
+		return errors.WrapAll(errs, files.PathRelativeToCWD(filePath))
```
vishalbollu: I think I prefer the abolute path to relative paths because I can copy and paste the path easily without worrying about where I am.

pkg/types/clusterconfig/clusterconfig.go
```diff
@@ -364,6 +384,28 @@ var UserValidation = &cr.StructValidation{
...
+		return image, nil
+	}
+
+	var tag string
+
+	if colonIndex := strings.LastIndex(image, ":"); colonIndex != -1 {
+		tag = image[colonIndex+1:]
+	}
+
+	if tag == "" {
```
vishalbollu: a quick comment to mention that no tag is latest in docker

PR #872, opened 2020-03-15
Update dev docs

RobertLucian:
This does:
* Add instructions on installing the dependencies required for developing cortex.
* Fix the steps on configuring the development environment (about `make registry-create`).
* Fix rerun command for `make devstart`.

Regarding the `rerun` command, I kept getting errors about `-watch` option not existing. Fixed it by using the appropriate options documented [here](https://github.com/alexch/rerun). Hope this is the package that was meant to be used. 

---

checklist:

- [ ] update examples


RobertLucian:
As David has mentioned, `rerun` is a go package and it doesn't coincide with what I had in mind. So the flag discrepancies are explained by this. I rolled it back to use the go's version of this package.

KalivarapuReshma:
@RobertLucian @deliahu 
Hello, Recently I was exploring this project and wanted to play with it, so started installing them. One pain point I observed is there are two versions of awscli available. The `make registry-all` [step](https://github.com/cortexlabs/cortex/blob/master/dev/registry.sh#L27-L35) uses `aws ecr get-login` which is deprecated in V2 of aws-cli. So it started failing at this point, had to uninstall v2 and install V1. It will be helpful if we can call out in the doc that V1 needs to be installed.

Thanks @RobertLucian for updating the doc, all other pain points are covered. Now someone new like me can also set it up seamlessly.

deliahu:
@KalivarapuReshma Thank you for your suggestion! I have updated the PR to include the AWS CLI version.

PR #878, opened 2020-03-18
Check AWS Admin access for cluster up and down commands

deliahu:


cli/cmd/lib_cluster_config.go
```diff
@@ -136,7 +136,7 @@ func getInstallClusterConfig(awsCreds AWSCredentials) (*clusterconfig.Config, er
 		return nil, err
 	}
 
-	awsClient, err := newAWSClient(*clusterConfig.Region, awsCreds)
+	awsClient, err := newAWSClient(*clusterConfig.Region, awsCreds, promptIfNotAdmin)
```
vishalbollu: It may be cleaner convert promptIfNotAdmin to a function rather than pass it as a flag to constructor. Likewise for warnIfNotAdmin.

PR #879, opened 2020-03-18
Cluster status from cloud formation

vishalbollu:
---

checklist:


pkg/types/clusterstate/clusterstate.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if any(statusMap, cloudformation.StackStatusCreateFailed) {
+		return StatusCreateFailed, nil
+	}
+
+	if any(statusMap, cloudformation.StackStatusDeleteFailed) {
+		return StatusDeleteFailed, nil
+	}
+
+	if all(statusMap, cloudformation.StackStatusCreateComplete) {
```
deliahu: Should we also check that statusMap includes the info for the nodegroups? Is it possible if you run this when the cluster stack is complete and before the nodegoup stack creation is initiated, that it will say it's ready?

PR #896, opened 2020-03-23
Update docker errors

deliahu:


cli/cmd/errors.go
```diff
@@ -77,9 +80,28 @@ func ErrorCortexYAMLNotFound() error {
 }
 
 func ErrorDockerDaemon() error {
```
vishalbollu: This name sounds a a bit vague, can you improve the name of this error function and the kind.

cli/cmd/errors.go
```diff
@@ -77,9 +80,28 @@ func ErrorCortexYAMLNotFound() error {
...
+	installMsg := "install it by following the instructions for you operating system: https://docs.docker.com/install"
```
vishalbollu: for you -> for your

PR #899, opened 2020-03-23
Update AWS region validations

deliahu:


pkg/lib/aws/sts.go
```diff
@@ -46,12 +46,12 @@ func (c *Client) AreCredentialsValid() (string, string, bool, error) {
 // Ignores cache, so will re-run on every call to this method
 func (c *Client) CheckCredentials() (string, string, error) {
 	_, _, isValid, err := c.AreCredentialsValid()
-	if !isValid {
```
vishalbollu: good catch

PR #915, opened 2020-03-26
Support multiple types of responses in predict()

### LINKED ISSUE #913: Support byte string return values from predict()
#### Description

1. If the user returns an instance of [`starlette.response`](https://www.starlette.io/responses/#response), send that through unmodified
1. If the user returns a `string` or `bytes`, send that through to the user, and do not set the `media_type` header
1. Otherwise attempt to encode as JSON using the default `json` package, and set the "application/json" media type header

The error message if encoding fails should list the user's three options.

RobertLucian:
Closes #913.

---

checklist:


RobertLucian:
What has to be kept in mind in the back of the mind is that with this method, other protocols for pickle can be used too: 1, 2, 3, and 4 (with the exclusion of 5). 5 cannot be used because the serving container is still based on Python 3.6.x).

Anyhow, the speedup over protocol 0 (in my experiments) is this:

- Protocol 1: ~0.5x speedup.
- Protocol 2: ~0.5x speedup.
- Protocol 3: ~3x speedup.
- Protocol 4: ~3x speedup.
- Protocol 5: ~6x speedup - but is not accessible yet.

So assuming protocol 4, which is 3 times as fast as protocol 0, and no `json.dumps` for it, the serialization process for a ~27 MB `DataFrame` takes about 58 ms. That's a very good improvement over the 820 ms using protocol 0 & `dumps.json`. And even much better if it weren't for pickle (aka using `df.to_dict` and `json.dumps`): ~8000 ms.

pkg/workloads/cortex/serve/serve.py
```diff
@@ -154,17 +154,25 @@ def predict(request: Any = Body(..., media_type="application/json"), debug=False
...
+        response = Response(content=prediction, media_type="text/plain")
+    elif isinstance(prediction, Response):
+        debug_obj("prediction", prediction, debug)
+        response = prediction
+    else:
+        try:
+            json_string = json.dumps(prediction)
+        except Exception as e:
+            raise UserRuntimeException(
+                "the return value of predict() or one of its nested values is not JSON serializable; please return a JSON serializeable object, a bytes object, a string object, or a starlette.response.Response object",
```
RobertLucian: @deliahu I'm thinking that *"the return value of predict() or one of its nested values is not JSON serializable"* is not necessary as it already appears in `str(e)`. Here's a snapshot of what I get without that:
```
cortex.lib.exceptions.UserRuntimeException: error: please return a JSON serializeable object, a bytes object, a string object, or a starlette.response.Response object: Object of type 'DataFrame' is not JSON serializable: runtime exception
```
The message that's coming from `str(e)` is `Object of type 'InsertTypeHere' is not JSON serializable: runtime exception`, so keeping the above string leads to duplication. 

What do you think?

deliahu: I can see both sides on that one... @vishalbollu wrote that error message originally, so I will let him chime in regarding which he prefers

RobertLucian: Hmm, so the only change would be the addition of `including its nested fields` text. I see what you mean about the nested fields. 

I have maybe another idea to reduce the verbosity: we know that whatever exception is thrown, you can be sure it comes from the serialization process, so we can just tailor the error message without appending `str(e)`. Thus, the error message could be this: `"<insert type here>" object is not JSON serializable; please return an object that is JSON serializable (including its nested fields), a bytes object, a string, or a starlette.response.Response object`.

What do you think?

vishalbollu: How would you get the type of the nested value that failed to be serialized? The error message you prepended is fairly close to the output of `str(e)` as you mentioned `Object of type 'DataFrame' is not JSON serializable`

RobertLucian: You wouldn't. I was thinking that the passed object could happen to be something else other than a dict/list - and that alone would be relevant. But yeah, you lose that key information of which nested field failed in the case of a dict/list. 

Maybe place `str(e)` before the custom message? Here's a sample: `cortex.lib.exceptions.UserRuntimeException: error: Object of type 'DataFrame' is not JSON serializable; please return an object that is JSON serializable (including its nested fields), a bytes object, a string, or a starlette.response.Response object: runtime exception`

vishalbollu: Sure ship it!

RobertLucian: Got it! Tested the changes and it's good to go.

PR #916, opened 2020-03-26
Update cleanup() to remove untagged images from ECR

RobertLucian:
I've seen untagged images persisting on ECR while developing. 
This PR purges them on every `make registry-all`/`make registry-dev`.

---

checklist:


deliahu:
This is great, thanks for adding it! This definitely helps keep ECR clean, and beats what I have been doing (which is to delete all the repositories once every month or two and re-create them).

Since it takes some time to list the repositories, `make registry-all` / `make registry-dev` are run frequently in dev, and ECR storage isn't too expensive, I split this out into a separate make command (`make registry-clean`). Does that sound reasonable to you?

RobertLucian:
I agree with this change. I had concerns about this cleanup making the make dev commands take too much time.

PR #946, opened 2020-04-06
Add S3 helpers to download and upload directories

deliahu:


pkg/lib/aws/s3.go
```diff
@@ -295,37 +328,98 @@ func (c *Client) UploadMsgpackToS3(obj interface{}, bucket string, key string) e
 	return c.UploadBytesToS3(msgpackBytes, bucket, key)
 }
 
-func (c *Client) ReadMsgpackFromS3(objPtr interface{}, bucket string, key string) error {
-	msgpackBytes, err := c.ReadBytesFromS3(bucket, key)
+func (c *Client) UploadDirToS3(localDirPath string, bucket string, s3Dir string, ignoreFns ...files.IgnoreFn) error {
+	localPaths, err := files.ListDirRecursive(localDirPath, false, ignoreFns...)
```
vishalbollu, line 332: Have you considered setting relative to true:
```
localPaths, err := files.ListDirRecursive(localDirPath, true, ignoreFns...)
```

The prefix trimming can be removed and in the for loop https://github.com/cortexlabs/cortex/blob/d5eed9d8ed3755e8649af2a62d4a5e7ddc16f929/pkg/lib/aws/s3.go#L346

```
key := filepath.Join(s3Dir, localPath)
``` 

because localPath is already relative?

pkg/lib/aws/s3.go
```diff
@@ -334,144 +428,218 @@ func (c *Client) ReadBytesFromS3Path(s3Path string) ([]byte, error) {
...
+	prefix := s.EnsureSuffix(s3Dir, "/")
+	return c.DownloadPrefixFromS3(bucket, prefix, localDirPath, shouldTrimDirPrefix, maxFiles)
 }
 
-func (c *Client) ListPathDir(s3Path string, maxResults int64) ([]*s3.Object, error) {
-	s3Path = s.EnsureSuffix(s3Path, "/")
-	return c.ListPathPrefix(s3Path, maxResults)
-}
+// if shouldTrimDirPrefix is true, the directory path of prefix will be trimmed when downloading files
+//   e.g. if prefix = "test/dir", "test/" will be trimmed when copying; if prefix = "test/dir/", "test/dir/" will be trimmed
```
vishalbollu, line 458: I think this comment can be improved with an example directory structure in S3 and show how the files will be copied to locally. I think I understand what this is saying but having an example will give me more confidence that I will use this function correctly.

PR #948, opened 2020-04-06
Support different serving images on a per-API basis

### LINKED ISSUE #900: Allow predictor base images to be overridden in API configuration
#### Motivation

* Different APIs may benefit from having different base images
* Allow rolling updates when updating the base image

#### Design

Add a top-level key under predictor `image`
```
- name: iris-classifier
   predictor:
    type: python
    path: predictor.py
    image: ecr/dockerhub link
```

* `image` defaults to cortex default images if user doesn't specify key
* validate that the image path, check access to image (prioritize DockerHub, we may be able to add ECR later)

RobertLucian:
Closes #900, #950. 

This PR includes support for public-accessible docker images (i.e. DockerHub) and ECR images under the cluster's AWS credentials.

Here is a set of errors that could possibly be returned by this new feature:
```
# some common ones that could appear due to setting an invalid value for image or tf_serve_image keys
error: cortex.yaml: api-1: predictor: image: xxxxx.dkr.ecr.eu-central-1.amazonaws.com/cortexlabs/python-serve:custom-d is not accessible
error: cortex.yaml: api-1: predictor: image: xxxxx.dkr.ecr.eu-central-1.amazonaws.com/cortexlabs/python-serve:custom:::dsfds is not a valid docker image resource
error: cortex.yaml: api-1: predictor: image: robertlucian/private-sample-image is not accessible

# some errors that should never happen (mostly internal use)
error: cortex.yaml: api-1: predictor: image: can't extract ECR credentials
error: cortex.yaml: api-1: predictor: image: index out of range: can't extract ECR credentials
error: cortex.yaml: api-1: predictor: image: illegal base64 data at input byte 2504: can't extract ECR credentials
error: cortex.yaml: api-1: predictor: failed to encode docker login credentials: json: <some error>
error: cortex.yaml: api-1: predictor: image: ServerException: failed to retrieve ECR auth token: <insert actual error>
```

---

checklist:


docs/cluster-management/config.md
```diff
@@ -51,6 +51,7 @@ spot: false
 The docker images used by Cortex are listed below. They can be overridden to use custom images by specifying them in your cluster configuration file.
 
 You can follow these [instructions](../deployments/system-packages.md) to build and push custom Docker images to a registry and configure Cortex to use them.
+These images can also be configured on a per-API basis using `image` keys as documented in these [instructions](../deployments/api-configuration.md).
```
deliahu, line 54: As I mentioned below, I think may be best to actually drop the "user-facing" images from cluster configuration. I'd be happy to discuss, or if you agree, do you mind updating these docs accordingly? (I can also take a stab if you'd like)

docs/deployments/api-configuration.md
```diff
@@ -16,6 +16,7 @@ Reference the section below which corresponds to your Predictor type: [Python](#
     path: <string>  # path to a python file with a PythonPredictor class definition, relative to the Cortex root (required)
     config: <string: value>  # arbitrary dictionary passed to the constructor of the Predictor (optional)
     python_path: <string>  # path to the root of your Python folder that will be appended to PYTHONPATH (default: folder containing cortex.yaml)
+    image: <string> # optional image to override the configured cluster serving image (e.g. cortexlabs/python-serve[-gpu])
```
deliahu, line 19: nit: "# docker image to use for the Predictor (default: cortexlabs/python-serve[-gpu])"

docs/deployments/api-configuration.md
```diff
@@ -58,6 +59,8 @@ See additional documentation for [autoscaling](autoscaling.md), [compute](comput
     signature_key: <string>  # name of the signature def to use for prediction (required if your model has more than one signature def)
     config: <string: value>  # arbitrary dictionary passed to the constructor of the Predictor (optional)
     python_path: <string>  # path to the root of your Python folder that will be appended to PYTHONPATH (default: folder containing cortex.yaml)
+    image: <string> # optional image to override the configured cluster api image (e.g. cortexlabs/tf-api)
```
deliahu, line 62: nit: "# docker image to use for the Predictor (default: cortexlabs/tf-api)"

docs/deployments/api-configuration.md
```diff
@@ -99,6 +102,7 @@ See additional documentation for [autoscaling](autoscaling.md), [compute](comput
     model: <string>  # S3 path to an exported model (e.g. s3://my-bucket/exported_model.onnx) (required)
     config: <string: value>  # arbitrary dictionary passed to the constructor of the Predictor (optional)
     python_path: <string>  # path to the root of your Python folder that will be appended to PYTHONPATH (default: folder containing cortex.yaml)
+    image: <string> # optional image to override the configured cluster serving image (e.g. cortexlabs/onnx-serve[-gpu])
```
deliahu, line 105: nit: "# docker image to use for the Predictor (default: cortexlabs/onnx-serve[-gpu])"

docs/deployments/api-configuration.md
```diff
@@ -58,6 +59,8 @@ See additional documentation for [autoscaling](autoscaling.md), [compute](comput
     signature_key: <string>  # name of the signature def to use for prediction (required if your model has more than one signature def)
     config: <string: value>  # arbitrary dictionary passed to the constructor of the Predictor (optional)
     python_path: <string>  # path to the root of your Python folder that will be appended to PYTHONPATH (default: folder containing cortex.yaml)
+    image: <string> # optional image to override the configured cluster api image (e.g. cortexlabs/tf-api)
+    tf_serve_image: <string> # optional image to override the configured cluster serving image (e.g. cortexlabs/tf-serve[-gpu]), which is based on tensorflow/serving image
```
deliahu, line 63: nit: "# docker image to use for the TensorFlow Serving container (default: cortexlabs/tf-serve[-gpu], which is based on tensorflow/serving)"

pkg/lib/aws/errors.go
```diff
@@ -131,3 +132,10 @@ func ErrorReadCredentials() error {
 		Message: "unable to read AWS credentials from credentials file",
 	})
 }
+
+func ErrorECRExtractingCredentials() error {
+	return errors.WithStack(&errors.Error{
+		Kind:    ErrECRExtractingCredentials,
+		Message: "can't extract ECR credentials",
```
deliahu, line 139: nit: "unable to extract ECR credentials"

pkg/lib/configreader/errors.go
```diff
@@ -33,6 +33,7 @@ const (
 	ErrTooShort                      = "configreader.too_short"
 	ErrAlphaNumericDashUnderscore    = "configreader.alpha_numeric_dash_underscore"
 	ErrAlphaNumericDashDotUnderscore = "configreader.alpha_numeric_dash_dot_underscore"
+	ErrValidDockerImage              = "configreader.valid_docker_image"
```
deliahu, line 36: nit:`ErrInvalidDockerImage = "configreader.invalid_docker_image"` (and `ErrorInvalidDockerImage()` below)

pkg/lib/configreader/errors.go
```diff
@@ -121,6 +122,13 @@ func ErrorAlphaNumericDashDotUnderscore(provided string) error {
 	})
 }
 
+func ErrorValidDockerImage(provided string) error {
+	return errors.WithStack(&errors.Error{
+		Kind:    ErrValidDockerImage,
+		Message: fmt.Sprintf("%s is not a valid docker image resource", s.UserStr(provided)),
```
deliahu, line 128: nit: "%s is not a valid docker image path"

pkg/lib/configreader/string.go
```diff
@@ -259,6 +260,12 @@ func ValidateStringVal(val string, v *StringValidation) error {
 		}
 	}
 
+	if v.DockerImage {
```
deliahu, line 263: Since this validation allows empty strings, we have two options for how to make that clear:

1. We can rename the validation to `v.DockerImageOrEmpty`
2. We can convert the struct field to be a pointer (since these validations aren't evaluated on `nil`)

Which do you think is better? I'd be happy to chat about this

pkg/lib/configreader/string.go
```diff
@@ -50,6 +50,7 @@ type StringValidation struct {
 	CastScalar                           bool
 	AllowCortexResources                 bool
 	RequireCortexResources               bool
+	DockerImage                          bool
```
deliahu, line 53: I definitely would not have expected you to catch this: do you mind updating `StringPtrValidation()` and `makeStringValValidation()` in `pkg/lib/configreader/string_ptr.go` to include the new field?

pkg/lib/regex/regex.go
```diff
@@ -40,3 +40,28 @@ var _alphaNumericDashUnderscoreRegex = regexp.MustCompile(`^[a-zA-Z0-9_\-]+$`)
 func IsAlphaNumericDashUnderscore(s string) bool {
 	return _alphaNumericDashUnderscoreRegex.MatchString(s)
 }
+
+// used the evaluated form of
+// https://github.com/docker/distribution/blob/3150937b9f2b1b5b096b2634d0e7c44d4a0f89fb/reference/regexp.go#L68-L70
```
deliahu, line 45: wow, nice!

pkg/lib/regex/regex.go
```diff
@@ -40,3 +40,28 @@ var _alphaNumericDashUnderscoreRegex = regexp.MustCompile(`^[a-zA-Z0-9_\-]+$`)
...
+
```
deliahu, line 46: nit: this new line is not necessary since the comment refers to the variable below

pkg/lib/regex/regex.go
```diff
@@ -40,3 +40,28 @@ var _alphaNumericDashUnderscoreRegex = regexp.MustCompile(`^[a-zA-Z0-9_\-]+$`)
...
+func IsValidDockerImage(s string) bool {
+	return _dockerValidImage.MatchString(s)
+}
+
+var _ecrPattern = regexp.MustCompile(
+	`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr(\-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?`,
+)
+
+func IsValidECR(s string) bool {
```
deliahu, line 65: nit: `IsValidECRImage()`

pkg/lib/regex/regex.go
```diff
@@ -40,3 +40,28 @@ var _alphaNumericDashUnderscoreRegex = regexp.MustCompile(`^[a-zA-Z0-9_\-]+$`)
...
+	return _ecrPattern.MatchString(s)
```
deliahu, line 66: Should this function also check `IsValidDockerImage()`?

pkg/lib/regex/regex_test.go
```diff
@@ -0,0 +1,493 @@
...
+	"testing"
+)
+
+type regexpMatch struct {
+	input string
+	match bool
+	subs  []string
+}
+
+func TestAlphaNumericDashDotUnderscoreRegex(t *testing.T) {
```
deliahu, line 30: Thanks for adding tests for these!

pkg/operator/operator/errors.go
```diff
@@ -51,6 +51,7 @@ const (
 	ErrNoAvailableNodeComputeLimit        = "operator.no_available_node_compute_limit"
 	ErrCortexPrefixedEnvVarNotAllowed     = "operator.cortex_prefixed_env_var_not_allowed"
 	ErrAPINotDeployed                     = "operator.api_not_deployed"
+	ErrInvalidOverriddenImage             = "operator.invalid_overridden_image"
```
deliahu, line 54: nit: `ErrDockerImageInaccessible` (and associated variables)

pkg/operator/operator/validations.go
```diff
@@ -452,6 +473,23 @@ func validatePredictor(predictor *userconfig.Predictor, projectFileMap map[strin
...
+	images := []string{
+		predictor.Image,
+		predictor.TFServeImage,
+	}
+	keys := []string{
+		userconfig.ImageKey,
+		userconfig.TFServeImageKey,
+	}
+	// do it in one go to avoid generating credentials twice
+	key, err := validateOverridenImages(images, keys)
```
deliahu, line 485: `validateOverridenImages()` should do the appropriate wrapping of images/keys, so here we can just return `err` without wrapping. After a long design discussion that we had a few months ago (which I'm happy to elaborate more on!), we decided that the "rule" is: assume that any information that's passed into the function is already included in the returned error message.

pkg/types/userconfig/api.go
```diff
@@ -249,6 +251,12 @@ func (predictor *Predictor) UserStr() string {
 	if predictor.PythonPath != nil {
 		sb.WriteString(fmt.Sprintf("%s: %s\n", PythonPathKey, *predictor.PythonPath))
 	}
+	if predictor.Image != "" {
+		sb.WriteString(fmt.Sprintf("%s: %s\n", ImageKey, predictor.Image))
+		if predictor.TFServeImage != "" {
```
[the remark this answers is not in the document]
RobertLucian: Yes, that's totally possible. In this case, `if predictor.TFServeImage` must be taken out of that branch. Doing it now.

pkg/types/userconfig/api.go
```diff
@@ -249,6 +251,12 @@ func (predictor *Predictor) UserStr() string {
 	if predictor.PythonPath != nil {
 		sb.WriteString(fmt.Sprintf("%s: %s\n", PythonPathKey, *predictor.PythonPath))
 	}
+	if predictor.Image != "" {
```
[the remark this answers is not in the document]
RobertLucian: Hmm I see what you mean.

pkg/lib/aws/errors.go
RobertLucian: Done!

docs/deployments/api-configuration.md
RobertLucian: Done!

docs/deployments/api-configuration.md
RobertLucian: Done!

docs/deployments/api-configuration.md
RobertLucian: Done!

docs/deployments/api-configuration.md
RobertLucian: Done!

pkg/lib/configreader/errors.go
RobertLucian: Done!

pkg/lib/configreader/errors.go
RobertLucian: Done!

pkg/lib/regex/regex.go
RobertLucian: Well, it's not actually necessary. When this function is called, we already know that the string we're passing to this function is a valid docker path, because, by that point, it would have already gone through the validation phase. What do you think?

RobertLucian: So `IsValidECR` goes hand in hand with `IsValidDockerImage`. On the other hand, we could add another 3rd function `IsValidECRImage` to use the previous two. What do you think?

pkg/operator/operator/k8s_specs.go
```diff
@@ -107,11 +107,18 @@ func tfAPISpec(api *spec.API, prevDeployment *kapps.Deployment) *kapps.Deploymen
 	}
 
 	servingImage := config.Cluster.ImageTFServe
+	apiImage := config.Cluster.ImageTFAPI
```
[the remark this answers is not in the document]
RobertLucian: I agree with keeping it simple. And since you no longer have to update the cluster if the user wants a different base image, then the incentive to use `cluster.yaml` to update the images is less significant.

pkg/operator/operator/validations.go
RobertLucian: > assume that any information that's passed into the function is already included in the returned error message

This! This is a very important bit. Thanks for giving me the perspective. And it makes sense now that I think of it. Consider it done.

pkg/operator/operator/errors.go
RobertLucian: Done!

docs/cluster-management/config.md
RobertLucian: Yep, I'll update these as soon as the whole backend is taken care of.

pkg/lib/configreader/string.go
RobertLucian: I suggest renaming it to `v.DockerImageOrEmpty` because I see another field following the same pattern: `AlphaNumericDashDotUnderscoreOrEmpty`. What do you say?

pkg/lib/configreader/string.go
RobertLucian: Yep, updated it. Let me know if I haven't caught something.

pkg/lib/regex/regex.go
RobertLucian: Done!

pkg/lib/configreader/string.go
deliahu: Sounds good!

pkg/operator/operator/validations.go
```diff
@@ -744,3 +786,53 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
...
+		dockerECRAuthConfig.Username = ECRAuthConfig.Username
+		dockerECRAuthConfig.Password = ECRAuthConfig.AccessToken
+		dockerECRAuthConfig.ServerAddress = ECRAuthConfig.ProxyEndpoint
+	}
+
+	ECRAuth, err := dockerlib.EncodeAuthConfig(dockerECRAuthConfig)
+	if err != nil {
+		return err
+	}
+	noAuth, _ := dockerlib.EncodeAuthConfig(dockertypes.AuthConfig{})
```
deliahu, line 851: This could be moved into a global variable in lib/docker, e.g. dockerlib.NoAuth, do you think that makes sense?

pkg/operator/operator/validations.go
```diff
@@ -744,3 +786,53 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
 
 	return virtualServices, maxMem, err
 }
+
+func validateOverridenImages(images, keys []string) error {
+	images = slices.RemoveEmpties(images)
```
deliahu: This may not be desired now that the values will be set to the defaults. However, I do think we can skip the accessibility validation if the image is in the cortexlabs dockerhub domain (we should still check the version in the tag though)

pkg/operator/operator/validations.go
```diff
@@ -744,3 +786,53 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
...
+		if err != nil {
+			return err
+		}
+
+		dockerECRAuthConfig.Username = ECRAuthConfig.Username
+		dockerECRAuthConfig.Password = ECRAuthConfig.AccessToken
+		dockerECRAuthConfig.ServerAddress = ECRAuthConfig.ProxyEndpoint
+	}
+
+	ECRAuth, err := dockerlib.EncodeAuthConfig(dockerECRAuthConfig)
```
deliahu, line 847: nit: `ecrAuth`

pkg/operator/operator/validations.go
```diff
@@ -744,3 +786,53 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
 
 	return virtualServices, maxMem, err
 }
+
+func validateOverridenImages(images, keys []string) error {
```
deliahu: I think this can be simplified a bit, although I'm trying to wrap my head around exactly how. I've typed out two ideas, realized there is a good reason you designed it this way, and this is now my 3rd version of my proposal, which is more modest of a change than my first two ideas :) :

Perhaps just a single loop over images/keys would be more clear. We could declare a pointer variable for `ecrAuth` outside of the loop, and if we end up needing it in the loop, we can check if it's still nil, and if it is, then generate it. Making the code to generate it into a helper function may also help with clarity. What do you think?

pkg/operator/operator/validations.go
```diff
@@ -744,3 +786,53 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
...
+	for i, image := range images {
+		if regex.IsValidECR(image) {
+			ECRImages[i] = true
+		}
+	}
+
+	ECRAuthConfig := aws.ECRAuthConfig{}
+	dockerECRAuthConfig := dockertypes.AuthConfig{}
+	if slices.HasTrue(ECRImages) {
+		tokenOutput, err := aws.GetECRAuthToken()
```
deliahu, line 833: This runs in the operator, and currently we do not include ECR permissions on the list of required permissions [here](https://www.cortex.dev/cluster-management/security#operator). Do you know if we need any permissions other than `ecr:GetAuthorizationToken` to run the accessibility check? If not, here's what I'm thinking we could do:

* Add `ecr:GetAuthorizationToken` to the list of operator credentials we ask for
* **If the image is on ECR:** is it possible to distinguish between the repo not existing vs it existing and the IAM user not having access to it? If so, the former should be an error, whereas the latter I think we can ignore silently, since the nodes themselves have an IAM policy (that eksctl creates) which is used to pull the ECR images. We may want to check that the account ID for the ECR path matches that of the IAM user, since that will avoid a weird case where the IAM user has access to a different account's repo (the role the eksctl created will not have access to repos in other accounts). See `GetCachedAccountID()` in `lib/aws/sts.go` for getting the account ID.
  * an alternative to this is to use the AWS go client to check if the ECR image exists, rather than the docker client. The may give us more control over the error handling
* **If the image is not on ECR:** any error is an error that we should propagate

pkg/operator/operator/validations.go
RobertLucian: Yeah, this is already removed.

pkg/operator/operator/validations.go
deliahu: This may not be necessary information for this code, but just FYI: the role that eksctl creates for pulling the images has the `AmazonEC2ContainerRegistryReadOnly` policy, which has these permissions:

```text
"Effect": "Allow",
"Action": [
    "ecr:GetAuthorizationToken",
    "ecr:BatchCheckLayerAvailability",
    "ecr:GetDownloadUrlForLayer",
    "ecr:GetRepositoryPolicy",
    "ecr:DescribeRepositories",
    "ecr:ListImages",
    "ecr:DescribeImages",
    "ecr:BatchGetImage",
    "ecr:GetLifecyclePolicy",
    "ecr:GetLifecyclePolicyPreview",
    "ecr:ListTagsForResource",
    "ecr:DescribeImageScanFindings"
],
"Resource": "*"
```

pkg/lib/aws/ecr.go
```diff
@@ -0,0 +1,88 @@
...
+			switch aerr.Code() {
+			case ecr.ErrCodeServerException:
+				return result, errors.Wrap(aerr, ecr.ErrCodeServerException, "failed to retrieve ECR auth token")
+			case ecr.ErrCodeInvalidParameterException:
+				return result, errors.Wrap(
+					aerr,
+					ecr.ErrCodeInvalidParameterException,
+					"failed to retrieve ECR auth token",
+				)
+			default:
```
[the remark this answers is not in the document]
RobertLucian: Good thoughts. Since you only care whether the get authorization function has failed or not, it's just better to return the result and the wrapped error.

pkg/lib/aws/ecr.go
```diff
@@ -0,0 +1,88 @@
...
+	}))
+	svc := ecr.New(sess)
+	input := &ecr.GetAuthorizationTokenInput{}
+	result, err := svc.GetAuthorizationTokenWithContext(context.Background(), input)
+
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case ecr.ErrCodeServerException:
+				return result, errors.Wrap(aerr, ecr.ErrCodeServerException, "failed to retrieve ECR auth token")
```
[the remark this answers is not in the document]
RobertLucian: Responded in the other related comment.

pkg/lib/aws/ecr.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+	return result, nil
+}
+
+func ExtractECRAuthConfigFromTokenOutput(auth *ecr.GetAuthorizationTokenOutput) (ECRAuthConfig, error) {
+	var authConfig ECRAuthConfig
+	if len(auth.AuthorizationData) < 1 {
+		return authConfig, errors.Wrap(ErrorECRExtractingCredentials(), "index out of range")
```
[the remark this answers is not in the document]
RobertLucian: Gotcha.

pkg/lib/aws/ecr.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return authConfig, errors.Wrap(ErrorECRExtractingCredentials(), err.Error())
+	}
+	credentialsString := string(credentials)
+	splitCredentials := strings.Split(credentialsString, ":")
+	if len(splitCredentials) != 2 {
+		return authConfig, ErrorECRExtractingCredentials()
+	}
+
+	authConfig.Username = splitCredentials[0]
```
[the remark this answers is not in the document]
RobertLucian: Good point.

pkg/operator/operator/validations.go
```diff
@@ -25,15 +25,20 @@ import (
 	"github.com/cortexlabs/cortex/pkg/lib/aws"
 	"github.com/cortexlabs/cortex/pkg/lib/cast"
 	cr "github.com/cortexlabs/cortex/pkg/lib/configreader"
+	dockerlib "github.com/cortexlabs/cortex/pkg/lib/docker"
```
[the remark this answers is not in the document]
RobertLucian: So we're gonna leave it this way?

docs/cluster-management/config.md
RobertLucian: Yep. You're right.

pkg/lib/aws/ecr.go
```diff
@@ -0,0 +1,88 @@
...
+			}
+		} else {
+			return result, errors.Wrap(err, "failed to retrieve ECR auth token")
+		}
+	}
+
+	return result, nil
+}
+
+func ExtractECRAuthConfigFromTokenOutput(auth *ecr.GetAuthorizationTokenOutput) (ECRAuthConfig, error) {
```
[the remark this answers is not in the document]
RobertLucian: I'm thinking that keeping them separate improves the code readability and maybe there will be a future case where `GetAuthorizationTokenOutput` is required and not `ECRAuthConfig`. Let me know what you think after seeing the newly modified versions of both functions.

pkg/operator/operator/validations.go
RobertLucian: Yeah, I could see it being used multiple times.

pkg/operator/operator/validations.go
RobertLucian: Done!

RobertLucian:
@deliahu 
This will also close #950. Here's what the errors can look like now:
```
# robertlucian/sample-image is a private image
error: cpy_pred.yaml: iris-classifier-2: predictor: image: robertlucian/sample-image is not accessible: docker client:
Error response from daemon: errors:
denied: requested access to the resource is denied
unauthorized: authentication required

# cluster has no read permissions on ECR
error: cpy_pred.yaml: iris-classifier-2: predictor: image: XXXXXXXXXXXX.dkr.ecr.eu-central-1.amazonaws.com/mock-repository/hello-world is not accessible: docker client:
Error response from daemon: denied: User: arn:aws:iam::XXXXXXXXXXXX:user/cloud-user is not authorized to perform: ecr:BatchGetImage on resource: arn:aws:ecr:eu-central-1:XXXXXXXXXXXX:repository/mock-repository/hello-world with an explicit deny

# docker image does not exist
error: cpy_pred.yaml: iris-classifier-2: predictor: image: cortexlabs/python-serve-non-existent:master is not accessible: docker client:
Error response from daemon: errors:
denied: requested access to the resource is denied
unauthorized: authentication required

# ECR image does not exist
error: cpy_pred.yaml: iris-classifier-2: predictor: image: XXXXXXXXXXXX.dkr.ecr.eu-central-1.amazonaws.com/cortexlabs/python-serve-non-existent is not accessible: docker client:
Error response from daemon: manifest unknown: Requested image not found

# Non-existent registry (checkout ddkr vs dkr)
error: cpy_pred.yaml: iris-classifier-2: predictor: image: XXXXXXXXXXXX.ddkr.ecr.eu-central-1.amazonaws.com/cortexlabs/python-serve:custom is not accessible: docker client:
Error response from daemon: Get https://XXXXXXXXXXXX.ddkr.ecr.eu-central-1.amazonaws.com/v2/: dial tcp: lookup XXXXXXXXXXXX.ddkr.ecr.eu-central-1.amazonaws.com: no such host

# Non-matching registry account IDs
error: cpy_pred.yaml: iris-classifier-2: predictor: image: AWS registry account ID YYYYYYYYYYYY doesn't match operator's account ID XXXXXXXXXXXX

# Version mismatch (at the tag level)
error: cpy_pred.yaml: iris-classifier-2: predictor: image: the specified image (cortexlabs/python-serve:0.15) has a tag (0.15) which does not match the version of your CLI (master); please update the image tag, remove the image from the API config file (to use the default value), or update your CLI by following the instructions at https://www.cortex.dev/install

# Unsupported field for Python/ONNX Predictors
error: cpy_pred.yaml: iris-classifier-2: predictor: tf_serve_image is not a supported field for the python predictor type

# Should never happen
error: cpy_pred.yaml: iris-classifier-2: predictor: image: failed to encode docker login credentials: json: <insert error>

# Should never happen
error: cpy_pred.yaml: iris-classifier-2: predictor: image: failed to retrieve ECR auth token: <insert error>

# Should never happen
error: cpy_pred.yaml: iris-classifier-2: predictor: image: unable to extract ECR credentials: <insert error>
```

One thing to keep in mind: when trying to inspect images from other registries other than ECR, the docker client can't tell whether the image simply does not exist or it's private. I think this stems from the fact that the docker client has to be authenticated in the said registry. This does work with ECR though, because the client has the authentification token. 

One a different note, there's no need to use the AWS ECR client to inspect images. As it appears, it's generally expected to use AWS ECR API to get the authentification token and then use the docker client instead. The AWS ECR API can also be used to modify policies and stuff like that, but  not things pertaining to the docker client. 

Here are the validations done for a generic docker image:
1. Check tag version.
1. Inspect image w/ docker client. If error, propagate.
Here are the validations done for an ECR image:
1. Check tag version.
1. Match account IDs.
1. Inspect image w/ docker client. If error, propagate.

Regarding the errors returned upon inspecting the image, I saw the errors can get quite large (verbosy like in the examples above) and so I decided to place them on a new line right after adding the `docker client: ` message to all the other wrapped messages. I'm also thinking that this is a way to put emphasis on the docker client error and separate it from other cortex errors.

pkg/operator/operator/validations.go
```diff
@@ -25,15 +26,20 @@ import (
...
 	"github.com/cortexlabs/cortex/pkg/lib/errors"
 	"github.com/cortexlabs/cortex/pkg/lib/k8s"
 	"github.com/cortexlabs/cortex/pkg/lib/parallel"
 	"github.com/cortexlabs/cortex/pkg/lib/pointer"
+	"github.com/cortexlabs/cortex/pkg/lib/regex"
+	// "github.com/cortexlabs/cortex/pkg/lib/slices"
```
deliahu, line 35: this can be deleted

pkg/operator/operator/validations.go
```diff
@@ -446,12 +470,25 @@ func validatePredictor(predictor *userconfig.Predictor, projectFileMap map[strin
 		if err := validateTensorFlowPredictor(predictor); err != nil {
 			return err
 		}
+		if err := validateDockerImagePath(predictor.TFServeImage); err != nil {
+			if errCause := errors.Cause(err); errCause != nil {
+				return errors.Wrap(errCause, userconfig.TFServeImageKey, err.Error())
```
deliahu, line 475: I believe the original error message should be included in `err` without the wrap being necessary, or was there a different reason you extracted the cause here? (also below)

pkg/operator/operator/validations.go
```diff
@@ -744,3 +789,54 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
 
 	return virtualServices, maxMem, err
 }
+
+func validateDockerImagePath(image string) error {
+	if _, err := cr.ValidateImageVersion(image); err != nil {
```
deliahu, line 794: I'm considering short-circuiting the check if the image is one of the defaults, what do you think? We could have a `strset` which includes the defaults. It's a little bit of extra state we'd have to keep consistent, but that would be the most common code path through here, so it might be nice to make that faster.

pkg/operator/operator/validations.go
```diff
@@ -744,3 +789,54 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
...
+
+	dockerAuth := dockerlib.NoAuth
+	if regex.IsValidECRURL(image) {
+		operatorID, _, err := config.AWS.GetCachedAccountID()
+		if err != nil {
+			return err
+		}
+		registryID := aws.GetAccountIDFromECRURL(image)
+
+		if strings.Compare(operatorID, registryID) != 0 {
```
deliahu, line 806: nit: `if operatorID != registryID` is simpler

pkg/operator/operator/errors.go
```diff
@@ -260,4 +262,20 @@ func ErrorAPINotDeployed(apiName string) error {
 		Kind:    ErrAPINotDeployed,
 		Message: fmt.Sprintf("%s is not deployed", apiName), // note: if modifying this string, search the codebase for it and change all occurrences
 	})
+
+}
+
+func ErrorRegistryAccountIDMismatch(regID, opID string) error {
+	return errors.WithStack(&errors.Error{
+		Kind:    ErrRegistryAccountIDMismatch,
+		Message: fmt.Sprintf("AWS registry account ID %s doesn't match operator's account ID %s", regID, opID),
```
deliahu, line 271: Since we may start calling this from the CLI with local, how do you feel about this:

```go
fmt.Sprintf("registry account ID (%s) doesn't match your AWS account ID (%s), and using an ECR registry in a different AWS account is not supported", regID, opID)
```

pkg/operator/operator/validations.go
```diff
@@ -744,3 +789,54 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
...
+			return ErrorRegistryAccountIDMismatch(registryID, operatorID)
+		}
+
+		ecrAuthConfig, err := config.AWS.GetECRAuthConfig()
+		if err != nil {
+			return err
+		}
+
+		dAuth, err := dockerlib.EncodeAuthConfig(dockertypes.AuthConfig{
```
deliahu, line 815: This is totally up to you, it's more of a style/clarity consideration: since `err` is already defined, you could do:

```go
dockerAuth, err = dockerlib.EncodeAuthConfig(dockertypes.AuthConfig{
```

and remove `dockerAuth = dAuth`

pkg/operator/operator/validations.go
```diff
@@ -744,3 +789,54 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
...
+		return err
+	}
+
+	if err := dockerlib.CheckImageAccessible(client, image, dockerAuth); err != nil {
+		// the concrete type of docker client errors is string
+		// it's just better to put them on a new line
+		// because they are quite verbosy
+		return ErrorDockerImageInaccessible(
+			image,
+			goerrs.New("docker client:\n"+err.Error()),
```
deliahu, line 837: I think it would be cleaner to pass in `err` and do the string manipulation in `ErrorDockerImageInaccessible()`

pkg/operator/operator/validations.go
```diff
@@ -744,3 +789,54 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
...
+		dockerAuth = dAuth
+	}
+
+	client, err := dockerclient.NewEnvClient()
+	if err != nil {
+		return err
+	}
+
+	if err := dockerlib.CheckImageAccessible(client, image, dockerAuth); err != nil {
+		// the concrete type of docker client errors is string
```
deliahu, line 832: This comment can be removed or moved to `ErrorDockerImageInaccessible()`

pkg/operator/operator/errors.go
```diff
@@ -260,4 +262,20 @@ func ErrorAPINotDeployed(apiName string) error {
...
+	})
+}
+
+func ErrorDockerImageInaccessible(image string, cause error) error {
+	return errors.WithStack(&errors.Error{
+		Kind:    ErrDockerImageInaccessible,
+		Message: fmt.Sprintf("%s is not accessible", image),
```
deliahu, line 278: the error message of the original error should be included in `Message`, here is [an example](https://github.com/cortexlabs/cortex/blob/master/pkg/lib/aws/errors.go#L70) (the only difference is here you probably want more than the first line, so you can use `errors.Message(err)` to get the whole thing (that helper pretty much just makes sure there's no leading/trailing whitespace)

pkg/operator/operator/validations.go
```diff
@@ -744,3 +789,54 @@ func getValidationK8sResources() ([]kunstructured.Unstructured, *kresource.Quant
...
+		// it's just better to put them on a new line
+		// because they are quite verbosy
+		return ErrorDockerImageInaccessible(
```
deliahu, line 835: Since there is no wrapping here, `CheckImageAccessible()` can return the `ErrorDockerImageInaccessible` error

pkg/lib/aws/ecr.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+type ECRAuthConfig struct {
+	Username      string
+	AccessToken   string
+	ProxyEndpoint string
+}
+
+func (c *Client) GetECRAuthToken() (*ecr.GetAuthorizationTokenOutput, error) {
+	svc := c.ECR()
```
deliahu, line 35: nit: you can remove this line and use `c.ECR()` below

pkg/lib/aws/ecr.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	return result, nil
+}
+
+func (c *Client) GetECRAuthConfig() (ECRAuthConfig, error) {
+	tokenOutput, err := c.GetECRAuthToken()
+	if err != nil {
+		return ECRAuthConfig{}, err
+	}
+	authData := tokenOutput.AuthorizationData[0]
```
deliahu, line 48: Is `len(tokenOutput.AuthorizationData) > 0` guaranteed?

pkg/lib/aws/ecr.go
```diff
@@ -0,0 +1,74 @@
...
+	return ECRAuthConfig{
+		Username:      splitCredentials[0],
+		AccessToken:   splitCredentials[1],
+		ProxyEndpoint: *authData.ProxyEndpoint,
+	}, nil
+}
+
+func GetAccountIDFromECRURL(path string) string {
+	if regex.IsValidECRURL(path) {
+		split := strings.Split(path, ".")
```
deliahu, line 69: nit: I think `return strings.Split(path, ".")[0]` wouldn't be less clear, and is more concise

pkg/lib/configreader/validators.go
```diff
@@ -149,3 +150,26 @@ func DurationParser(v *DurationValidation) func(string) (interface{}, error) {
 		return d, nil
 	}
 }
+
+func ValidateImageVersion(image string) (string, error) {
```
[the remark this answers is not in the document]
RobertLucian: Yeah, I think that makes sense. I think that's okay.

pkg/lib/configreader/errors.go
```diff
@@ -345,3 +355,10 @@ func ErrorCortexResourceNotAllowed(resourceName string) error {
 		Message: fmt.Sprintf("@%s: cortex resource references (which start with @) are not allowed in this context", resourceName),
 	})
 }
+
+func ErrorImageVersionMismatch(image string, tag string) error {
```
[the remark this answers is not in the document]
RobertLucian: Yep, done.

pkg/lib/configreader/validators.go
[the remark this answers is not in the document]
RobertLucian: That's fine to me. One thing to keep in mind: I had to create a wrapper function for `ValidateImageVersion(image, version string)` inside `clusterconfig/clusterconfig.go` called `validateImageVersion(image string)` because this function is still required as a validator to other cluster-configured images and because the validator expects a single parameter.

pkg/lib/configreader/errors.go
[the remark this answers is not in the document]
RobertLucian: Yep, I agree. It's done.

pkg/lib/configreader/errors.go
```diff
@@ -345,3 +355,10 @@ func ErrorCortexResourceNotAllowed(resourceName string) error {
...
+	return errors.WithStack(&errors.Error{
+		Kind:    ErrImageVersionMismatch,
+		Message: fmt.Sprintf("the specified image (%s) has a tag (%s) which does not match the version of your CLI (%s); please update the image tag, remove the image from the API config file (to use the default value), or update your CLI by following the instructions at https://www.cortex.dev/install", image, tag, consts.CortexVersion),
```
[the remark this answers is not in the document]
RobertLucian: That looks good to me. Yeah, the error message should be a bit vaguer since it can also happen in the operator.

pkg/operator/operator/errors.go
```diff
@@ -260,4 +262,20 @@ func ErrorAPINotDeployed(apiName string) error {
 		Kind:    ErrAPINotDeployed,
 		Message: fmt.Sprintf("%s is not deployed", apiName), // note: if modifying this string, search the codebase for it and change all occurrences
 	})
+
```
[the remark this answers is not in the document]
RobertLucian: Done!

pkg/operator/operator/validations.go
RobertLucian: Gone!

pkg/operator/operator/validations.go
RobertLucian: Agreed. Done!

pkg/operator/operator/errors.go
RobertLucian: I like it - mainly because it abstracts the operator, so this could lead to less confusion.

pkg/operator/operator/validations.go
RobertLucian: Good call yeah. Makes sense. Done! Otherwise, if the error has to be returned from some other place, the string manipulation could be forgotten.

pkg/operator/operator/validations.go
RobertLucian: Moved!

pkg/lib/aws/ecr.go
RobertLucian: Done!

pkg/lib/aws/ecr.go
RobertLucian: Done!

pkg/lib/aws/ecr.go
RobertLucian: Judging from what I've seen in the AWS SDK docs, yes, that appears to be guaranteed unless you get an error. Apparently, there has to be a 1:1 mapping between what's being requested and what you get back, otherwise, you get a non-200 response. At least that's my understanding. It doesn't help that the docs are not overly clear about that. Here are the docs: https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_GetAuthorizationToken.html

What do you think of checking the length and if not return an error?

pkg/lib/docker/docker.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to encode docker login credentials")
+	}
+	registryAuth := base64.URLEncoding.EncodeToString(encoded)
+	return registryAuth, nil
+}
+
+func CheckImageAccessible(c *dockerclient.Client, dockerImage, registryAuth string) error {
+	_, err := c.DistributionInspect(context.Background(), dockerImage, registryAuth)
+	return err
```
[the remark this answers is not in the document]
RobertLucian: I wouldn't do this because then, you wouldn't get the full picture of why the operator couldn't pull the images - like in the errors I have referenced above this. I've double-checked, and the docker client errors come from `c.DistributionInspect`. Here's an example of an error that shouldn't be muted:
```
Error response from daemon: denied: User: arn:aws:iam::XXXXXXXXXXXX:user/cloud-user is not authorized to perform: ecr:BatchGetImage on resource: arn:aws:ecr:eu-central-1:XXXXXXXXXXXX:repository/mock-repository/hello-world with an explicit deny
```

What do you say?

pkg/operator/operator/errors.go
RobertLucian: Agreed. Keeping it consistent with all other error functions makes sense.

pkg/operator/operator/validations.go
RobertLucian: Modified it as you have suggested it in the other comment (aka [example](https://github.com/cortexlabs/cortex/blob/master/pkg/lib/aws/errors.go#L70)).

pkg/operator/operator/validations.go
RobertLucian: I like this suggestion. It can speed up the validation process when default paths are used.

cli/cmd/lib_manager.go
```diff
@@ -240,6 +240,13 @@ func runManagerUpdateCommand(entrypoint string, clusterConfig *clusterconfig.Con
 			"CORTEX_TELEMETRY_SEGMENT_WRITE_KEY=" + os.Getenv("CORTEX_TELEMETRY_SEGMENT_WRITE_KEY"),
 			"CORTEX_CLUSTER_CONFIG_FILE=" + mountedConfigPath,
 			"CORTEX_CLUSTER_WORKSPACE=" + clusterWorkspace,
+			"CORTEX_IMAGE_PYTHON_SERVE=" + consts.DefaultImagePythonServe,
```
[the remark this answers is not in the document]
RobertLucian: Apparently, yeah. I would actually suggest keeping it as it is, because:
1. It's consistent with how all other images are referenced in the other manifests.
1. There's only one place where the default paths are defined: in `consts` package.

docs/deployments/api-configuration.md
```diff
@@ -44,7 +45,7 @@ Reference the section below which corresponds to your Predictor type: [Python](#
     max_unavailable: <string | int>  # maximum number of replicas that can be unavailable during an update; can be an absolute number, e.g. 5, or a percentage of desired replicas, e.g. 10% (default: 25%)
 ```
 
-See additional documentation for [autoscaling](autoscaling.md), [compute](compute.md), and [prediction monitoring](prediction-monitoring.md).
+See additional documentation for [autoscaling](autoscaling.md), [compute](compute.md), and [prediction monitoring](prediction-monitoring.md) and [system packages](system-packages.md).
```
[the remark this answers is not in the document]
RobertLucian: My thinking is this: since 2 new fields can appear in an API, `image` and `tf_serve_image`, you (the user) might want to have a look at how custom images can be built so that they can override the default ones with theirs. What do you think?

cli/cmd/lib_manager.go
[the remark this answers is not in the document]
deliahu: Makes sense! I'll resolve this comment now

docs/deployments/api-configuration.md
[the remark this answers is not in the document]
deliahu: Ah I see, yes that makes sense. Perhaps we can make it more clear, and replace the hyperlink text with "custom images" or "overriding API images" or similar, what do you think?

pkg/lib/docker/docker.go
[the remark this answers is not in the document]
deliahu: Yes, I am not proposing to hide any error messages, just to wrap, so maybe in this case it would be:

```
unable to access docker image XXXXX.dkr.ecr.us-west-2.amazonaws.com/XXXXX/XXXXX:latest: Error response from daemon: denied: User: arn:aws:iam::XXXXXXXXXXXX:user/cloud-user is not authorized to perform: ecr:BatchGetImage on resource: arn:aws:ecr:eu-central-1:XXXXXXXXXXXX:repository/mock-repository/hello-world with an explicit deny
```

given that, what do you think?

pkg/lib/aws/ecr.go
deliahu: Yeah, we have been bit by something like this with the AWS SDK before, so I think we might as well do the check if it's straightforward

pkg/lib/docker/docker.go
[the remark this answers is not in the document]
deliahu: Actually I see that this message is already incorporated in `ErrorImageInaccessible()`, correct?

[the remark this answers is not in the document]
RobertLucian: Oh yeah, definitely. They are already incorporated. Should I resolve this discussion?

docs/deployments/api-configuration.md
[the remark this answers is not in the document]
RobertLucian: Yes, I think that would be better. Name them like you said "overriding API images". I'm gonna do it now.

pkg/lib/aws/ecr.go
RobertLucian: I did a `return ..., ErrorECRExtractingCredentials()` if `AuthorizationData` is empty. Do you think that's not enough, or should the error's message be more clear?

deliahu: I think that is sufficient, given we do not expect this to happen. Mostly I was motivated to avoid the nil pointer.

deliahu:
@RobertLucian Thanks again for your great work on this!

PR #949, opened 2020-04-06
Log unexpected TensorFlow Serving gRPC errors

deliahu:


pkg/workloads/cortex/lib/client/tensorflow.py
```diff
@@ -117,11 +117,15 @@ def get_signature_def(stub):
...
             sigmap = json_format.MessageToDict(signature_def_map)
             return sigmap["signatureDef"]
-        except:
-            if i > 6:
-                cx_logger().warn(
-                    "unable to read model metadata - model is still loading, retrying..."
-                )
+        except Exception as e:
+            if isinstance(e, grpc.RpcError) and e.code() == grpc.StatusCode.UNAVAILABLE:
+                if i > 6:  # only start logging this after 1 minute
```
vishalbollu, line 122: since time.sleep(5) I believe this is 30 seconds and not one minute

deliahu: Yes, good catch, thank you!

PR #972, opened 2020-04-18
added rolling update to deamonset

### LINKED ISSUE #630: Perform rolling update of daemonsets on cluster update
### Description

Update one daemonset at a time and wait for completion before continuing to not trigger operator cluster autoscaling (in `manager/install.sh`)

Docs: https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/

tthebst:
updated cloudwatch-agent-statsd, fluentd and nvidia to perform rolling update in cluster update command. 

closes #630

---

checklist:

- [ x] run `make test` and `make lint`
- [x ] test manually (i.e. build/push all images, restart operator, and re-deploy APIs)
- [ x] update examples
- [x ] update docs and add any new files to `summary.md` (view in [gitbook](https://cortex.dev/v/master) after merging)
- [x ] cherry-pick into release branches if applicable
- [ x] alert the dev team if the dev environment changed

### DIFF
--- a/manager/manifests/fluentd.yaml
+++ b/manager/manifests/fluentd.yaml
@@ -176,7 +176,6 @@ spec:
     metadata:
       labels:
         app: fluentd
-        hallo: dummy
     spec:
       serviceAccountName: fluentd
       priorityClassName: fluentd
--- a/manager/manifests/fluentd.yaml
+++ b/manager/manifests/fluentd.yaml
@@ -168,10 +168,15 @@ metadata:
   name: fluentd
   namespace: default
 spec:
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
   template:
     metadata:
       labels:
         app: fluentd
+        hallo: dummy
     spec:
       serviceAccountName: fluentd
       priorityClassName: fluentd
--- a/manager/manifests/nvidia.yaml
+++ b/manager/manifests/nvidia.yaml
@@ -28,7 +28,6 @@ metadata:
   name: nvidia-device-plugin-daemonset
   namespace: kube-system
 spec:
-
   selector:
     matchLabels:
       name: nvidia-device-plugin-ds
--- a/manager/manifests/nvidia.yaml
+++ b/manager/manifests/nvidia.yaml
@@ -28,11 +28,14 @@ metadata:
   name: nvidia-device-plugin-daemonset
   namespace: kube-system
 spec:
+
   selector:
     matchLabels:
       name: nvidia-device-plugin-ds
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
   template:
     metadata:
       # This annotation is deprecated. Kept here for backward compatibility
--- a/manager/manifests/statsd.yaml
+++ b/manager/manifests/statsd.yaml
@@ -66,7 +66,6 @@ spec:
     metadata:
       labels:
         name: cloudwatch-agent-statsd
-        hallo: dummy
     spec:
       priorityClassName: statsd
       containers:
--- a/manager/manifests/statsd.yaml
+++ b/manager/manifests/statsd.yaml
@@ -55,6 +55,10 @@ metadata:
   name: cloudwatch-agent-statsd
   namespace: default
 spec:
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
   selector:
     matchLabels:
       name: cloudwatch-agent-statsd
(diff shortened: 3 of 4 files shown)


manager/manifests/fluentd.yaml
```diff
@@ -168,10 +168,15 @@ metadata:
   name: fluentd
   namespace: default
 spec:
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
```
deliahu, line 174: Is there a reason you set this to 1, rather than 0 (or left the default)? I imagine that with 0, it would ensure that there is always one replica of this pod running on each node (although I am not as familiar with how this setting works with daemonsets vs deployments).

The current behavior is to take down the old pod before bringing up the new one (i.e. `maxUnavailable` is essentially 100% now). I prefer the path you've taken; my guess is that with 2 fluentd pods we might get logs duplicated in cloudwatch, which seems less bad than dropping logs if we had a period with 0 fluentd pods running.

I think this applies to the other two as well; does setting `maxUnavailable: 0` make sense to you, or was there a reason you went with `maxUnavailable: 1`?

manager/manifests/fluentd.yaml
```diff
@@ -168,10 +168,15 @@ metadata:
...
   template:
     metadata:
       labels:
         app: fluentd
+        hallo: dummy
```
deliahu, line 179: This is probably left over from testing :)

manager/manifests/nvidia.yaml
```diff
@@ -28,11 +28,14 @@ metadata:
   name: nvidia-device-plugin-daemonset
   namespace: kube-system
 spec:
+
```
deliahu, line 31: nit: extra new line

manager/manifests/statsd.yaml
```diff
@@ -55,13 +55,18 @@ metadata:
...
+    rollingUpdate:
+      maxUnavailable: 1
   selector:
     matchLabels:
       name: cloudwatch-agent-statsd
   template:
     metadata:
       labels:
         name: cloudwatch-agent-statsd
+        hallo: dummy
```
deliahu, line 69: This is probably left over from testing :)

manager/manifests/fluentd.yaml
```diff
@@ -168,10 +168,15 @@ metadata:
...
     spec:
       serviceAccountName: fluentd
       priorityClassName: fluentd
```
deliahu, line 182: (I'm putting this comment here because GitHub won't let me put it lower): did you try testing the rolling update after changing the container image? If not I can, just checking!

manager/manifests/fluentd.yaml
tthebst: I looked at this too. But i think Kubernetes does not support maxUnavailable:0 (See: https://stackoverflow.com/questions/54116493/update-a-daemonset-without-downtime-like-a-deployment). maxUnavailable: 1 is the default and I included it because of visability.

manager/manifests/fluentd.yaml
tthebst: will correct

manager/manifests/nvidia.yaml
tthebst: will correct

manager/manifests/statsd.yaml
tthebst: will correct

manager/manifests/fluentd.yaml
tthebst: yes i tried the rolling update and it looks like it works.

manager/manifests/fluentd.yaml
deliahu: Sounds good, thanks for looking into that!

deliahu:
@tthebst Thanks for addressing my comments! I'll try this out on my end soon, and then I'll merge it in.

tthebst:
@deliahu Thank you for your feedback. These are my first real contributions to an opensource project. I like the project and will try to find some good other addressable issues

deliahu:
@tthebst sounds great! All of the features/improvements/bugs that are on our radar are listed in our [GitHub Project](https://github.com/cortexlabs/cortex/projects/1). That list may be a bit overwhelming, so (as you probably noticed when you searched for this issue) we have created the ["good first issue" label](https://github.com/cortexlabs/cortex/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) and added it to some of our issues. That said, the other ones are all fair game too, and some of them may be good first issues as well - we plan on making a pass on all of our issues and seeing if we can add the label to more of them.

Feel free to let us know which ones sound interesting to you. We'd be able to advise on which one might be best to start with based on your interests and the scope of the task, as well as point you in the right direction (since some of our descriptions are vague and and some of our issues need additional design thought).

PR #992, opened 2020-04-30
Add slim predictor images

### LINKED ISSUE #781: Make slim versions of serving images
#### Description

Serving images such as `python-serve`, `python-serve-gpu` have a lot of python packages and system libraries to make onboarding easy. 

Have a slim version that only contains the python packages and system libraries to run Cortex serving.

#### Motivation

Having unnecessary libraries in docker images can increase storage costs and also network costs.

deliahu:
closes #781


images/python-predictor-gpu/Dockerfile
```diff
@@ -44,11 +33,50 @@ RUN /opt/conda/bin/conda create -n env -c conda-forge python=$PYTHONVERSION pip
...
+                scipy==1.4.1 \
+                six==1.14.0 \
+                statsmodels==0.11.1 \
+                sympy==1.5.1 \
+                tensorflow-gpu==2.1.0 \
+                tensorflow-hub==0.7.0 \
+                torch==1.4.0 \
+                torchvision==0.5.0 \
+                xgboost==1.0.2" \
+    || true
```
vishalbollu, line 79: If there is a failure in one of the pip installs or apt-get would this `|| true` swallow the failure and still move to the next line? I think something along the lines of `test "${SLIM}" == "true" || (apt-get .... )" might be safer. I am not sure what if shell boolean logic short circuits conditions

Makefile
```diff
@@ -77,10 +77,37 @@ operator-stop:
...
 
 registry-all:
-	@./dev/registry.sh update
+	@./dev/registry.sh update all
+registry-all-local:
+	@./dev/registry.sh update all --skip-push
+registry-all-slim:
+	@./dev/registry.sh update all --include-slim
+registry-all-slim-local:
+	@./dev/registry.sh update all --include-slim --skip-push
```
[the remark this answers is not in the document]
deliahu: yes

images/python-predictor-gpu/Dockerfile
deliahu: Good catch!!!

PR #1023, opened 2020-05-05
Add environment to deploy output sample commands

deliahu:


cli/cmd/deploy.go
```diff
@@ -241,16 +241,22 @@ func didAllResultsError(results []schema.DeployResult) bool {
...
+func getAPICommandsMessage(results []schema.DeployResult, envName string) string {
 	apiName := "<api_name>"
 	if len(results) == 1 {
 		apiName = results[0].API.Name
 	}
 
+	defaultEnv := getDefaultEnv(_generalCommandType)
+	var envArg string
+	if envName != defaultEnv {
+		envArg = " --env " + envName
```
vishalbollu, line 253: It may be easier to read the code below if the leading space was removed e.g. `envArg = "--env " + envName`

PR #1024, opened 2020-05-06
Add zsh completion

### LINKED ISSUE #1020: Completion doesn't work with zsh
#### Version
0.16.0

#### Description
The cli client completion doesn't work in `zsh` shell owning to some compatibility issues in the `cortex completion` script. Throws an error -> `bad option: -t`

#### Steps to reproduce

1. Run source ` source <(cortex completion)` in a zsh shell

#### Expected behavior

cli autocomplete should work with zsh.

#### Actual behavior

Since the script failed, no autocomplete with cli in zsh

#### Screenshots
image

deliahu:
closes #1020


cli/cmd/completion.go
```diff
@@ -53,6 +37,60 @@ else
...
+                   apt update && apt install -y bash-completion  # you may need sudo
+                2) open ~/.bashrc and uncomment the bash completion section, or add this:
+                   if [ -f /etc/bash_completion ] && ! shopt -oq posix; then . /etc/bash_completion; fi
+                3) log out and back in, or close your terminal window and reopen it
+
+    zsh:
+        option 1:
+            add this to ~/.zshrc:
+                source <(cortex completion zsh)
+            if that failed, you can try adding this to ~/.zshrc above the source command:
```
vishalbollu, line 69: if that failed, you can try adding this line above the source command in ~/.zshrc:

cli/cmd/completion.go
```diff
@@ -53,6 +37,60 @@ else
...
+			_rootCmd.GenZshCompletion(os.Stdout)
+			fmt.Print("alias cx='cortex'\n\n")
+
+			// https://github.com/spf13/cobra/pull/887
+			// https://github.com/corneliusweig/rakkess/blob/master/cmd/completion.go
+			// https://github.com/GoogleContainerTools/skaffold/blob/master/cmd/skaffold/app/cmd/completion.go
+			// https://github.com/spf13/cobra/issues/881
+			// https://github.com/asdf-vm/asdf/issues/266
+			fmt.Println("if compquote '' 2>/dev/null; then _cortex; else compdef _cortex cortex; fi")
+		}
```
vishalbollu, line 94: maybe we should add a default case saying autocomplete for this shell is not supported?

deliahu: Great idea!

PR #1044, opened 2020-05-08
Search all environments for AWS credential defaults

### LINKED ISSUE #726: Use default environment keys as default in cortex configure prompts for new environments
#### Description

Default values in ```cortex configure -e <env>``` prompts have the following precedence:

Currently: 
1. Previous keys used in <env> if it exists
2. Blank

Suggested order:
1. Previous keys used in <env> if they exist
2. Keys from the default environment if they exist
3. Blank

#### Motivation

Prevent users from copying the same key multiple times if they have multiple operator endpoints but they are all deployed using the same key.

deliahu:
closes #726

---

checklist:

### DIFF
--- a/cli/cmd/lib_cli_config.go
+++ b/cli/cmd/lib_cli_config.go
@@ -641,6 +641,25 @@ func getEnvConfigDefaults(envName string) cliconfig.Environment {
 	if defaults.AWSRegion == nil && os.Getenv("AWS_REGION") != "" {
 		defaults.AWSRegion = pointer.String(os.Getenv("AWS_REGION"))
 	}
+
+	if defaults.AWSAccessKeyID == nil && defaults.AWSSecretAccessKey == nil {
+		// search other envs for credentials (favoring the env named "aws", or the last entry in the list)
+		regionWasNil := defaults.AWSRegion == nil
+		cliConfig, _ := readCLIConfig()
+		for _, env := range cliConfig.Environments {
+			if env.AWSAccessKeyID != nil && env.AWSSecretAccessKey != nil {
+				defaults.AWSAccessKeyID = env.AWSAccessKeyID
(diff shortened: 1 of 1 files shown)


cli/cmd/lib_cli_config.go
```diff
@@ -641,6 +641,25 @@ func getEnvConfigDefaults(envName string) cliconfig.Environment {
...
+	if defaults.AWSAccessKeyID == nil && defaults.AWSSecretAccessKey == nil {
+		// search other envs for credentials (favoring the env named "aws", or the last entry in the list)
+		regionWasNil := defaults.AWSRegion == nil
+		cliConfig, _ := readCLIConfig()
+		for _, env := range cliConfig.Environments {
+			if env.AWSAccessKeyID != nil && env.AWSSecretAccessKey != nil {
+				defaults.AWSAccessKeyID = env.AWSAccessKeyID
+				defaults.AWSSecretAccessKey = env.AWSSecretAccessKey
+			}
+			if regionWasNil && env.AWSRegion != nil {
```
vishalbollu, line 654: nit: can `defaults.AWSRegion == nil` be inlined here instead of using `regionWasNil`?

deliahu: I separated it to guard against possibly having multiple envs with the region set. Since the loop runs to completion (I think taking the last config makes the most sense), if we clobber `defaults.AWSAccessKeyID` and `defaults.AWSSecretAccessKey` from a previous iteration, I would also want to clobber region too (to make sure region "tags along" with the credentials).

Right now I don't think this will happen because currently local is the only provider that has region, and there can be only one local, so therefore there can be only one region in `cli.yaml`. So I can in-line it if you think it's better, but I was just trying to be on the safe side (keeping the region attached to the credentials).

Let me know what you think!

vishalbollu: SGTM

PR #1049, opened 2020-05-08
Update default CLI env after env delete

### LINKED ISSUE #1048: Update the default CLI environment after cortex env delete

deliahu:
closes #1048

---

checklist:

### DIFF
--- a/cli/cmd/lib_cli_config.go
+++ b/cli/cmd/lib_cli_config.go
@@ -828,6 +828,8 @@ func removeEnvFromCLIConfig(envName string) error {
 		return err
 	}
 
+	prevDefault := getDefaultEnv(_generalCommandType)
+
 	var updatedEnvs []*cliconfig.Environment
 	deleted := false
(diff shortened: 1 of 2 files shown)


cli/cmd/lib_cli_config.go
```diff
@@ -838,12 +840,19 @@ func removeEnvFromCLIConfig(envName string) error {
 		updatedEnvs = append(updatedEnvs, env)
 	}
 
-	if deleted == false && envName != types.LocalProviderType.String() {
+	if !deleted {
+		if envName == types.LocalProviderType.String() {
```
vishalbollu, line 844: If nothing was deleted and env is local, it likely means that cli config got into an erroneous state somehow. We should allow `writeCLIConfig(cliConfig)` because it will add the expected local config before writing.

deliahu: Agreed. Fixed!

PR #1059, opened 2020-05-15
Read and validate from buckets in any regions

vishalbollu:
---

checklist:

### DIFF
--- a/pkg/lib/aws/aws.go
+++ b/pkg/lib/aws/aws.go
@@ -43,6 +43,14 @@ func NewFromCreds(region string, accessKeyID string, secretAccessKey string) (*C
 	return New(region, creds)
 }
 
+func NewFromClientS3Path(s3Path string, awsClient *Client) (*Client, error) {
+	if !awsClient.IsAnonymous {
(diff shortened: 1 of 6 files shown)


pkg/lib/aws/aws.go
```diff
@@ -43,6 +44,14 @@ func NewFromCreds(region string, accessKeyID string, secretAccessKey string) (*C
 	return New(region, creds)
 }
 
+func NewFromClientS3Path(s3Path string, awsClient *Client) (*Client, error) {
+	if !awsClient.IsAnonymous {
+		return NewFromCredsS3Path(s3Path, *awsClient.AccessKeyID(), *awsClient.SecretAccessKey())
```
deliahu, line 49: Should we be safe and check that `awsClient.AccessKeyID()` and  `awsClient.SecretAccessKey()` are not nil? And if so, what do you think we should we do in that case?

pkg/lib/aws/aws.go
```diff
@@ -43,6 +44,14 @@ func NewFromCreds(region string, accessKeyID string, secretAccessKey string) (*C
...
+	}
+
+	return NewAnonymousClient()
```
deliahu, line 52: Should this still be in the region of the s3 path?

PR #1081, opened 2020-05-23
Add multi-model endpoints guide

### LINKED ISSUE #986: Add a guide for multi-model endpoints
#### Description

Multi-model endpoints are possible using the Python Predictor, but we don't yet have an example of how to do this.

https://github.com/cortexlabs/cortex/issues/619 tracks adding support for a model cache, so that all models need not be able to fit in memory at the same time

deliahu:
closes #986

### DIFF
--- a/docs/guides/multi-model.md
+++ b/docs/guides/multi-model.md
@@ -37,6 +37,7 @@ transformers==2.9.*
 ```python
 import torch
 from transformers import pipeline
+from starlette.responses import JSONResponse
 
 
 class PythonPredictor:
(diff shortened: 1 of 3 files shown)


docs/guides/multi-model.md
```diff
@@ -0,0 +1,108 @@
...
+        self.summarizer = pipeline(task="summarization", device=device)
+
+    def predict(self, query_params, payload):
+        if query_params.get("model") == "sentiment":
+            return self.analyzer(payload["text"])[0]
+        elif query_params.get("model") == "summarizer":
+            summary = self.summarizer(payload["text"])
+            return summary[0]["summary_text"]
+        else:
+            return {"error": f"unknown model: {query_params.get('model')}"}
```
vishalbollu, line 57: The error code would still be 200 for this error. We can showcase returning a starlette exception here or raise an exception.

PR #1083, opened 2020-05-27
Use the function defined by the library instead, remove unnecessary assignments

zouyee:
Use the function defined by the library instead, remove unnecessary assignments, because the object obtained according to the version is specified

closes #<issue ID>

---

checklist:

- [x] update examples

### DIFF
--- a/pkg/lib/k8s/configmap.go
+++ b/pkg/lib/k8s/configmap.go
@@ -18,11 +18,10 @@ package k8s
 
 import (
 	"github.com/cortexlabs/cortex/pkg/lib/errors"
-
 	kcore "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	kmeta "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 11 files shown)


zouyee:
> Regarding setting `TypeMeta` for the k8s objects, I remember having to add this for some reason. This was a while ago and now I forget (I should have added a comment!), but for it was necessary at least at the time that I added it, so I am hesitant to remove it. Since the client version matches the `TypeMeta` that we are setting, is it safe to leave it as is? If that's the case that would be my preference for now.

When we use the specified version of the interface, such as `k8s.io/api/core/v1`, then the version obtained `TypeMeta` must be v1, I think it may not be necessary to define it.

deliahu:
> When we use the specified version of the interface, such as `k8s.io/api/core/v1`, then the version obtained `TypeMeta` must be v1, I think it may not be necessary to define it.

Yes I think you are correct, although for some reason at some point it was necessary for me, and since I don't think it hurts, I'm inclined to leave it as it was.

Using the labels package is definitely an improvement, so I'll go ahead and merge that now! Thanks again!

PR #1102, opened 2020-06-01
Fix shell variable in development makefile

RobertLucian:
Any make command that uses the `shell` variable will fail because there's no `shell` variable declared, but a `SHELL` variable with capital letters. Also, every invocation of the shell variable has the dollar sign missing.

These are some examples that will fail: `cluster-up`, `cluster-down`, `cluster-configure`.

---

checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,27 +49,27 @@ kubectl:
 cluster-up:
 	@$(MAKE) registry-all
 	@$(MAKE) cli
-	@kill $(shell pgrep -f make) >/dev/null 2>&1 || true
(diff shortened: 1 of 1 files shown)


deliahu:
@RobertLucian after merging https://github.com/cortexlabs/cortex/pull/1103, I was not able to reproduce the error you mentioned. Also I see this which seems to imply that it should work: https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html

Do you mind pulling the latest from master, and checking if it's still failing for you?

RobertLucian:
@deliahu with #1103 in, it no longer fails. Since this is no longer an issue, I'm closing this PR.

PR #1106, opened 2020-06-02
Surface pod creation errors

### LINKED ISSUE #955: Surface pod creation errors
#### Motivation

Some errors (e.g. ImagePullBackoff) do not get shown in `cortex get`, which can lead to confusion as to why an API has errored (there may be no API logs).

zouyee:
closes #955

---

checklist:

- [x] update examples

### DIFF
--- a/cli/cmd/get.go
+++ b/cli/cmd/get.go
@@ -40,13 +40,18 @@ import (
 	"github.com/cortexlabs/cortex/pkg/lib/telemetry"
 	libtime "github.com/cortexlabs/cortex/pkg/lib/time"
 	"github.com/cortexlabs/cortex/pkg/lib/urls"
+	"github.com/cortexlabs/cortex/pkg/operator/config"
 	"github.com/cortexlabs/cortex/pkg/operator/schema"
 	"github.com/cortexlabs/cortex/pkg/types"
 	"github.com/cortexlabs/cortex/pkg/types/metrics"
 	"github.com/cortexlabs/cortex/pkg/types/spec"
 	"github.com/cortexlabs/cortex/pkg/types/status"
 	"github.com/cortexlabs/cortex/pkg/types/userconfig"
+
 	"github.com/spf13/cobra"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


deliahu:
@zouyee thanks for looking into this!

Since we don't require the CLI user to have kubectl configured, it would be better to implement this in the Cortex operator.

A better approach would be to add additional status code(s) in the Cortex operator to capture additional error cases. All of the status logic is implemented in [operator/status.go](https://github.com/cortexlabs/cortex/blob/master/pkg/operator/operator/status.go) and [GetPodStatus()](https://github.com/cortexlabs/cortex/blob/master/pkg/lib/k8s/pod.go#L140) in k8s/pod.go. It should be possible to add new statuses [here](https://github.com/cortexlabs/cortex/blob/master/pkg/lib/k8s/pod.go#L43), add them to [SubReplicaCounts](https://github.com/cortexlabs/cortex/blob/master/pkg/types/status/status.go#L32), and then update [addPodToReplicaCounts()](https://github.com/cortexlabs/cortex/blob/master/pkg/operator/operator/status.go#L126) and [getStatusCode()](https://github.com/cortexlabs/cortex/blob/master/pkg/operator/operator/status.go#L163) accordingly.

Does that make sense? Let us know if you have any questions!

zouyee:
According to the design concept of k8s, it is not recommended to summarize the status when pull image or oom into the state, but the reason why the pod presents the pending state

deliahu:
I see what you're saying... however there is a nuance here, which is that the status we show in `cortex get` is a combination of the statuses of all the API replicas. This is computed in [`getStatusCode()`](https://github.com/cortexlabs/cortex/blob/master/pkg/operator/operator/status.go#L163), and is why we took the approach of wrapping the k8s statuses (since we eventually have to end up with a single descriptive status for the entire k8s deployment). Does that make sense?

deliahu:
@zouyee I have incorporated your change into https://github.com/cortexlabs/cortex/pull/1167 (and https://github.com/cortexlabs/cortex/commit/239316ae7cb0596d87abe90419c86ea2edd9014d), so I will go ahead and close this one. I mostly just moved the logic you added into the operator, and added a new API status for it. Thanks for your contribution!

Issue #890, opened 2020-03-21
Support multi-model endpoints for TensorFlow and ONNX predictors


deliahu:
Currently muti-model endpoints is only supported for the Python predictor type. Support should be added for the TensorFlow and ONNX types.

#### Potential design

Add a `models` key to the API configuration, which is a map of model name -> model path. Then, models could be accessed by the TensorFlow/ONNX client by name, e.g. `client.predict("model-1")`

#### Motivation

Share GPU resources across multiple models

RobertLucian:
@deliahu taking this one alongside https://github.com/cortexlabs/cortex/issues/619. Finally, guides for each predictor type will be required as described in https://github.com/cortexlabs/cortex/issues/986.

PR #1107, opened 2020-06-02
Add multi-model endpoint feature to TF & ONNX predictors

RobertLucian:
Closes https://github.com/cortexlabs/cortex/issues/890.

---

checklist:

- [x] update examples


pkg/workloads/cortex/lib/client/tensorflow.py
```diff
@@ -24,47 +24,73 @@
...
         Returns:
             dict: TensorFlow Serving response converted to a dictionary.
         """
-        validate_model_input(self._input_signature, model_input)
-        prediction_request = create_prediction_request(
-            self._signature, self._signature_key, model_input
-        )
-        response_proto = self._stub.Predict(prediction_request, timeout=300.0)
-        return parse_response_proto(response_proto)
+        if model in self._model_indexes:
```
deliahu, line 64: use the same pattern we came up with for onnx

pkg/workloads/cortex/lib/client/tensorflow.py
```diff
@@ -24,47 +24,73 @@
...
         self._stub = prediction_service_pb2_grpc.PredictionServiceStub(channel)
-        self._signature = get_signature_def(self._stub)
-        parsed_signature_key, parsed_signature = extract_signature(self._signature, signature_key)
-        self._signature_key = parsed_signature_key
-        self._input_signature = parsed_signature
-        self._tf_serving_url = tf_serving_url
 
-    def predict(self, model_input):
+        # in the same order as self._models
+        self._signatures = get_signature_defs(self._stub, models)
```
deliahu, line 47: I think we can convert this to map, and then we can remove `Model.get_model_indexes`

pkg/workloads/cortex/lib/client/tensorflow.py
```diff
@@ -24,47 +24,73 @@
...
         return self._stub
 
     @property
-    def input_signature(self):
-        return self._input_signature
+    def input_signatures(self):
+        input_signatures = {}
+        for name, sign in zip(self._model_names, self._input_signatures):
+            input_signatures[name] = sign
+        return input_signatures
```
deliahu, line 93: can return this when it's a map

pkg/workloads/cortex/lib/client/tensorflow.py
```diff
@@ -24,47 +24,73 @@
...
+        )
+        self._signature_keys = parsed_signature_keys
+        self._input_signatures = parsed_signatures
+
+    def predict(self, model_input, model="default"):
         """Validate model_input, convert it to a Prediction Proto, and make a request to TensorFlow Serving.
 
         Args:
             model_input: Input to the model.
+            model: Model to use when multiple models are deployed per each API.
```
deliahu, line 59: change to `model_name`

pkg/workloads/cortex/lib/client/onnx.py
```diff
@@ -17,48 +17,62 @@
...
+                numpy_type = ONNX_TO_NP_TYPE.get(meta.type, meta.type)
+                metadata[meta.name] = {"shape": meta.shape, "type": numpy_type}
+            self._input_signatures[model.name] = metadata
+
+    def predict(self, model_input, model="default"):
         """Validate input, convert it to a dictionary of input_name to numpy.ndarray, and make a prediction.
 
         Args:
             model_input: Input to the model.
+            model: Model to use when multiple models are deployed per each API.
```
deliahu, line 53: change to `model_name`

pkg/types/spec/validations.go
```diff
@@ -828,6 +919,19 @@ func FindDuplicateNames(apis []userconfig.API) []userconfig.API {
 	return nil
 }
 
+func checkDuplicateModelNames(modelResources []*userconfig.ModelResource) error {
+	var names []string
```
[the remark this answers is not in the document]
RobertLucian: I'd still go with a slice because the names have to be extracted from the struct and then you also have to keep a record of the duplicate model. Then you'd have to compare the lengths.

RobertLucian:
@deliahu the models used in this PR can be [found here](https://www.dropbox.com/s/hsqv87ogwptnltr/multi-model-endpoint-models.zip?dl=0). They have to be uploaded to the public Cortex S3 bucket. Inside that zip, you will find the following models:

1. ResNet50 model in SavedModel and ONNX format.
1. MobileNet model in ONNX format.
1. ShuffleNet model in ONNX format.

Next, there was a bug with the ONNX predictor where if you used a model with a dynamically sized input signature, the `predict` method would then try to reshape the model input to `(1,1,1)` - where each 1 was the replacement value for an unknown shape (that would otherwise be permissible). This is now fixed. This is the commit https://github.com/cortexlabs/cortex/pull/1107/commits/ad6ec6f1f40168b6a150de046e9c243b40840b82 in question.

Following this, here's my answer to @vishalbollu 's question: assuming there are at least 2 threads per worker and a single worker, when a request for a time-consuming model is being processed (say 5-10 seconds), if a new request lands on the other thread and it's for a model that takes little time to run the inference on, the new request will be parallelized and hence, will finish first.

pkg/workloads/cortex/lib/type/predictor.py
```diff
@@ -21,24 +21,29 @@
...
         self.provider = provider
         self.type = kwargs["type"]
         self.path = kwargs["path"]
         self.python_path = kwargs.get("python_path")
         self.config = kwargs.get("config", {})
         self.env = kwargs.get("env")
 
+        self.model_dir = model_dir
         self.models = []
         if kwargs.get("model"):
```
deliahu, line 38: Check if we can remove this section (I pushed the change already)

pkg/workloads/cortex/lib/type/api.py
```diff
@@ -158,3 +160,18 @@ def prediction_metrics(self, dimensions, prediction_value):
                 "Dimensions": dimensions,
                 "Value": float(prediction_value),
             }
+
+
+def get_spec(provider, storage, cache_dir, spec_path):
+    if provider == "local":
+        return read_msgpack(spec_path)
+
+    local_spec_path = os.path.join(cache_dir, "api_spec.msgpack")
```
vishalbollu, line 169: We should check if `api_spec.msgpack` doesn't exist before downloading, otherwise just read from local_spec_path. This should have been there before your PR. That check is even more useful now because the api spec is already downloaded in start_uvicorn

pkg/workloads/cortex/lib/type/model.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+class Model:
+    def __init__(self, name=None, model=None, signature_key=None, base_path=None):
```
vishalbollu, line 17: In what scenario would either name, model or base_path be None? Maybe they should be required positional args?

pkg/workloads/cortex/lib/type/predictor.py
```diff
@@ -20,36 +20,67 @@
...
-        self.model = kwargs.get("model")
         self.python_path = kwargs.get("python_path")
         self.config = kwargs.get("config", {})
         self.env = kwargs.get("env")
-        self.signature_key = kwargs.get("signature_key")
+
+        self.model_dir = model_dir
+        self.models = []
+        if kwargs.get("models"):
+            for model in kwargs.get("models"):
```
vishalbollu, line 39: nit: `kwargs.get("models")` -> `kwargs["models"]` because you already checked models exists

pkg/workloads/cortex/lib/type/predictor.py
```diff
@@ -20,36 +20,67 @@
...
+        signature_message = None
         if self.type == "onnx":
             from cortex.lib.client.onnx import ONNXClient
 
-            model_path = os.path.join(model_dir, os.path.basename(self.model))
-            client = ONNXClient(model_path)
-            cx_logger().info("ONNX model signature: {}".format(client.input_signature))
+            client = ONNXClient(self.models)
+            if self.models[0].name == consts.SINGLE_MODEL_NAME:
+                signature_message = "ONNX model signature: {}".format(self.models[0].signature_key)
```
vishalbollu, line 60: This objective of the signature messages is to print out the model signatures of the models you are using as opposed to the signature keys you specified in the YAML. This should be `client.input_signatures` instead of `self.models[0].signature_key`

pkg/workloads/cortex/lib/type/predictor.py
```diff
@@ -20,36 +20,67 @@
...
+            else:
+                signature_message = "ONNX model signatures: {}".format(
+                    get_model_signature_map(self.models)
```
vishalbollu, line 63: See above. This should be `client.input_signatures` instead of `get_model_signature_map(self.models)`

pkg/workloads/cortex/lib/type/predictor.py
```diff
@@ -20,36 +20,67 @@
...
+                validate_model_dir(model.base_path)
+
             tf_serving_address = tf_serving_host + ":" + tf_serving_port
-            validate_model_dir(model_dir)
-            client = TensorFlowClient(tf_serving_address, self.signature_key)
-            cx_logger().info("TensorFlow model signature: {}".format(client.input_signature))
+            client = TensorFlowClient(tf_serving_address, self.models)
+            if self.models[0].name == consts.SINGLE_MODEL_NAME:
+                signature_message = "TensorFlow model signature: {}".format(
+                    self.models[0].signature_key
```
vishalbollu, line 77: See above. This should be `client.input_signatures` instead of `self.models[0].signature_key`

pkg/workloads/cortex/lib/type/predictor.py
```diff
@@ -20,36 +20,67 @@
...
+                )
+            else:
+                signature_message = "TensorFlow model signatures: {}".format(
+                    get_model_signature_map(self.models)
```
vishalbollu, line 81: See above. This should be client.input_signatures instead of get_model_signature_map(self.models)

pkg/workloads/cortex/serve/serve.py
```diff
@@ -234,32 +233,18 @@ def build_predict_args(request: Request):
...
-        return msgpack.load(msgpack_file, raw=False)
+    return response
 
 
 def start():
     cache_dir = os.environ["CORTEX_CACHE_DIR"]
     provider = os.environ["CORTEX_PROVIDER"]
     spec_path = os.environ["CORTEX_API_SPEC"]
     project_dir = os.environ["CORTEX_PROJECT_DIR"]
-    model_dir = os.getenv("CORTEX_MODEL_DIR", None)
```
vishalbollu, line 262: Good catch

pkg/workloads/cortex/lib/server/tensorflow.py
```diff
@@ -0,0 +1,81 @@
...
+from cortex.lib.log import cx_logger
+
+
+class TensorFlowServing:
+    def __init__(self, address):
+        self.address = address
+        self.model_platform = "tensorflow"
+        self.channel = grpc.insecure_channel(self.address)
+        self.stub = model_service_pb2_grpc.ModelServiceStub(self.channel)
+        self.timeout = 900
```
vishalbollu, line 32: Is this timeout in seconds or milliseconds? If it is seconds, it looks like the timeout is 15 minutes to get a response for HandleReloadConfigRequest. I am not sure how long this request should take. Why not 10 seconds? We could catch the timeout and retry.

pkg/workloads/cortex/lib/server/tensorflow.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def add_models_config(self, names, base_paths, replace_models=False):
+        request = model_management_pb2.ReloadConfigRequest()
+        model_server_config = model_server_config_pb2.ModelServerConfig()
+
+        # create model(s) configuration
+        config_list = model_server_config_pb2.ModelConfigList()
+        for i in range(len(names)):
```
vishalbollu, line 40: nit: `for i in range(len(names)):` -> `for i, name in enumerate(names):`

pkg/workloads/cortex/lib/client/onnx.py
```diff
@@ -79,13 +103,16 @@ def input_signature(self):
...
-def transform_to_numpy(input_pyobj, input_metadata):
+def transform_to_numpy(input_pyobj, input_metadata, model_name):
     target_dtype = ONNX_TO_NP_TYPE[input_metadata.type]
     target_shape = input_metadata.shape
 
     try:
         for idx, dim in enumerate(target_shape):
-            if type(dim) is not int:
+            # dynamic_axes implies that a given dimension can have any size
+            if type(dim) is str and "dynamic_axes" in dim:
```
vishalbollu, line 113: Thanks for fixing this! Did you run into an issue where this wasn't supported?

pkg/workloads/cortex/lib/type/predictor.py
RobertLucian: It's used in one place: `self._compute_model_basepath`. I didn't remove it because `self._compute_model_basepath` method also uses another attribute of the class, so it would be too much to pass 2 extra arguments to the function when they can be kept as attributes instead. The reason is that you wouldn't remove just `self.model_dir` but also the other attribute `self.type` from `self._compute_model_basepath`.

What do you think?

cli/cmd/get.go
```diff
@@ -504,23 +504,36 @@ func describeModelInput(status *status.Status, apiEndpoint string) string {
...
+		if modelName == consts.CortexSingleModelName {
+			usesDefaultModel = true
 		}
-		rowNum++
 	}
 
 	t := table.Table{
 		Headers: []table.Header{
+			{Title: "model name", MaxWidth: 32, Hidden: usesDefaultModel},
 			{Title: "model input", MaxWidth: 32},
```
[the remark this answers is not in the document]
RobertLucian: Agreed.

cli/local/delete.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	"github.com/cortexlabs/cortex/pkg/lib/docker"
 	"github.com/cortexlabs/cortex/pkg/operator/schema"
+	"github.com/cortexlabs/cortex/pkg/types/spec"
 )
 
 func Delete(apiName string, keepCache bool) (schema.DeleteResponse, error) {
```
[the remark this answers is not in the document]
RobertLucian: Looking at this, I see that `keepCache` contains `keep-cache`'s value (which is an option passed at the command line when `cortex delete <api> --keep-cache`. Assuming the user wants to keep the cache, then `apiSpec` pointer in this function will be `nil`, which in the context of `DeleteAPI(apiName, apiSpec, nil)` will end up not deleting the models and therefore keep the cache instead. Let me know what you think!

cli/local/api.go
```diff
@@ -161,22 +154,31 @@ func DeleteAPI(apiName string, keepCache bool) error {
...
-		modelsToDelete.Subtract(modelsInUse)
-		for modelID := range modelsToDelete {
-			err := files.DeleteDir(filepath.Join(_modelCacheDir, modelID))
-			if err != nil {
-				errList = append(errList, err)
+		prevModelIDs := strset.FromSlice(prevAPISpec.ModelIDs())
+		newModelIDs := strset.FromSlice(newAPISpec.ModelIDs())
+
+		if !prevModelIDs.IsEqual(newModelIDs) {
+			toDeleteModels := strset.Difference(prevModelIDs, newModelIDs)
```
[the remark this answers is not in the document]
RobertLucian: Yep, I agree. At David's suggestion, we went with `strset.Difference(prevModelIDs, newModelIDs, modelsInUse)`. `Difference` can apparently accept any number of terms for subtraction.

cli/local/model_cache.go
```diff
@@ -31,8 +31,31 @@ import (
...
 )
 
+func CacheModels(apiSpec *spec.API, awsClient *aws.Client) ([]*spec.LocalModelCache, error) {
+	modelPaths := make([]string, len(apiSpec.Predictor.Models))
+	for i, modelResource := range apiSpec.Predictor.Models {
+		modelPaths[i] = modelResource.Model
+	}
+
+	localModelCaches := make([]*spec.LocalModelCache, len(modelPaths))
+	var err error
```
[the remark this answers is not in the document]
RobertLucian: Doing `localModelCaches[i], err := CacheModel(modelPath, awsClient)` gets me a syntax error because `localModelCaches` is already defined. I moved the error declaration on line 46 though `var err error`.

examples/pytorch/multi-model-text-analyzer/cortex.yaml
```diff
@@ -0,0 +1,12 @@
...
+- name: multi-model-text-analyzer
+  predictor:
+    type: python
+    path: predictor.py
+  compute:
+    cpu: 1
+    gpu: 1
+    mem: 6G
+  autoscaling:
+    threads_per_worker: 1
```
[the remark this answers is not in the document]
RobertLucian: Done!

examples/tensorflow/multi-model-classifier/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```bash
+export ENDPOINT=your-api-endpoint
+```
+
+When making a prediction with [sample.json](sample.json), the following image will be used:
+
+![sports car](https://i.imgur.com/zovGIKD.png)
+
+### Iris Classifier Model
```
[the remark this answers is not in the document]
RobertLucian: Went as discussed.

pkg/lib/sets/strset/strset.go
```diff
@@ -37,6 +37,10 @@ func New(ts ...string) Set {
 	return s
 }
 
+func FromSlice(items []string) Set {
```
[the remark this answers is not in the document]
RobertLucian: This one is called a few times in `cli/local/api.go`. It is used because in the context of those utilizations, going with `New(items...)` would become confusing to the reader.

pkg/types/spec/errors.go
```diff
@@ -32,6 +32,8 @@ const (
 	ErrDuplicateName                        = "spec.duplicate_name"
 	ErrDuplicateEndpointInOneDeploy         = "spec.duplicate_endpoint_in_one_deploy"
 	ErrDuplicateEndpoint                    = "spec.duplicate_endpoint"
+	ErrIllegalModelName                     = "spec.illegal_model_name"
```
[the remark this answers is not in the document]
RobertLucian: Yep, it's modified.

pkg/types/spec/errors.go
```diff
@@ -212,13 +230,27 @@ func ErrorInvalidTensorFlowModelPath() error {
 	})
 }
 
+func ErrorMissingTensorFlowModel(singleModelField string, multiModelField string, predictorType userconfig.PredictorType) error {
+	return errors.WithStack(&errors.Error{
+		Kind:    ErrMissingTensorFlowModel,
```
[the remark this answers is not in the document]
RobertLucian: Makes sense!

pkg/types/spec/errors.go
```diff
@@ -212,13 +230,27 @@ func ErrorInvalidTensorFlowModelPath() error {
...
+}
+
 func ErrorInvalidONNXModelPath() error {
 	return errors.WithStack(&errors.Error{
 		Kind:    ErrInvalidONNXModelPath,
 		Message: "onnx model path must be an onnx exported file ending in `.onnx`",
 	})
 }
 
+func ErrorDuplicateModels(duplicateModel string) error {
```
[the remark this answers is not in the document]
RobertLucian: Done!

pkg/types/userconfig/api.go
```diff
@@ -55,6 +56,12 @@ type Predictor struct {
 	SignatureKey           *string                `json:"signature_key" yaml:"signature_key"`
 }
 
+type ModelResource struct {
+	Name         string  `json:"name" yaml:"name"`
+	Model        string  `json:"model" yaml:"model"`
```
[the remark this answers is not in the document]
RobertLucian: Done!

pkg/types/userconfig/config_key.go
```diff
@@ -31,13 +31,17 @@ const (
...
+	ModelsKey                 = "models"
 	PythonPathKey             = "python_path"
 	ImageKey                  = "image"
 	TensorFlowServingImageKey = "tensorflow_serving_image"
 	ConfigKey                 = "config"
 	EnvKey                    = "env"
 	SignatureKeyKey           = "signature_key"
 
+	// ModelResource
+	ModelsNameKey = "name"
```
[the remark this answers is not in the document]
RobertLucian: `NameKey` is already used for the API's name. If we ever want to change the model's key name, it could get a bit more complex since it would be shared. What do you think?

pkg/workloads/cortex/lib/type/model.py
RobertLucian: None that I can think of. I'm making them positional arguments.

pkg/workloads/cortex/lib/type/predictor.py
RobertLucian: Yep. Likewise for everything else in `for model in kwargs["models"]` loop.

pkg/workloads/cortex/lib/client/onnx.py
RobertLucian: Yep, with GPT-2.

pkg/workloads/cortex/lib/server/tensorflow.py
RobertLucian: I've seen this thing need even 60-90 seconds. And this was just for a model (unfortunately, I can't remember which one it was). Basically, the request won't return until all models are loaded, which if they are a handful (> 10) it might take some time. I set `replace_models` to `False` (where it's called) so that it uses the `MergeFrom` function instead. Using the `MergeFrom` function instead of `CopyFrom` won't discard the already loaded models if there's a new request after a failed one.

pkg/workloads/cortex/lib/type/api.py
RobertLucian: Got it. Makes sense.

pkg/workloads/cortex/lib/type/predictor.py
deliahu: I didn't follow exactly what you are saying, let me know when is a good time to discuss it!

cli/local/api.go
```diff
@@ -119,18 +107,27 @@ func writeAPISpec(apiSpec *spec.API) error {
 	return nil
 }
 
-func DeleteAPI(apiName string, keepCache bool) error {
+func areAPIsEqual(a1, a2 *spec.API) bool {
+	if a1 != nil && a2 != nil {
+		return strset.FromSlice(a1.ModelIDs()).IsEqual(strset.FromSlice(a2.ModelIDs())) && a1.ID == a2.ID && a1.Compute.Equals(a2.Compute)
+	} else if a1 == nil && a2 == nil {
```
vishalbollu, line 113:
```
if a1 != nil && a2 != nil {
    ...
}

return a1 == nil && a2 == nil
```

RobertLucian:
@vishalbollu reducing the timeout for the process of loading big TF models works (that take more than the said timeout), but it can bring an avalanche of logs to `cortex logs <api-name>` coming from TFS. 

Basically, for each new gRPC call, the list of desired models gets logged to stdout as a string representation of `ModelServerConfig` [object](https://www.tensorflow.org/tfx/serving/serving_config). I think the easiest thing is to just set the timeout to something big anyway (i.e. 600 seconds). While the gRPC call is waiting for a response,  a message like _"model(s) are still loading ..."_ gets repeatedly printed every 5 seconds to stdout.

cli/local/model_cache.go
```diff
@@ -134,6 +135,42 @@ func CacheModel(modelPath string, awsClient *aws.Client) (*spec.LocalModelCache,
 	return &localModelCache, nil
 }
 
+func DeleteCachedModels(apiName string, modelsToDelete []string) error {
+	errList := []error{}
+	modelsInUse := strset.New()
+	apiSpecList, err := ListAPISpecs()
+	errList = append(errList, err)
+
+	for _, apiSpec := range apiSpecList {
```
vishalbollu, line 144: This for loop should be wrapped in a:

```
if err !=nil {
   ...
}
```

The way it is written right now works well. I think for readability and not having issues, we should make certain that this code path of iterating through API specs is only run when there isn't an error.

RobertLucian: Done!

pkg/types/spec/api.go
```diff
@@ -72,6 +74,40 @@ func GetAPISpec(apiConfig *userconfig.API, projectID string, deploymentID string
...
+		}
+	}
+
+	return names
+}
+
+func (api *API) SubtractModelIDs(apis ...*API) []string {
+	modelIDs := strset.FromSlice(api.ModelIDs())
+	for _, a := range apis {
+		modelIDs.Subtract(strset.FromSlice(a.ModelIDs()))
```
deliahu, line 102: nit: I think this would work and would be a little simpler:

```go
modelIDs.Remove(a.ModelIDs()...)
```

RobertLucian: Yep, that's a nice alternative. Gonna change this now.

RobertLucian: This is now done!

PR #1124, opened 2020-06-15
support grpc server

### LINKED ISSUE #1056: Add gRPC support to RealtimeAPI kind
#### Description

https://grpc.io/docs/guides/concepts/
https://blog.feathersjs.com/http-vs-websockets-a-performance-comparison-da2533f13a77

1. For best performance, the protocol would be customizable in the API spec (REST vs gRPC). A field called `protocol` would be added to the `predictor` section.
1. User is responsible for providing a protobuf file in the API spec as well. Cortex would automatically generate the server files from the protobuf.
1. Both REST and gRPC cannot be served for the same API.
1. Istio doesn't have to be changed - it's got native support for HTTP 2.0. We probably need to change the service mode of APIs to headless mode.
1. The protobuf would only support a single method/service - `predict`. Its input/output is defined by the user. We don't provide a default protobuf.

Example of python predictor for gRPC:

```python
# UserProvidedServicer provides an implementation of the methods of the RouteGuide service.
class PythonPredictor(user_pb2_grpc.UserServicer):
  def start(self, config):
    self.config = config
  def predict(self, request_iterator, context):
    prev_notes = []
    for new_note in request_iterator:
      for prev_note in prev_notes:
        if prev_note.location == new_note.location:
          yield prev_note
      prev_notes.append(new_note)
```
We don't necessarily need (or can) to ask the user to subclass the generated server files - we can wrap this class instead. This way, the class could look like

```python
class PythonPredictor:
  def __init__(self, config):
    self.config = config
  def predict(self, request_iterator, context):
    prev_notes = []
    for new_note in request_iterator:
      for prev_note in prev_notes:
        if prev_note.location == new_note.location:
          yield prev_note
      prev_notes.append(new_note)
```

Creating a gRPC server is straightforward. We'll need multiple servers as dictated by `processes_per_replica` field.

```python
def serve():
  server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
  user_pb2_grpc.add_UserServices_to_server(
      PythonPredictor(), server)
  server.add_insecure_port('[::]:50051')
  server.start()
  server.wait_for_termination()
```

Example API spec:

```yaml
- name: iris-classifier
  kind: RealtimeAPI
  predictor:
    type: python
    path: predictor.py
    protocol: grpc/rest
    protobuf: predictor.proto
  compute:
    cpu: 0.2
    mem: 200M
```

To parse proto files in Go: https://github.com/tallstoat/pbparser

#### Motivation

Because it has lower latency and higher throughput than REST. It's already well known within the community:
* https://medium.com/@EmperorRXF/evaluating-performance-of-rest-vs-grpc-1b8bdf0b22da

#### Additional context

Maybe use something like Linkerd *(maybe this isn't needed)* for load-balancing gRPC requests:
* https://kubernetes.io/blog/2018/11/07/grpc-load-balancing-on-kubernetes-without-tears/

zouyee:
closes #1056

---

checklist:

- [x] update examples


deliahu:
@zouyee Thank you for taking a stab at this!

Unfortunately the title that we wrote for https://github.com/cortexlabs/cortex/issues/1056 was not clear; it was our intention for this to refer to the user's APIs that are generated with `cortex deploy`, not the Cortex operator's API (I have now updated the title from "Add support for gRPC requests" to "Add support for gRPC requests to user's APIs"). It is not a high priority for us to convert the Cortex operator's API to gRPC, since it seems to be working well enough as implemented today.

Feel free to unassign this issue, or if you are still interested in adding gRPC support to the user's APIs, let me know and we can discuss how we might want to approach this (it's tricky since right now, we use FastAPI for the user's APIs, and FastAPI doesn't seem to support gRPC).

zouyee:
> Feel free to unassign this issue, or if you are still interested in adding gRPC support to the user's APIs, let me know and we can discuss how we might want to approach this (it's tricky since right now, we use FastAPI for the user's APIs, and FastAPI doesn't seem to support gRPC).

Is it currently necessary to convert fastapi to golang implementation ? I found that no matter tf or onnx support grpc.

zouyee:
> @zouyee Thank you for taking a stab at this!
> 
> Unfortunately the title that we wrote for #1056 was not clear; it was our intention for this to refer to the user's APIs that are generated with `cortex deploy`, not the Cortex operator's API (I have now updated the title from "Add support for gRPC requests" to "Add support for gRPC requests to user's APIs"). It is not a high priority for us to convert the Cortex operator's API to gRPC, since it seems to be working well enough as implemented today.
> 

If we need user API to support grpc (as mentioned above, convert the implementation language？),  it can also be implemented, but operator API is also required for grpc support.

deliahu:
> Is it currently necessary to convert fastapi to golang implementation? I found that no matter tf or onnx support grpc.

That's correct, I think converting fastai to golang would create a new issue, which is how to run the user's Python code.

One option could be to add a sidecar gRPC service which forwards requests to the FastAPI container, do you think that would work? Do you know how much overhead this would introduce?

deliahu:
@zouyee I'll close this PR for now, feel free to reach out in the future if you plan on working on gRPC for users' APIs (see my comment above).

PR #1146, opened 2020-06-20
Call __init__() from request thread pool

deliahu:
@RobertLucian @vishalbollu what do you think about this? Are there any issues you can think of with this approach?


vishalbollu:
This looks like an interesting idea. I have a few points:

If each thread runs its own predictor constructor, does this mean that each thread loads a model into memory separately? At that point, it may be safer to use processes instead of threads. If that is the case, we can also only initialize the threadpool if the thread is of size 1. At least this way, running TensorFlow sessions should work by default when threads_per_worker is set to 1.

Another consideration, which isn't mutually exclusive to the above, is to import the threadpool from serve.py in the predictor implementation and run initializations in the predictor's constructor. We can have a guide showcasing this pattern if it works (and I am not sure that it does work).

PR #1153, opened 2020-06-23
Add API Gateway endpoint to info command output

deliahu:


manager/info.sh
```diff
@@ -34,8 +36,10 @@ fi
 eksctl utils write-kubeconfig --cluster=$CORTEX_CLUSTER_NAME --region=$CORTEX_REGION | grep -v "saved kubeconfig as" | grep -v "using region" | grep -v "eksctl version" || true
 
 operator_endpoint=$(get_operator_endpoint)
-apis_endpoint=$(get_apis_endpoint)
+api_load_balancer_endpoint=$(get_api_load_balancer_endpoint)
+api_gateway_endpoint=$(get_api_gateway_endpoint)
 
 # before modifying this, search for this prefix
```
tthebst, line 42: maybe at title **endpoints**.
`echo -e "\033[1mEndpoints:\033[0m"`

manager/info.sh
```diff
@@ -34,8 +36,10 @@ fi
...
-echo "operator endpoint: $operator_endpoint"
-echo "apis endpoint:     $apis_endpoint"
+echo "operator endpoint:          $operator_endpoint"
+echo "api load balancer endpoint: $api_load_balancer_endpoint"
+echo "api gateway endpoint:       $api_gateway_endpoint"
```
tthebst, line 45: add newline `echo ""`
to have at least one new line to next output `fetching cluster status ...`

tthebst:
So previously looked like this.
clusterinfo-previous
Now:
Screenshot 2020-06-23 at 17 38 29


What do you think?

deliahu:
I think it looks great!

What do you think about removing "endpoints" from the items?

```text
**Endpoints**
operator:          ***
api load balancer: ***
api gateway:       ***
```

Also, I see a new line before `fetching cluster status ...` on my end, I wonder why it doesn't show up on yours (I think it comes from [here](https://github.com/cortexlabs/cortex/blob/info-api-gateway/cli/cmd/cluster.go#L428).

tthebst:
Yes. If we have the title **endpoint** it makes sense to remove endpoint. 

You're right with the newline. I removed the `fmt.Println()` because I thought it was me who put it there for testing lol. So no need for extra new line

Issue #1147, opened 2020-06-20
Add possibility to export environment variables with .env file


RobertLucian:
#### Description

Add support for exporting environment variables from an `.env` file placed in the root directory of a Cortex project.

#### Motivation

In case the user doesn't want to export environment variables using the `predictor:env` field in `cortex.yaml`. A reason for that could be to keep the `cortex.yaml` deployment clean.

lminer:
Does this expose the environment variables to the tensorflow-serving image as well? I'm trying to read directly from s3 from tensorflow-serving and it isn't seeing my credentials.

RobertLucian:
@lminer only the `predictor.image` container gets the exported variables from `.env`. So the answer is no, they don't get exported for the `predictor.tensorflow_serving_image` container.

`predictor.tensorflow_serving_image` at least in our view is only for TFS - anything else is done in the API container (`predictor.image`). Is it possible for you to start a service or anything that you want in `dependencies.sh` instead? `dependencies.sh` includes the exported environment variables and it runs before the API, inside the API container. Here's more on [`dependencies.sh`](https://docs.cortex.dev/workloads/managing-dependencies/system-packages).

Also, your API will inherit the cluster's AWS credentials (provided it's an AWS cluster) and so you don't have to export them again - unless these credentials are for S3 buckets that fall outside the API credentials' scope. Can you tell us more about this use case?

I see that you want to read from an S3 bucket. If it's a model you want to download, Cortex already handles that awesomely. More on that [here](https://docs.cortex.dev/workloads/realtime/models). Is this what you want? If not, can you tell us about your case?

lminer:
@RobertLucian because of the GRPC limit on the amount of data that I can send to tensorflow-serving, I thought that instead I would simply send a  link to an object in an s3 bucket and then read and write to s3 directly within the model itself. However, in order to do this, I need some way to get the AWS credentials to the model and it appears as if this can only be done via environment variables.

RobertLucian:
@lminer I see. That would not be an optimal architecture.

We have increased the priority on https://github.com/cortexlabs/cortex/issues/1740 and it's the next thing for me to work on. Since this seems to be urgent to you, we will probably make a patch release for you or just create a customised image that will work for you. Will this be acceptable to you?

lminer:
That would be great. Thank you!

deliahu:
[...]

PR #1154, opened 2020-06-23
Add .env to export environment variables

RobertLucian:
Closes https://github.com/cortexlabs/cortex/issues/1147.

---

checklist:

### DIFF
--- a/pkg/lib/files/files.go
+++ b/pkg/lib/files/files.go
@@ -456,7 +456,7 @@ func IsFilePathPython(path string) bool {
 type IgnoreFn func(string, os.FileInfo) (bool, error)
 
 func IgnoreHiddenFiles(path string, fi os.FileInfo) (bool, error) {
-	if !fi.IsDir() && strings.HasPrefix(fi.Name(), ".") {
+	if !fi.IsDir() && strings.HasPrefix(fi.Name(), ".") && fi.Name() != ".env" {
 		return true, nil
 	}
 	return false, nil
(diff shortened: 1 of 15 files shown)


pkg/lib/files/files.go
```diff
@@ -456,7 +456,7 @@ func IsFilePathPython(path string) bool {
 type IgnoreFn func(string, os.FileInfo) (bool, error)
 
 func IgnoreHiddenFiles(path string, fi os.FileInfo) (bool, error) {
```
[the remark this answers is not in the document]
RobertLucian: What do you think of renaming it to `IgnoreHiddenFilesExceptDotEnv` instead? Otherwise, I think, it would be trickier to remove the ignored file (`.env`) from `IgnoreHiddenFiles` if it were to be separate.

I've pushed a commit to address this.

pkg/types/spec/validations.go
```diff
@@ -870,7 +870,7 @@ func GetTFServingExportFromLocalPath(path string) (string, error) {
 	if err := files.CheckDir(path); err != nil {
 		return "", err
 	}
-	paths, err := files.ListDirRecursive(path, false, files.IgnoreHiddenFiles, files.IgnoreHiddenFolders)
+	paths, err := files.ListDirRecursive(path, false, files.IgnoreHiddenFilesExceptDotEnv, files.IgnoreHiddenFolders)
```
deliahu, line 873: Is it necessary to include `.env` files here? It probably doesn't matter either way, so it's ok to leave it as is I think (the alternative would be to bring back `files.IgnoreHiddenFiles()`)

RobertLucian: Modified it to not rely on `IgnoreHiddenFiles` but by adding the `.env` path to the list of paths if the file exists.

PR #1155, opened 2020-06-23
Prevent endpoint updates from triggering rolling replica updates

### LINKED ISSUE #1116: Prevent endpoint updates from triggering rolling replica updates
#### Description

* Check that endpoint is not used in the api server
* Fix should be to remove it from api ID calculation, would need to update `operator/api.go` to correctly identify api updates

deliahu:
closes #1116

### DIFF
--- a/pkg/workloads/cortex/lib/type/api.py
+++ b/pkg/workloads/cortex/lib/type/api.py
@@ -35,7 +35,6 @@ class API:
         self.key = kwargs["key"]
         self.metadata_root = kwargs["metadata_root"]
         self.name = kwargs["name"]
-        self.endpoint = kwargs["endpoint"]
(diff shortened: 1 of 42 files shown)


pkg/workloads/cortex/lib/type/api.py
```diff
@@ -35,7 +35,6 @@ def __init__(self, provider, storage, model_dir, cache_dir=".", **kwargs):
         self.key = kwargs["key"]
         self.metadata_root = kwargs["metadata_root"]
         self.name = kwargs["name"]
-        self.endpoint = kwargs["endpoint"]
```
vishalbollu, line 38: This change should be a part of the networking move PR because this code path should break in this PR https://github.com/cortexlabs/cortex/pull/1151

PR #1157, opened 2020-06-23
Remove local port from API ID calculation

deliahu:
I think we should revisit API IDs soon, but for now, it seemed to me that local port should be removed from it. For now, I am thinking that the "API ID" should be used mostly as the "predictor ID". We should make guidelines around this, and improve the naming, but for now I think this makes sense; what do you think?

### DIFF
--- a/cli/local/api.go
+++ b/cli/local/api.go
@@ -26,6 +26,7 @@ import (
 	"github.com/cortexlabs/cortex/pkg/lib/errors"
 	"github.com/cortexlabs/cortex/pkg/lib/files"
 	"github.com/cortexlabs/cortex/pkg/lib/msgpack"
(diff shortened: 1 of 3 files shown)


deliahu:
@RobertLucian Yes I think that is a good point. It probably makes sense to have a few different IDs; right now, API ID is probably what we would end up calling "predictor ID": it is an ID that represents everything about your API container, and nothing outside of it (which is why things like `networking` and `autoscaling` are not included in that ID). This ID is useful for example if you want stats about your API; simply changing the endpoint or the autoscaling configuration should not e.g. create a new cloudwatch dashboard. In addition, as you said, having an ID that incorporates everything would be useful for seeing if anything has changed.

> This implementation could also be used until we revisit this at a later time.

Yes, I think we can go forward with this PR and discuss this matter in the future.

Issue #1090, opened 2020-05-29
Move replica parallelism API configurations to separate key


deliahu:
#### Description

Move `workers_per_replica` and `threads_per_worker` configurations out of `autoscaling` and into a different key, since these configurations apply to batch prediction and local APIs as well.

Unblocks https://github.com/cortexlabs/cortex/issues/960

Issue #960, opened 2020-04-13
Enable workers and threads_per_worker for local deployments


deliahu:
#### Description

The `workers_per_replica` and `threads_per_worker` configuration fields can be applied locally as well, although currently they are blocked (since the entire `autoscaling` config is blocked for local deployments).

#### Questions

* Should the fields be moved out of `autoscaling`? If not, should the local config use `workers` instead of `workers_per_replica`?

PR #1158, opened 2020-06-23
Move & rename replica parallelism fields

RobertLucian:
Closes https://github.com/cortexlabs/cortex/issues/960, closes https://github.com/cortexlabs/cortex/issues/1090.

---

checklist:

- [x] update examples


cli/local/docker_spec.go
```diff
@@ -88,9 +88,9 @@ func getAPIEnv(api *spec.API, awsClient *aws.Client) []string {
 		"CORTEX_MODELS="+strings.Join(api.ModelNames(), ","),
 		"CORTEX_API_SPEC="+filepath.Join("/mnt/workspace", filepath.Base(api.Key)),
 		"CORTEX_PROJECT_DIR="+_projectDir,
-		"CORTEX_WORKERS_PER_REPLICA=1",
-		"CORTEX_THREADS_PER_WORKER=1",
-		"CORTEX_MAX_WORKER_CONCURRENCY=1000",
+		"CORTEX_PROCESSES_PER_REPLICA="+s.Int32(api.Predictor.ProcessesPerReplica),
+		"CORTEX_THREADS_PER_PROCESS="+s.Int32(api.Predictor.ThreadsPerProcess),
+		"CORTEX_MAX_PROCESS_CONCURRENCY=1000",
```
deliahu, line 93: maybe this should match what we do [here](https://github.com/cortexlabs/cortex/blob/refactor/parallelism-fields/pkg/operator/operator/k8s_specs.go#L634), but replace `api.Autoscaling.MaxReplicaConcurrency` with `1024` (since that is the default value and you can't update it for local). Or maybe `api.Autoscaling.MaxReplicaConcurrency` should be moved out of autoscaling too (that could be a separate PR, and less urgent). What do you think?

RobertLucian: Yes, you are totally right.

RobertLucian:
Tested this with various examples. In these examples, it has been tested with different values for `processes_per_replica` and `threads_per_process`. This also included verifications that the CloudWatch dashboard, the endpoints and the multi-model feature work.

cli/local/docker_spec.go
RobertLucian: I'm also thinking that if `MaxReplicaConcurrency` is taken out, then so `TargetReplicaConcurrency` should be. Probably move all these 4 fields in a separate section too called `parallelism`, or `concurrency`.

deliahu: Yes that could make sense. We can discuss this more in the future, for now I think we can go ahead with this PR.

PR #1163, opened 2020-06-25
Update local model downloading new lines

deliahu:
@RobertLucian Does this change look ok to you? Do you know if we'd misformat something else by making this change?

Before:

```text
$ cx deploy 

￮ downloading model s3://cortex-examples/tensorflow/image-classifier/inception/1569014553 . ✓

￮ downloading model s3://cortex-examples/tensorflow/iris-classifier/nn/1569001258 . ✓

￮ downloading model s3://cortex-examples/tensorflow/resnet50/1 . ✓

creating multi-model-classifier

cortex get                          (show api statuses)
cortex get multi-model-classifier   (show api info)
cortex logs multi-model-classifier  (stream api logs)
```

After:

```text
$ cx deploy 

￮ downloading model s3://cortex-examples/tensorflow/image-classifier/inception/1569014553 . ✓
￮ downloading model s3://cortex-examples/tensorflow/iris-classifier/nn/1569001258 . ✓
￮ downloading model s3://cortex-examples/tensorflow/resnet50/1 . ✓

creating multi-model-classifier

cortex get                          (show api statuses)
cortex get multi-model-classifier   (show api info)
cortex logs multi-model-classifier  (stream api logs)
```

### DIFF
--- a/cli/local/model_cache.go
+++ b/cli/local/model_cache.go
@@ -54,7 +54,9 @@ func CacheModels(apiSpec *spec.API, awsClient *aws.Client) ([]*spec.LocalModelCa
 		localModelCaches[i].TargetPath = apiSpec.Predictor.Models[i].Name
 	}
 
-	fmt.Println("") // Newline to group all of the model information
+	if len(localModelCaches) > 0 {
+		fmt.Println("") // Newline to group all of the model information
+	}
 
 	return localModelCaches, nil
 }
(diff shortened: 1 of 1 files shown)


RobertLucian:
If `CacheModels` function is ever to be called with no models, then the newline would mess up the output a bit. It's just better to check if any models have been printed to stdout and then print the newline as well. 

Currently, `CacheModels` is only called when there are models to be cached, but still, it's better to be safe than sorry.

PR #1164, opened 2020-06-25
Set tags on log groups

### LINKED ISSUE #1078: Tags are not propagated to API log groups
### Description

Tags specified in cluster configuration are not propagated to API log groups because API log groups are created by FluentD.

#### Suggested solution

- explore ways to add tags to log groups created by FluentD
- create log groups upon API creation in operator

tthebst:
closes #1078

---

How do you format yaml's? My VSCode formats them like this.

checklist:

- [ ] update examples

### DIFF
--- a/manager/install.sh
+++ b/manager/install.sh
@@ -238,6 +238,7 @@ function main() {
   echo "✓"
 
   echo -n "￮ configuring logging "
+  export CORTEX_TAGS_JSON=$(echo $CORTEX_TAGS  | jq -sR '[split(",")[] | rtrimstr("\n") | split("=") | {(.[0]): .[1]} ] | add' | tr -d '\n' )
   envsubst < manifests/fluentd.yaml | kubectl apply -f - >/dev/null
   echo "✓"
 
--- a/manager/install.sh
+++ b/manager/install.sh
@@ -238,7 +238,6 @@ function main() {
   echo "✓"
 
   echo -n "￮ configuring logging "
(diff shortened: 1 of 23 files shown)


deliahu:
We haven't been using a YAML formatter, but maybe we should! Which one do you use?

Regardless, I think we should make a separate PR after this one if we do decide to reformat, to keep the diff in the PR more relevant

manager/install.sh
```diff
@@ -238,6 +238,7 @@ function main() {
   echo "✓"
 
   echo -n "￮ configuring logging "
+  export CORTEX_TAGS_JSON=$(echo $CORTEX_TAGS  | jq -sR '[split(",")[] | rtrimstr("\n") | split("=") | {(.[0]): .[1]} ] | add' | tr -d '\n' )
```
deliahu, line 241: Nice job with the `jq` query! I have always struggled to write those :)

There is a simpler place we can do this, which is in `cluster_config_env.py` (that is the file which exports these env vars. Right below `print(f"export CORTEX_TAGS={inlined_tags}")`, we should be able to add a line like `print(f"export CORTEX_TAGS_JSON='{json.dumps(value)}'")`, what do you think about this approach?

manager/manifests/fluentd.yaml
```diff
@@ -168,9 +169,6 @@ metadata:
   name: fluentd
   namespace: default
 spec:
-  selector:
```
deliahu, line 171: Good catch! Do you mind moving this to a separate PR, along with the change in `operator.yaml`?

manager/install.sh
tthebst: everything is better than jq lol

PR #1167, opened 2020-06-26
Display image pull error

### LINKED ISSUE #955: Surface pod creation errors

deliahu:
closes #955

Note: ideally this error should not be reached, since the operator now validates access to Docker images before deploying an API.

---

checklist:

### DIFF
--- a/pkg/lib/k8s/pod.go
+++ b/pkg/lib/k8s/pod.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/cortexlabs/cortex/pkg/lib/errors"
+	"github.com/cortexlabs/cortex/pkg/lib/sets/strset"
 	kcore "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	kmeta "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


pkg/lib/k8s/pod.go
```diff
@@ -45,6 +46,7 @@ const (
 	PodStatusPending      PodStatus = "Pending"
 	PodStatusInitializing PodStatus = "Initializing"
 	PodStatusRunning      PodStatus = "Running"
+	PodStatusErrImagePull PodStatus = "Image pull error"
```
vishalbollu, line 49: nit: `PodStatusErrImagePull` -> `PodStatusImagePullError`

pkg/lib/k8s/pod.go
```diff
@@ -59,6 +61,9 @@ var _killStatuses = map[int32]bool{
 	129: true, // SIGHUP
 }
 
+// https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/images/types.go#L27
```
vishalbollu, line 64: Use the Kubernetes version used in Cortex here https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/images/types.go#L27?

PR #1172, opened 2020-06-27
Delete API Gateway if cluster up fails

deliahu:
---

checklist:

### DIFF
--- a/cli/cmd/cluster.go
+++ b/cli/cmd/cluster.go
@@ -168,13 +168,13 @@ var _upCmd = &cobra.Command{
 
 		out, exitCode, err := runManagerUpdateCommand("/root/install.sh", clusterConfig, awsCreds, _flagClusterEnv)
 		if err != nil {
-			_, _ = awsClient.DeleteAPIGatewayByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName) // best effort deletion
-			_, _ = awsClient.DeleteVPCLinkByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName)    // best effort deletion
+			awsClient.DeleteAPIGatewayByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName) // best effort deletion
+			awsClient.DeleteVPCLinkByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName)    // best effort deletion
 			exit.Error(err)
 		}
 		if exitCode == nil || *exitCode != 0 {
-			_, _ = awsClient.DeleteAPIGatewayByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName) // best effort deletion
-			_, _ = awsClient.DeleteVPCLinkByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName)    // best effort deletion
+			awsClient.DeleteAPIGatewayByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName) // best effort deletion
+			awsClient.DeleteVPCLinkByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName)    // best effort deletion
 			helpStr := "\nDebugging tips (may or may not apply to this error):"
 			helpStr += fmt.Sprintf("\n* if your cluster started spinning up but was unable to provision instances, additional error information may be found in the activity history of your cluster's autoscaling groups (select each autoscaling group and click the \"Activity History\" tab): https://console.aws.amazon.com/ec2/autoscaling/home?region=%s#AutoScalingGroups:", *clusterConfig.Region)
 			helpStr += fmt.Sprintf("\n* if your cluster started spinning up, please ensure that your CloudFormation stacks for this cluster have been fully deleted before trying to spin up this cluster again (you can delete your CloudFormation stacks from the AWS console: %s)", getCloudFormationURL(clusterConfig.ClusterName, *clusterConfig.Region))
--- a/cli/cmd/cluster.go
+++ b/cli/cmd/cluster.go
@@ -138,7 +138,7 @@ var _upCmd = &cobra.Command{
 			exit.Error(err)
 		}
 
-		err = CreateBucketIfNotFound(awsClient, clusterConfig.Bucket)
+		err = createBucketIfNotFound(awsClient, clusterConfig.Bucket)
 		if err != nil {
 			exit.Error(err)
 		}
@@ -147,7 +147,7 @@ var _upCmd = &cobra.Command{
 			exit.Error(err)
 		}
 
-		err = CreateLogGroupIfNotFound(awsClient, clusterConfig.LogGroup)
+		err = createLogGroupIfNotFound(awsClient, clusterConfig.LogGroup)
 		if err != nil {
 			exit.Error(err)
 		}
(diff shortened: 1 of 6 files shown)


cli/cmd/cluster.go
```diff
@@ -768,10 +779,38 @@ func createDashboard(awsClient *aws.Client, dashboardName string) error {
...
 		return err
 	}
 
 	fmt.Println(" ✓")
 
 	return nil
 }
+
+// createOrClearAPIGateway creates an API gateway for the cluster (or clears an existing one if it already exists)
+func createOrClearAPIGateway(awsClient *aws.Client, clusterName string) error {
```
tthebst, line 792: maybe change function name because it clears AND creates the gateway

cli/cmd/cluster.go
```diff
@@ -753,8 +764,8 @@ func CreateLogGroupIfNotFound(awsClient *aws.Client, logGroup string) error {
 	return nil
 }
 
-// createDashboard creates a new dashboard (or clears an existing one if it already exists)
-func createDashboard(awsClient *aws.Client, dashboardName string) error {
+// createOrClearDashboard creates a new dashboard (or clears an existing one if it already exists)
+func createOrClearDashboard(awsClient *aws.Client, dashboardName string) error {
```
tthebst, line 768: maybe change function name because it clears AND creates the dashboard

cli/cmd/cluster.go
```diff
@@ -768,10 +779,38 @@ func createDashboard(awsClient *aws.Client, dashboardName string) error {
...
+		return errors.Append(err, fmt.Sprintf("\n\nunable to delete existing vpc link with tag %s=%s; please delete it via the api gateway console: https://%s.console.aws.amazon.com/apigateway/main/vpc-links", clusterconfig.ClusterNameTag, clusterName, awsClient.Region))
+	}
+
+	_, err = awsClient.DeleteAPIGatewayByTag(clusterconfig.ClusterNameTag, clusterName)
+	if err != nil {
+		fmt.Print("\n\n")
+		return errors.Append(err, fmt.Sprintf("\n\nunable to delete existing api gateway with tag %s=%s; please delete it via the api gateway console: https://%s.console.aws.amazon.com/apigateway/main/apis", clusterconfig.ClusterNameTag, clusterName, awsClient.Region))
+	}
+
+	tags := map[string]string{clusterconfig.ClusterNameTag: clusterName}
```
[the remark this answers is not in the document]
deliahu: Yes absolutely, good catch!

cli/cmd/cluster.go
deliahu: I renamed it to `createOrReplaceAPIGateway`, does that sounds ok to you? My thinking is it creates it if it doesn't exist, and it replaces it with a new one if it did exist.

cli/cmd/cluster.go
deliahu: My thinking was that it creates it if it doesn't exist, or it clears it if it did exist. Do you think that makes sense, or you still suggest changing the name?

cli/cmd/cluster.go
tthebst: this is better

cli/cmd/cluster.go
tthebst: with dashboard clear makes sense

cli/cmd/cluster.go
```diff
@@ -156,16 +156,25 @@ var _upCmd = &cobra.Command{
...
+		}
+
+		err = createOrReplaceAPIGateway(awsClient, clusterConfig.ClusterName, clusterConfig.Tags)
 		if err != nil {
 			exit.Error(err)
 		}
 
 		out, exitCode, err := runManagerUpdateCommand("/root/install.sh", clusterConfig, awsCreds, _flagClusterEnv)
 		if err != nil {
+			_, _ = awsClient.DeleteAPIGatewayByTag(clusterconfig.ClusterNameTag, clusterConfig.ClusterName) // best effort deletion
```
vishalbollu, line 171: `_, _ = ` can be removed

manager/get_api_gateway_endpoint.py
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import boto3
+import os
+
+
+def get_api_gateway_endpoint():
+    cluster_name = os.environ["CORTEX_CLUSTER_NAME"]
```
vishalbollu, line 20: Since there are only 2 variables, it may be easier to follow the code that calls this python file if these values are accepted as args

manager/get_api_gateway_id.py
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import boto3
+import os
+
+
+def get_api_gateway_id():
+    cluster_name = os.environ["CORTEX_CLUSTER_NAME"]
```
vishalbollu, line 20: Since there are only 2 variables, it may be easier to follow the code that calls this python file if these values are accepted as args

pkg/lib/aws/apigateway.go
```diff
@@ -25,6 +25,34 @@ import (
...
+	}
+
+	_, err = c.APIGatewayV2().CreateStage(&apigatewayv2.CreateStageInput{
+		ApiId:      createAPIResponse.ApiId,
+		AutoDeploy: aws.Bool(true),
+		StageName:  aws.String("$default"),
+		Tags:       aws.StringMap(tags),
+	})
+	if err != nil {
+		_ = c.DeleteAPIGateway(*createAPIResponse.ApiId) // best effort cleanup
```
vishalbollu, line 49: `_ =` can be dropped here

manager/get_api_gateway_endpoint.py
```diff
@@ -0,0 +1,36 @@
...
+    client_apigateway = boto3.client("apigatewayv2", region_name=region)
+
+    paginator = client_apigateway.get_paginator("get_apis")
+    for api_gateway_page in paginator.paginate():
+        for api_gateway in api_gateway_page["Items"]:
+            if api_gateway["Tags"].get("cortex.dev/cluster-name") == cluster_name:
+                return api_gateway["ApiEndpoint"]
+
+    raise Exception(
+        f"Could not find api gateway with tag cortex.dev/cluster-name={cluster_name} in {region}"
```
vishalbollu, line 31: `Could not find api gateway with tag` -> `Could not get endpoint for api gateway with tag`

manager/get_api_gateway_endpoint.py
deliahu: The other python files, e.g. `create_gateway_integration.py`, also pull from the environment variables, so I think for now I lean towards leaving it as is. I'm actually leaning toward moving more of the api gateway code from `install.sh` to python, so at that point we can convert to only using args, does that sound ok?

manager/get_api_gateway_id.py
deliahu: (see comment above)

manager/get_api_gateway_endpoint.py
deliahu: I think the current error message is a little more descriptive/accurate, since if this exception is raised, it means that there was not API gateway that had the tag. I'll reword it to make it more clear: "your cluster's api gateway (with tag cortex.dev/cluster-name={cluster_name}) does not exist". Does that sound ok?

PR #1184, opened 2020-06-29
Remove unnecessary env var & reposition env var source

RobertLucian:
`CORTEX_MAX_REPLICA_CONCURRENCY` environment variable is no longer required and placing the sourcing of `.env` before any export in `run.sh` might give us more leeway when it comes to various workarounds for our users.

---

checklist:

### DIFF
--- a/pkg/operator/resources/syncapi/k8s_specs.go
+++ b/pkg/operator/resources/syncapi/k8s_specs.go
@@ -638,10 +638,6 @@ func getEnvVars(api *spec.API, container string) []kcore.EnvVar {
 				Value: s.Int32(api.Predictor.ThreadsPerProcess),
 			},
 			kcore.EnvVar{
-				Name:  "CORTEX_MAX_REPLICA_CONCURRENCY",
-				Value: s.Int64(api.Autoscaling.MaxReplicaConcurrency),
-			},
-			kcore.EnvVar{
 				Name: "CORTEX_MAX_PROCESS_CONCURRENCY",
 				// add 1 because it was required to achieve the target concurrency for 1 process, 1 thread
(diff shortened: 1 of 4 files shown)


pkg/operator/resources/syncapi/k8s_specs.go
```diff
@@ -637,10 +637,6 @@ func getEnvVars(api *spec.API, container string) []kcore.EnvVar {
 				Name:  "CORTEX_THREADS_PER_PROCESS",
 				Value: s.Int32(api.Predictor.ThreadsPerProcess),
 			},
-			kcore.EnvVar{
-				Name:  "CORTEX_MAX_REPLICA_CONCURRENCY",
```
vishalbollu, line 641: Thanks for catching this merge error. On second thought `CORTEX_MAX_REPLICA_CONCURRENCY` is the more accurate term name. Perhaps we should rename `CORTEX_MAX_PROCESS_CONCURRENCY` to `CORTEX_MAX_REPLICA_CONCURRENCY`

RobertLucian: By the look of it, `CORTEX_MAX_PROCESS_CONCURRENCY` is computed by taking the `max_replica_concurrency` (the old term being `CORTEX_MAX_REPLICA_CONCURRENCY`) and dividing it by the number of processes `processes_per_replica`. So, it doesn't look like it's the same thing. The current variable that holds `CORTEX_MAX_REPLICA_CONCURRENCY` seems to be `CORTEX_MAX_CONN`.

RobertLucian:
@vishalbollu as per our conversation, it has been confirmed that the `limit_concurrency` param of the `uvicorn.run` function represents the maximum number of concurrent connections a worker can handle.

PR #1188, opened 2020-06-30
Add retries for resource tagging

deliahu:


cli/cmd/cluster.go
```diff
@@ -737,14 +730,30 @@ func createBucketIfNotFound(awsClient *aws.Client, bucket string) error {
 			fmt.Print("\n\n")
 			return err
 		}
-		fmt.Println(" ✓")
 	} else {
-		fmt.Println("￮ using existing s3 bucket:", bucket, "✓")
+		fmt.Print("￮ using existing s3 bucket:", bucket)
 	}
+
+	numTaggingAttempts := 0
```
vishalbollu, line 737: nit: any reason not to use `for i := 0; i++; i < 10` style for loop here?

cli/cmd/cluster.go
```diff
@@ -756,11 +765,26 @@ func createLogGroupIfNotFound(awsClient *aws.Client, logGroup string) error {
...
 			return err
 		}
-		fmt.Println(" ✓")
 	} else {
-		fmt.Println("￮ using existing cloudwatch log group:", logGroup, "✓")
+		fmt.Print("￮ using existing cloudwatch log group:", logGroup)
 	}
 
+	numTaggingAttempts := 0
+	for true {
```
vishalbollu, line 773: Same comment as above

cli/cmd/cluster.go
vishalbollu, line 737: drop a comment explaining why retries may be necessary

deliahu: No particular reason, that was just how it occurred to me to implement it. How does this one look?

```go
for i := 0; i < 10; i++ {
	err = awsClient.TagLogGroup(logGroup, tags)
	if err == nil {
		fmt.Println(" ✓")
		return nil
	}
	if !aws.IsErrCode(err, cloudwatchlogs.ErrCodeResourceNotFoundException) {
		break
	}
}

fmt.Print("\n\n")
return err
```

Issue #1130, opened 2020-06-16
Add YOLOv5 example in PyTorch


RobertLucian:
#### Description

Add YOLOv5 example using the PyTorch implementation from https://github.com/ultralytics/yolov5.

The example should support prediction for video or image-based formats.

dsuess:
I am happy to give this a shot as I've got some experience with both this particular model in ONNX and video-predictions in Cortex. 

What kind of output did you envision for the video part? Should we add a simple tracker and return tracks?

RobertLucian:
@dsuess yes, you can go ahead with this one. If you've got any questions along the way, please don't hesitate to give us a shout. I think going with the tracks (bounding boxes) overlayed over the actual video should be more interesting to the end-user when it's run. Otherwise, a client would be needed to draw the bounding boxes. What do you think?

I think a good template would be this example:
https://github.com/cortexlabs/cortex/tree/master/examples/keras/document-denoiser

Thanks!

dsuess:
Ok, I'll just start with overlaying the video/image first and add tracking later.

Issue #954, opened 2020-04-08
Add post_predict() function to predictor API


deliahu:
#### Description

Add `post_predict()` to the Predictor API for any actions that should be taken after the request has been responded to.

#### Questions

* What information should be passed in? Probably the input and return value of `predict()`.

#### Motivation

Some potential user tasks (e.g. sending metrics, updating in-memory cache, etc) do not have to be on the critical path of responding to the request, so running them after the response is sent will decrease total latency.

Charlitos96:
Hi @deliahu, I'm looking for an Issue to do my first contribution, can you give me more details about what you want to do in `post_predict()`??
Also, can you confirm me that the file to be modified is this [predictor.py](https://github.com/cortexlabs/cortex/blob/dd9d8b3967f1046489253dfeb82ed0b408ef2f71/pkg/workloads/cortex/lib/type/predictor.py) ?

deliahu:
@Charlitos96 I appreciate your interest in contributing!

Yes, we will need to add `post_predict()` as an optional function in `PYTHON_CLASS_VALIDATION`, `TENSORFLOW_CLASS_VALIDATION`, and `ONNX_CLASS_VALIDATION` in [predictor.py](https://github.com/cortexlabs/cortex/blob/master/pkg/workloads/cortex/lib/type/predictor.py#L166) (you can add a new top-level key to the dictionary, similar to "required", but called "optional"). Then the validation should be automatically handed.

I think a reasonable schema is:

```python
***_CLASS_VALIDATION = {
    "required": [
        # ...
    ],
    "optional": [
        {
            "name": "predict",
            "required_args": ["self"],
            "optional_args": ["response", "payload", "query_params", "headers"],
        },
    ],
}
```

Where the args are the same as `predict()`, plus it also includes the return value from `predict()` (called `response`). Do you think that makes sense?

Then, the user's `post_predict()` function will be added to the [background tasks](https://www.starlette.io/background/) (if it's defined) in [serve.py](https://github.com/cortexlabs/cortex/blob/master/pkg/workloads/cortex/serve/serve.py#L184) at some point after `predictor_impl.predict(**args)` (since we need to pass in the result). We currently only add background tasks if monitoring is enabled (see [here](https://github.com/cortexlabs/cortex/blob/master/pkg/workloads/cortex/serve/serve.py#L210-L213)), so the logic of setting the background tasks in `predict(request: Request)` will have to be updated a little, but not much. 

[...]

PR #1201, opened 2020-07-03
Add yolov5-youtube example

dsuess:
addresses #1130

This is work in progress. I opened the PR to discuss some details.

Here's what this example does: 
- you send a request with a youtube URL 
- the predictor runs a YoloV5 model in ONNX format over it
- the overlayed video is returned to the user

This demonstrates both, how to load any Yolo model from the ultralytics repo and how to process video pretty easily.

I am using  ffmpeg-python (instead of the more standard opencv/scikkit-video) since it allows you to do large parts of the pre-processing (resizing & padding) in a separate process. The reason I am using youtube-dl (instead of passing in a video directly) is that we don't have to provide an example video ourselves.

Would this be a good example? Any suggestions for improvements? How do you want to store the ONNX file?

Here's what still needs to be done:
- [X] document how to export the ONNX model from the ultralytics repo. There's a minor change that makes our life much easier that I'll need to document.
- [X] improve the overlay to show class names
- [X] document & clean up the python code (e.g. delete intermediate files)
- [ ] ~update the documentation on the website~
- [ ] ~a bit of performance optimization~
- [ ] ~async the video-download?~

---

checklist:

- [ ] update examples


deliahu:
@dsuess this is really awesome!! I like how simple the API is: send a youtube link, and get back and annotated video!

> I am using ffmpeg-python (instead of the more standard opencv/scikkit-video) since it allows you to do large parts of the pre-processing (resizing & padding) in a separate process.

Does this help improve latency or just throughput? Are you running the API with `processes_per_replica=1` and `threads_per_process=1`? (note that `processes_per_replica` refers just to the uvicorn processes; ffmpeg will spawn additional processes). I'd be curious to understand how all of the threading/parallelism interplays, but also, there's no need to spend a lot of time on it, since it's just an example and not running in production :)

> The reason I am using youtube-dl (instead of passing in a video directly) is that we don't have to provide an example video ourselves.

Yes, I agree, this makes for a nice example!

> Would this be a good example? Any suggestions for improvements? 

Yes, I think this is a great example! I can't think of any suggestions for improvements at the moment.

> How do you want to store the ONNX file?

I will add it to our `cortex-examples` bucket; feel free to send me the model when it's ready (via google drive, dropbox, etc), and I can upload it there.

> async the video-download?

This is also an interesting one, similar to the discussion about ffmpeg. My intuition says that downloading the video async could increase throughput (and not latency) assuming `processes_per_replica>1` and/or `threads_per_process>1`. Like before, there's no need to spend a lot of time on this for the example's sake.

Thanks again for adding this!

dsuess:
> Does this help improve latency or just throughput? Are you running the API with `processes_per_replica=1` and `threads_per_process=1`? (note that `processes_per_replica` refers just to the uvicorn processes; ffmpeg will spawn additional processes). I'd be curious to understand how all of the threading/parallelism interplays, but also, there's no need to spend a lot of time on it, since it's just an example and not running in production :)

To be honest, the main reason for using ffmpeg-python was that I didn't want to implement the resize-padding for the hundredth time 😄 

I think it improves both (especially when you run on a GPU) since it allows you to run model inference and data preprocessing (decoding, resizing) in parallel. But I haven't done any tests and there are definitely better solutions if you actually need to optimize either performance metric.

> > async the video-download?
> 
> This is also an interesting one, similar to the discussion about ffmpeg. My intuition says that downloading the video async could increase throughput (and not latency) assuming `processes_per_replica>1` and/or `threads_per_process>1`. Like before, there's no need to spend a lot of time on this for the example's sake.

You're right on this one, this would only benefit throughput. And just increasing either `process_per_replica` or `threads_per_replica will probably be the better solution anyway

dsuess:
OK, I've send through the ONNX file and fixed the things that needed fixing. So if someone with the necessary permissions can a) upload the ONNX file and change the path in the config and b) fix the failing test, we're good to go IMO.

RobertLucian:
@dsuess Thank you a lot for adding this!

I tried pushing to your branch, but apparently, I don't have write permissions. I think this might be because `Allow edits from maintainer` hasn't been checked. This option should be somewhere on the right-hand side of the PR I think.

Also, I'd want to test this over the weekend. For that, do you have a public link to your ONNX model I could use?

dsuess:
The box is ticked as far as I can tell, so not sure what's going on.

For the ONNX file: https://drive.google.com/file/d/1p0nbSHUFpZhp6RxR2scNFaMk0ANHu_1e/view?usp=sharing

I've also included the steps how to create that ONNX file from the original repo, so you could also try to reproduce those setps

RobertLucian:
@dsuess thanks for the link. As for the write access, that was a false alarm (an old peculiarity of my setup). It now works!

examples/onnx/yolov5-videos/README.md
```diff
@@ -0,0 +1,57 @@
+# YoloV5 Detection model
+
+This example deploys a detection model trained using [ultralytics' yolo repo](https://github.com/ultralytics/yolov5) using ONNX.
+We'll use the `yolov5s` model as an example here.
+In can be used to run inference on youtube videos and returns the annotated video with bounding boxes.
+
+The example can be run on both CPU and on GPU hardware.
```
deliahu, line 7: Nice!

examples/onnx/yolov5-videos/README.md
```diff
@@ -0,0 +1,57 @@
...
+Originally, the ultralytics repo does not export postprocessing steps of the model, e.g. the conversion from the raw CNN outputs to bounding boxes.
+With newer ONNX versions, these can be exported as part of the model making the deployment much easier.
+
+With this modified script, the ONNX graph used for this example has been exported using
+```
+python models/export.py --weights weights/yolov5s.pt --img 416 --batch 1
+```
+
+
+## Sample Prediction
```
deliahu, line 33: Since people will probably try to use the example with the pre-exported model before exporting their own, how do you feel about switching the order of this and the previous section?

examples/onnx/yolov5-videos/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Deploy the model by running:
+
+```bash
+cortex deploy
+```
+
+And wait for it to become live by tracking its status with `cortex get --watch`.
+
+Once the API has been successfully deployed, export the API's endpoint for convenience. You can get the API's endpoint by running `cortex get youtube-yolov5`.
```
deliahu, line 43: nit: what do you think about calling it `yolov5-youtube` instead? (we would have to update it in `cortex.yaml` as well)

examples/onnx/yolov5-videos/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+When making a prediction with [sample.json](sample.json), [this](https://www.youtube.com/watch?v=aUdKzb4LGJ) youtube video will be used.
+
+To make a request to the model:
+
+```bash
+curl "${ENDPOINT}" -X POST -H "Content-Type: application/json" -d @sample.json --output video.mp4
+```
+
+After a few seconds, `curl` will save the resulting video `video.mp4` in the current working directory.
```
deliahu, line 57: Awesome!

examples/onnx/yolov5-videos/cortex.yaml
```diff
@@ -0,0 +1,13 @@
+# WARNING: you are on the master branch, please refer to the examples on the branch that matches your `cortex version`
+
+- name: youtube-yolov5
```
deliahu, line 3: See my comment on `README.md` regarding potentially changing this to `yolov5-youtube`

examples/onnx/yolov5-videos/cortex.yaml
```diff
@@ -0,0 +1,13 @@
...
+  kind: SyncAPI
+  predictor:
+    type: onnx
+    path: predictor.py
+    model_path: yolov5s.onnx
```
deliahu, line 8: I've uploaded it to `s3://cortex-examples/onnx/yolov5-youtube/yolov5s.onnx` for now, and I can move it depending on what we decide regarding the API name

examples/onnx/yolov5-videos/cortex.yaml
```diff
@@ -0,0 +1,13 @@
...
+    config:
+      iou_threshold: 0.5
+      confidence_threshold: 0.3
+  compute:
+    gpu: 1
```
deliahu, line 13: Consider adding a comment on this line, e.g. `# This is optional, since the API will run on CPUs (albeit slower)`

examples/onnx/yolov5-videos/predictor.py
```diff
@@ -0,0 +1,178 @@
...
+            # 4 output tensors, the last three are intermediate values and
+            # not necessary for detection
+            output, *_ = self.client.predict(x[None])
+            boxes, class_ids, confidence = self.postprocess(output)
+            overlay_boxes(frame, boxes, class_ids, self.labels, self.color_map)
+            writer.write(frame)
+
+        del writer
+        os.remove(in_path)
+        # We cant remove out_path, so the deployment will run out of memory
```
deliahu, line 176: Will this cause the system to run out of memory or disk space?

deliahu, line 176: Would resolving https://github.com/cortexlabs/cortex/issues/954 allow us to delete this file?

examples/onnx/yolov5-videos/predictor.py
```diff
@@ -0,0 +1,178 @@
...
+        for frame in frame_reader(in_path, size=self.input_size):
+            x = (frame.astype(np.float32) / 255).transpose(2, 0, 1)
+            # 4 output tensors, the last three are intermediate values and
+            # not necessary for detection
+            output, *_ = self.client.predict(x[None])
+            boxes, class_ids, confidence = self.postprocess(output)
+            overlay_boxes(frame, boxes, class_ids, self.labels, self.color_map)
+            writer.write(frame)
+
+        del writer
```
deliahu, line 174: Would it make sense to place the cleanup steps in a `finally` clause? (that would also require the code above to be in a `try`)

deliahu:
@RobertLucian @dsuess I've uploaded the model to `s3://cortex-examples/onnx/yolov5-youtube/yolov5s.onnx` for now, and I can move it depending on what we decide regarding the API name

examples/onnx/yolov5-videos/README.md
dsuess: Nice catch! Fully agree

examples/onnx/yolov5-videos/predictor.py
dsuess: Disk space fortunately. #954 would fix it since we could remove the file there. I also considered using a combination of `BytesIO` and `StreamingResponse` to load the file into memory and delete it from disk. But since this is meant as an example, I think I prefer the current simpler version

examples/onnx/yolov5-videos/predictor.py
dsuess: Yes, the whole thing needs some proper error handling.

examples/onnx/yolov5-youtube/predictor.py
```diff
@@ -160,19 +166,21 @@ def postprocess(self, output):
...
+        with open(out_path, "rb") as f:
+            output_bytes = f.read()
+
         os.remove(in_path)
-        # We cant remove out_path, so the deployment will run out of memory
-        # sooner or later
-        return FileResponse(out_path)
+        os.remove(out_path)
+
+        return output_bytes
```
dsuess, line 186: What do you think about changing this to

```python
buf = BytesIO(output_bytes)
return StreamingResponse(buf, media_type="video/mp4")
```

With the current version, the returned mime-type is `application/octet-stream`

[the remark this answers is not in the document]
RobertLucian: Yep, I like this more. Pushing this now!

dsuess:
Yes, all good from my side

PR #1202, opened 2020-07-04
Fix python_path config field

deliahu:


pkg/lib/files/files.go
```diff
@@ -142,6 +142,10 @@ func WriteFile(data []byte, path string) error {
 	return nil
 }
 
+func IsAbs(path string) bool {
+	return strings.HasPrefix(path, "/") || strings.HasPrefix(path, "~/")
```
vishalbollu, line 146: It might be confusing if this returns true for tilde prefixed paths. Rename to IsAbsOrTildePrefixed

pkg/types/spec/project_files.go
```diff
@@ -17,7 +17,14 @@ limitations under the License.
 package spec
 
 type ProjectFiles interface {
-	GetConfigFilePath() string
-	GetAllPaths() []string
+	// Return all paths in the project, relative to the root
+	AllPaths() []string
+	// Return the contents of a file, given the path (relative to the root)
```
vishalbollu, line 22: relative to the root -> relative to the project root

pkg/types/userconfig/api.go
```diff
@@ -38,7 +38,7 @@ type API struct {
 	Autoscaling    *Autoscaling    `json:"autoscaling" yaml:"autoscaling"`
 	UpdateStrategy *UpdateStrategy `json:"update_strategy" yaml:"update_strategy"`
 	Index          int             `json:"index" yaml:"-"`
-	FilePath       string          `json:"file_path" yaml:"-"`
+	FileName       string          `json:"file_name" yaml:"-"`
```
vishalbollu, line 41: this renaming is a not so obvious change but it makes a lot of sense

PR #1217, opened 2020-07-10
Fix OOM when reading files that don't fit into memory

RobertLucian:
When running `cortex deploy -e aws`, if the project's files don't fit into memory, then an OOM error is returned and it panics. In Go, when that happens, the panic is non-recoverable. Because of this, the program must ensure it doesn't encounter an OOM.

Wanted to use [runtime.ReadMemStats](https://golang.org/pkg/runtime/#ReadMemStats) to get the available allocable memory (looking at _HeapIdle_), but after running a few tests, it doesn't look like it's giving what we need. The stats are mainly about what the Go program uses (with GC and all that) and not about the system's available memory.

And it wasn't like we could just read the stats from `/proc` because that's platform-specific. [shirou/gopsutil](https://github.com/shirou/gopsutil) appears to be a well-maintained project that's cross-platform and gives us what we need.

---

checklist:

### DIFF
--- a/pkg/lib/files/errors.go
+++ b/pkg/lib/files/errors.go
@@ -24,18 +24,19 @@ import (
 )
 
 const (
-	ErrCreateDir         = "files.create_dir"
-	ErrDeleteDir         = "files.delete_dir"
-	ErrReadFormFile      = "files.read_form_file"
-	ErrCreateFile        = "files.create_file"
-	ErrReadDir           = "files.read_dir"
-	ErrReadFile          = "files.read_file"
-	ErrFileAlreadyExists = "files.file_already_exists"
-	ErrUnexpected        = "files.unexpected"
-	ErrFileDoesNotExist  = "files.file_does_not_exist"
-	ErrDirDoesNotExist   = "files.dir_does_not_exist"
-	ErrNotAFile          = "files.not_a_file"
-	ErrNotADir           = "files.not_a_dir"
+	ErrCreateDir                    = "files.create_dir"
+	ErrDeleteDir                    = "files.delete_dir"
+	ErrReadFormFile                 = "files.read_form_file"
+	ErrCreateFile                   = "files.create_file"
+	ErrReadDir                      = "files.read_dir"
+	ErrReadFile                     = "files.read_file"
+	ErrInsufficientMemoryToReadFile = "files.insufficient_memory_to_read_file"
+	ErrFileAlreadyExists            = "files.file_already_exists"
+	ErrUnexpected                   = "files.unexpected"
+	ErrFileDoesNotExist             = "files.file_does_not_exist"
+	ErrDirDoesNotExist              = "files.dir_does_not_exist"
+	ErrNotAFile                     = "files.not_a_file"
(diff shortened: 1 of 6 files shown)


pkg/lib/files/files.go
```diff
@@ -106,6 +111,19 @@ func ReadFileBytesErrPath(path string, errMsgPath string) ([]byte, error) {
 		return nil, err
 	}
 
+	var fileSizeBytes int64
+	if fileSizeBytes, err = CheckFileSizeBytes(path); err != nil {
```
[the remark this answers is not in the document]
RobertLucian: Good call!

pkg/lib/files/files.go
```diff
@@ -301,6 +319,23 @@ func CheckFileErrPath(path string, errMsgPath string) error {
 	return nil
 }
 
+func CheckFileSizeBytes(path string) (int64, error) {
```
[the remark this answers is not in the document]
RobertLucian: Yep! Done!

pkg/lib/files/files.go
```diff
@@ -527,6 +532,22 @@ func IgnoreSpecificFiles(absPaths ...string) IgnoreFn {
 	}
 }
 
+func ErrorOnBigFiles(path string, fi os.FileInfo) (bool, error) {
```
vishalbollu, line 535: I would restructure this function to follow the signature for `GitIgnoreFn` to return (IgnoreFn, error) and it takes in maxFileSize as a parameter.

RobertLucian: I think we should only return `IgnoreFn` and not `(IgnoreFn, error)`. There shouldn't be any error returned from a function like `ErrorOnBigFilesFn`. As for its parameters, I think we actually need 2 of them: `maxFileSizeBytes` and `maxMemoryUsagePercent`. This way, we also get rid of 2 non-`files.go`-specific variables.

pkg/lib/files/errors.go
```diff
@@ -87,6 +90,35 @@ func ErrorFileAlreadyExists(path string) error {
 	})
 }
 
+func ErrorInsufficientMemoryToReadFile(path string, fileSizeBytes, availableMemBytes int64) error {
```
vishalbollu, line 93: This function doesn't accept path

pkg/lib/files/files.go
```diff
@@ -552,13 +555,44 @@ func PromptForFilesAboveSize(size int, promptMsgTemplate string) IgnoreFn {
...
 		return false, nil
 	}
 }
 
+func ErrorOnBigFilesFn(maxFileSizeBytes int64, maxMemoryUsagePercent float64) IgnoreFn {
+	return func(path string, fi os.FileInfo) (bool, error) {
+		if !fi.IsDir() {
+			fileSizeBytes := fi.Size()
+			virtual, _ := mem.VirtualMemory()
+			if float64(fileSizeBytes) > float64(virtual.Available) ||
```
vishalbollu, line 570: `float64(fileSizeBytes) > float64(virtual.Available) ||` condition may not be necessary

PR #1230, opened 2020-07-16
Improve shorthand env flag detection

deliahu:


cli/cmd/root.go
```diff
@@ -175,20 +177,45 @@ func updateRootUsage() {
 	})
 }
 
-func wasEnvFlagProvided() bool {
+func getCmdShorthandFlags(cmd *cobra.Command) strset.Set {
```
RobertLucian, line 180: I wonder if we could remove this function `getCmdShorthandFlags` and instead just implement `wasEnvFlagProvided` slightly to do the same thing. `getCmdShorthandFlags` only appears to be used in `wasEnvFlagProvided`.

cli/cmd/root.go
```diff
@@ -175,20 +177,45 @@ func updateRootUsage() {
...
+	shorthands := strset.New()
+	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
+		if flag.Shorthand != "" {
+			shorthands.Add(flag.Shorthand)
+		}
+	})
+	return shorthands
+}
+
+func wasEnvFlagProvided(cmd *cobra.Command) bool {
```
RobertLucian, line 190: I'm thinking of something like:
```golang
func wasEnvFlagProvided(cmd *cobra.Command) bool {
	envFlagProvided := false
	cmd.Flags().VisitAll(func(flag *pflag.Flag) {
		if flag.Shorthand == "e" && flag.Changed {
			envFlagProvided = true
		}
	})

	return envFlagProvided
}
```

This should cover all scenarios: `-e <var>`, `-e=<var>`, `-e<var>`, `-we<var>`, `--env <var>`, `--env=<var>`. What do you think?

deliahu: Good catch, I will do it this way!

PR #1238, opened 2020-07-28
Check result for docker image

### LINKED ISSUE #1236: Image download shown as successful but see error: "No such image: cortexlabs/python-predictor-cpu:0.18.1"
#### Version

0.18.1

#### Description

On Mac OS (High Sierra) docker pull silently fails when downloading the python-predictor-cpu image:

```
￮ downloading docker image cortexlabs/python-predictor-cpu:0.18.1 .................................................................... ✓
/Users/rocco.cammisola/Documents/[PROJECT DIR]/cortex.yaml: [API ENDPOINT NAME]: Error response from daemon: No such image: cortexlabs/python-predictor-cpu:0.18.1
```

The output suggests that the download was successful but when I performed `docker pull cortexlabs/python-predictor-cpu:0.18.1` I found out that my docker VM had run out of space.

Performing `docker system prune` (or similar) to clear the VM resolved the issue.

#### Steps to reproduce

1. Set docker VM size on Mac to be smaller than the cortexlabs python predictor image
2. cortex deploy

#### Expected behavior

The download is not marked as successful (tick not shown) and the error about being out of space is reported.

#### Actual behavior

Download marked as successful, no error directly related to not being able to pull the image.

zouyee:
closes #1236

---

checklist:

- [x] update examples


pkg/lib/docker/docker.go
```diff
@@ -161,12 +161,20 @@ func PullImage(image string, encodedAuthConfig string, pullVerbosity PullVerbosi
...
+				fmt.Print(" ✓\n")
+			} else {
+				fmt.Print(" x\n")
+			}
+		}()
 		dotCron := cron.Run(print.Dot, nil, 2*time.Second)
 		defer dotCron.Cancel()
 		// wait until the pull has completed
-		if _, err := ioutil.ReadAll(pullOutput); err != nil {
+		_, err = ioutil.ReadAll(pullOutput)
```
vishalbollu, line 176: @zouyee It looks like an error will only be thrown if the io reading fails. I could be wrong but it doesn't look like `ioutil.ReadAll(pullOutput)` would fail if the actual docker image pull fails.

Have you tried to reproduce the issue and test this solution? If you have, it would be great if you can share the error message. Ideally, the error message from docker should be displayed if the error message is easy to understand for the user. Otherwise, the error should be detected and a more understandable error should be thrown.

I think you may have to parse output from `pullOutput`, here is an example: https://stackoverflow.com/questions/44452679/golang-docker-api-parse-result-of-imagepull (instead of creating your own event struct you can use [jsonmessage.JSONMessage](https://godoc.org/github.com/docker/docker/pkg/jsonmessage).) Once the JSONMessages have been parsed, you can iterate through the messages and look for any JSONMessages with errors. From those errors, we can come up with a meaningful error message to expose to the user.

zouyee:
@vishalbollu PTAL

Issue #1218, opened 2020-07-13
Investigate validity of error message when re-deploying on a different CLI version


RobertLucian:
#### Description

Trying to re-deploy an API locally with a newer version of the CLI will return the following type of message:

```bash
api model was deployed using CLI version 0.17.1 but the current CLI version is 0.18.1; please run `cortex deploy` to redeploy the api or `cortex delete model` to delete the api`
```

As reported by @elliotzeitgold.

#### Expected behavior

Running any of the two suggested commands should work - and not return an error.

#### Actual behavior

Running any of the two suggested commands in the error message will fail with the same error message.

RobertLucian:
Part of the error message is indeed contradictory/recursive: `please run cortex deploy to redeploy the api or cortex delete model to delete the api`. Running `cortex deploy` or `cortex delete` would return the same error.

PR #1247, opened 2020-07-31
Fix recursive error message for deploy/delete CLI commands

RobertLucian:
Fixes https://github.com/cortexlabs/cortex/issues/1218.

---

checklist:


cli/local/api.go
```diff
@@ -27,17 +27,38 @@ import (
...
+		if errors.GetKind(err) == ErrCortexVersionMismatch {
+			encounteredVersionMismatch = true
+			if incompatibleVersion, err = GetVersionFromAPISpec(apiConfig.Name); err != nil {
+				return nil, "", err
+			}
+			if !deployDisallowPrompt {
+				prompt.YesOrExit(
+					fmt.Sprintf(
+						"api %s was deployed using CLI version %s but the current CLI version is %s; "+
+							"re-deploying %s with current CLI version %s might break the CLI; any cached models won't be deleted\n\n"+
```
vishalbollu, line 52: `re-deploying %s with current CLI version %s might break the CLI` -> `re-deploying %s with current CLI version %s might yield an unexpected outcome`

cli/cmd/delete.go
```diff
@@ -66,8 +66,8 @@ var _deleteCmd = &cobra.Command{
 				exit.Error(err)
 			}
 		} else {
-			// local only supports deploying 1 replica at a time so _flagDeleteForce will be ignored
-			deleteResponse, err = local.Delete(args[0], _flagDeleteKeepCache)
+			// local only supports deploying 1 replica at a time, so _flagDeleteForce is only useful when APIs were deployed with different CLI version
```
vishalbollu, line 69: `is only useful when APIs were deployed with different CLI version` -> `is only useful when attempting to delete an API that has been deployed with different CLI version`

cli/local/delete.go
```diff
@@ -19,29 +19,58 @@ package local
...
+		if errors.GetKind(err) == ErrCortexVersionMismatch {
+			var incompatibleVersion string
+			if incompatibleVersion, err = GetVersionFromAPISpec(apiName); err != nil {
+				return schema.DeleteResponse{}, err
+			}
+			if !deleteForce {
+				prompt.YesOrExit(
+					fmt.Sprintf(
+						"api %s was deployed using CLI version %s but the current CLI version is %s; "+
+							"deleting %s with current CLI version %s might break the CLI; any cached models won't be deleted\n\n"+
```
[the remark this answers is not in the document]
RobertLucian: I'm thinking instead of `download an older version of the CLI using github.com/cortexlabs/cortex/blob/VERSION/get-cli.sh, delete the API using the older version of the CLI and then redeploy the API using the latest version of the CLI` -> `it is recommended to download version %s of the CLI from https://docs.cortex.dev/v/<minor-mismatched-version>/install, delete the API using version <minor-mistmatched-version> of the CLI and then re-deploy the API using the latest version of the CLI`.

This can happen regardless of whether the version the API has been deployed with is newer or older. And I think specifying the link to the documentation for installing the CLI is better. Same as for the other comment down below. Thoughts?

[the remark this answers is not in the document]
vishalbollu: Yes, you are correct, Ship it!

cli/local/delete.go
```diff
@@ -18,30 +18,62 @@ package local
...
+func Delete(apiName string, keepCache, deleteForce bool) (schema.DeleteResponse, error) {
 	_, err := docker.GetDockerClient()
 	if err != nil {
 		return schema.DeleteResponse{}, err
 	}
 
 	var apiSpec *spec.API = nil
-	if !keepCache {
-		if apiSpec, err = FindAPISpec(apiName); err != nil {
+	if apiSpec, err = FindAPISpec(apiName); err != nil {
```
vishalbollu, line 38: It looks like this error isn't being handled if `errors.GetKind(err) != ErrCortexVersionMismatch` and `keepCache == false`.

cli/local/api.go
```diff
@@ -88,6 +111,13 @@ func UpdateAPI(apiConfig *userconfig.API, configPath string, projectID string, a
 	}
 
 	if prevAPISpec == nil && len(prevAPIContainers) == 0 {
+		if encounteredVersionMismatch {
+			return newAPISpec, fmt.Sprintf(
+				"re-creating api %s with current CLI version %s",
```
vishalbollu, line 116: nit: `re-creating` -> `re-deploying`

cli/local/api.go
```diff
@@ -27,17 +27,40 @@ import (
...
 var _deploymentID = "local"
 
-func UpdateAPI(apiConfig *userconfig.API, configPath string, projectID string, awsClient *aws.Client) (*spec.API, string, error) {
+func UpdateAPI(apiConfig *userconfig.API, configPath string, projectID string, deployDisallowPrompt bool, awsClient *aws.Client) (*spec.API, string, error) {
+	var incompatibleVersion string
+	encounteredVersionMismatch := false
 	prevAPISpec, err := FindAPISpec(apiConfig.Name)
 	if err != nil {
-		if errors.GetKind(err) != ErrAPINotDeployed {
+		if errors.GetKind(err) == ErrCortexVersionMismatch {
```
vishalbollu, line 43: I believe `ErrCortexVersionMismatch` will be raised if the Cortex versions don't match exactly. Let's say you've deployed an API with 0.18.0 and you upgraded your CLI to 0.18.1. You try to redeploy but you run into a version mismatch:

```
api iris-classifier was deployed using CLI version 0.18.0 but the current CLI version is 0.18.1; re-deploying iris-classifier with current CLI version 0.18.1 might yield an unexpected outcome; any cached models won't be deleted

it is recommended to download version 0.18.0 of the CLI from https://docs.cortex.dev/v/0.18/install, delete the API using version 0.18.0 of the CLI and then re-deploy the API using the latest version of the CLI

do you still want to re-deploy?
```

If you install the CLI docs by following `https://docs.cortex.dev/v/0.18/install`, you will still download 0.18.1 CLI because you will download the latest patch version of that minor version. You would run into the same prompt again. Maybe this prompt should only be shown if the minor versions don't match. In a scenario where minor versions do match, I think we can assume that it is safe to re-deploy or delete. Let me know what you think.

cli/local/delete.go
RobertLucian: The reason why the error isn't handled is that `apiSpec` is only required for `DeleteCachedModels` function as `apiSpec.ModelIDS()`.

You could say that `DeleteAPI` also has to call `FindAPISpec` too, but then, if we were to just return the error, then the containers wouldn't get deleted, which is a part of what `DeleteAPI` does. I'd say the thinking would be like this: _"don't care whether FindAPISpec fails when the cache is to be kept because I'll do the DeleteAPI call which does the proper clean-up anyway and if it errors, then I'll get the correct error (wrapped and all that)"_.

We could add `DeleteAPI` and then just return the error in that `if` branch. As it is right now, that would have the same effect. What do you think?

cli/local/api.go
RobertLucian: Done. What do you think of doing it for `creating` -> `deploying` too?

cli/local/api.go
RobertLucian: Agreed. The prompt should only be shown when minor versions don't match. Patches shouldn't change the functionality, but just add fixes.

cli/local/api.go
vishalbollu: Good point, instead of `re-deploying` just `creating api %s with current CLI version %s` is fine

cli/local/api.go
```diff
@@ -27,17 +27,41 @@ import (
...
+			encounteredVersionMismatch = true
+			if incompatibleVersion, err = GetVersionFromAPISpec(apiConfig.Name); err != nil {
+				return nil, "", err
+			}
+
+			minorVersionMismatch := strings.Join(strings.Split(incompatibleVersion, ".")[:2], ".")
```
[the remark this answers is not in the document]
RobertLucian: Done.

cli/local/api.go
RobertLucian: Done.

PR #1253, opened 2020-08-08
Fix API error/status messages

RobertLucian:
This PR addresses three issues:

1. When no APIs are deployed, doing a `cortex get` (with no `-e` specified) will print the API Splitter's column names instead of outputting `no apis are deployed` message.

1. For API kind `SyncAPI`, when an error is encountered, the API's identity gets printed twice instead of once. That identity has `<filepath>: <api-name> <api-kind>` format.

1. Missing `networking` key when an error with the `endpoint` field is encountered with the `APISplitter` kind.

---

checklist:

### DIFF
--- a/pkg/operator/resources/validations.go
+++ b/pkg/operator/resources/validations.go
@@ -93,10 +93,10 @@ func ValidateClusterAPIs(apis []userconfig.API, projectFiles spec.ProjectFiles) 
 		api := &apis[i]
 		if api.Kind == userconfig.SyncAPIKind {
 			if err := spec.ValidateAPI(api, projectFiles, types.AWSProviderType, config.AWS); err != nil {
-				return errors.Wrap(err, api.Identify())
+				return err
 			}
 			if err := validateK8s(api, virtualServices, maxMem); err != nil {
-				return errors.Wrap(err, api.Identify())
+				return err
 			}
 
 			if !didPrintWarning && api.Networking.LocalPort != nil {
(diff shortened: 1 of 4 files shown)


pkg/operator/resources/validations.go
```diff
@@ -93,10 +93,10 @@ func ValidateClusterAPIs(apis []userconfig.API, projectFiles spec.ProjectFiles)
 		api := &apis[i]
 		if api.Kind == userconfig.SyncAPIKind {
 			if err := spec.ValidateAPI(api, projectFiles, types.AWSProviderType, config.AWS); err != nil {
-				return errors.Wrap(err, api.Identify())
```
deliahu, line 96: I think the wrapping of `api.Identify()` should remain at this level, and should be removed from all levels below (e.g. `ValidateAPISplitter()`, and presumably somewhere else which prompted you to make this change).

In general, we've found the most maintainable pattern is to wrap error context info at the highest level possible.

RobertLucian: As per our conversation, the requested changes have been addressed.

pkg/operator/resources/validations.go
```diff
@@ -96,7 +96,7 @@ func ValidateClusterAPIs(apis []userconfig.API, projectFiles spec.ProjectFiles)
 				return errors.Wrap(err, api.Identify())
 			}
 			if err := validateK8s(api, virtualServices, maxMem); err != nil {
-				return errors.Wrap(err, api.Identify())
+				return err
```
deliahu, line 99: Since the api context is being created in this function and is being passed in, would it make sense to wrap the error with `api.Identify()` here instead of below?

pkg/operator/resources/validations.go
```diff
@@ -112,7 +112,7 @@ func ValidateClusterAPIs(apis []userconfig.API, projectFiles spec.ProjectFiles)
 				return errors.Wrap(err, api.Identify())
 			}
 			if err := validateEndpointCollisions(api, virtualServices); err != nil {
-				return errors.Wrap(err, api.Identify())
```
deliahu, line 115: (same comment as above)

pkg/operator/resources/validations.go
RobertLucian: I was running on the impression that `validateEndpointCollisions` had the API identity placed as the 3rd argument for `errors.Wrap` function, thus thinking that it's not possible to move it up the call stack, LOL. Moved it up.

Issue #923, opened 2020-03-27
Support different cuda versions in Python Predictor base image


deliahu:
#### Motivation

As shown in [TensorFlow's compatibility matrix](https://www.tensorflow.org/install/source#gpu), different versions of TensorFlow rely on different versions of CUDA. The same is true for Pytorch.

#### Solution

Build multiple versions of `python-gpu-slim`:

```yaml
cortexlabs/python-predictor-gpu-slim:<cortex_version>-cuda10.0
cortexlabs/python-predictor-gpu-slim:<cortex_version>-cuda10.1
cortexlabs/python-predictor-gpu-slim:<cortex_version>-cuda10.2
cortexlabs/python-predictor-gpu-slim:<cortex_version>-cuda11.0
```

And for the full version of the image `python-gpu`:

```yaml
cortexlabs/python-predictor-gpu:<cortex_version>
```

Here is an example of using a Docker build arg:

```dockerfile
ARG CUDA_VERSION=10.1
FROM nvidia/cuda:$CUDA_VERSION-cudnn7-devel-ubuntu18.04
```

PR #1263, opened 2020-08-15
Support different CUDA versions for the slim Python Predictor image

### LINKED ISSUE #1254: Update driver version of EC2 image
#### Description

The current docker-images use CUDA 10.1, since that seems to be the least common denominator between all the packages used. However, when we tried to run pytorch with CUDA 10.2 (necessary because of pytorch extensions compiled with 10.2), we're getting a runtime error
```
RuntimeError: cuda runtime error (35) : CUDA driver version is insufficient for CUDA runtime version ...
```

#### Motivation

Pytorch 1.5 and 1.6 come by default compiled against CUDA 10.2 so it would be nice if the underlying EC2 image's driver would be compatible with CUDA 10.2.

RobertLucian:
Closes https://github.com/cortexlabs/cortex/issues/923, closes #1254

---

checklist:


build/build-image.sh
```diff
@@ -41,14 +41,37 @@ dir=$1
...
+if [ "${image}" == "python-predictor-gpu" ]; then
+  cuda=("10.0" "10.1" "10.2" "11.0")
+  cudnn=("7" "7" "7" "8")
+
+  # the only tag for the fat version is for CUDA 10.1 and CUDNN 7
+  docker build "$ROOT" \
+    -f $dir/Dockerfile \
+    --build-arg CUDA_VERSION=${cuda[1]} \
+    --build-arg CUDNN=${cudnn[1]} \
+    -t cortexlabs/${image}:${CORTEX_VERSION}-cuda${cuda[1]}
```
deliahu, line 65: I'm thinking maybe this section is not necessary, since this image will already have been built above, and perhaps we don't need the cuda tag on the fat version (since there is only one). What do you think? We can discuss it live too if you prefer

build/build-image.sh
```diff
@@ -41,14 +41,37 @@ dir=$1
...
+
+  for i in ${!cudnn[@]}; do
+    if [ "$slim" == "true" ]; then
```
deliahu, line 68: nit: consider switching the order of these two if's

build/push-image.sh
```diff
@@ -44,3 +44,16 @@ docker push cortexlabs/${image}:${CORTEX_VERSION}
 if [ "$slim" == "true" ]; then
   docker push cortexlabs/${image}-slim:${CORTEX_VERSION}
 fi
+
+if [ "$image" == "python-predictor-gpu" ]; then
+  cuda=("10.0" "10.1" "10.2" "11.0")
+
+  # the only tag for the fat version is for CUDA 10.1 and CUDNN 7
+  docker push cortexlabs/${image}:${CORTEX_VERSION}-cuda${cuda[1]}
```
deliahu, line 52: (same comment as above about potentially removing this)

build/push-image.sh
```diff
@@ -44,3 +44,16 @@ docker push cortexlabs/${image}:${CORTEX_VERSION}
...
+
+  for i in ${!cuda[@]}; do
+    if [ "$slim" == "true" ]; then
```
deliahu, line 55: (same comment regarding switching the order of the if statements)

build/build-image.sh
```diff
@@ -41,14 +41,37 @@ dir=$1
 image=$2
 
 docker build "$ROOT" \
```
deliahu, line 43: As discussed, we decided to not build e.g. `cortexlabs/python-predictor-gpu-slim:<version>`

build/push-image.sh
```diff
@@ -44,3 +44,16 @@ docker push cortexlabs/${image}:${CORTEX_VERSION}
 if [ "$slim" == "true" ]; then
```
deliahu, line 44: As discussed, we decided to not push e.g. `cortexlabs/python-predictor-gpu-slim:<version>`

build/push-image.sh
```diff
@@ -42,5 +42,13 @@ echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 docker push cortexlabs/${image}:${CORTEX_VERSION}
 
 if [ "$slim" == "true" ]; then
-  docker push cortexlabs/${image}-slim:${CORTEX_VERSION}
+  if [ "$image" == "python-predictor-gpu" ]; then
+    cuda=("10.0" "10.1" "10.2" "11.0")
+
+    for i in ${!cuda[@]}; do
+      docker push cortexlabs/${image}-slim:${CORTEX_VERSION}-cuda${cuda[$i]}
+    done
```
deliahu, line 50: Another options is to do something like this, feel free to choose whichever you prefer:

```bash
for cuda in 10.0 10.1 10.2 11.0; do
  docker push cortexlabs/${image}-slim:${CORTEX_VERSION}-cuda${cuda}
done
```

docs/deployments/system-packages.md
```diff
@@ -50,7 +50,7 @@ Cortex's base Docker images are listed below. Depending on the Cortex Predictor
 
 <!-- CORTEX_VERSION_BRANCH_STABLE x6 -->
 * Python Predictor (CPU): `cortexlabs/python-predictor-cpu-slim:master`
-* Python Predictor (GPU): `cortexlabs/python-predictor-gpu-slim:master`
+* Python Predictor (GPU): `cortexlabs/python-predictor-gpu-slim:master-cuda10.1` (also available in cuda10.0, cuda10.2, and cuda11.0)
```
deliahu, line 53: I updated these docs, let me know if it looks ok to you

build/push-image.sh
RobertLucian: Yep, I like this one too. I went with it.

docs/deployments/system-packages.md
RobertLucian: One thing to keep in mind: if we get any users in the future asking about the CUDNN versions, then at that point, it may be necessary to revisit this. So for now, I think it's okay the way it is.

PR #1272, opened 2020-08-18
Update docs sidebar

deliahu:


docs/summary.md
```diff
@@ -7,7 +7,6 @@
 * [Examples](https://github.com/cortexlabs/cortex/tree/master/examples)  <!-- CORTEX_VERSION_MINOR -->
 * [Chat with us](https://gitter.im/cortexlabs/cortex)
 * [Email us](mailto:hello@cortex.dev)
-* [We're hiring](https://angel.co/cortex-labs-inc/jobs)
```
RobertLucian, line 10: About "Chat with us",  "Email us" and "We're hiring", I'm thinking of having these moved to an "About" section at the bottom of the sidebar.

Another alternative would be to just move "Chat with us" and "Email us" to this "About" section and remove "We're hiring" altogether - we would only have it on cortex.dev, which should be okay.

And then there's this alternative, where "We're hiring" is removed altogether and that's it.

I'm slightly leaning towards the option(s) that would remove as many bullet points as possible from the sidebar's top (first and second). Thinking that the actual documentation should be the first thing the user sees.

What do you think? Let me know if we should discuss about this.

deliahu: Thanks for the suggestions! Let's discuss it live

PR #1311, opened 2020-08-27
Fix tensor shape validation for the TensorFlow predictor

### LINKED ISSUE #1310: Fix scalar and no-rank/no-shape tensor validation when using the TensorFlow predictor
#### Description

2 properties of a tensor (in TensorFlow) are the rank and shape. Right now, when using the TensorFlow predictor, only the 2 following kinds of tensors are supported:

* Tensors with known rank and shape. Such as `(3, 224, 224`).
* Tensors with known rank and unknown shape. Such as `(3, -1, -1)`.

#### Reproduce

When tensors have the following properties, the validation will fail, when in reality, it shouldn't:

* Tensors with no rank and empty shape - for scalars. Such as `()`.
* Tensors with unknown rank and shape. Such as `(-1, -1, -1, ...)`.

When an input tensor has the above properties, the following error pops up the logs `cortex logs <api-name>`.

```text
Traceback (most recent call last):
File "/src/cortex/serve/serve.py", line 282, in start_fn
tf_serving_host=tf_serving_host, tf_serving_port=tf_serving_port
File "/src/cortex/lib/type/predictor.py", line 72, in initialize_client
client = TensorFlowClient(tf_serving_address, self.models)
File "/src/cortex/lib/client/tensorflow.py", line 48, in init
self._signatures, get_model_signature_map(models)
File "/src/cortex/lib/client/tensorflow.py", line 186, in extract_signatures
signature_defs[model_name], signature_keys[model_name], model_name,
File "/src/cortex/lib/client/tensorflow.py", line 246, in extract_signature
"shape": [int(dim["size"]) for dim in input_metadata["tensorShape"]["dim"]],
KeyError: 'dim'
```

#### Additional

References that show that no-rank/no-shape and scalar tensors do indeed exist and are used:

* The first 2 paragraphs from https://pgaleone.eu/tensorflow/2018/07/28/understanding-tensorflow-tensors-shape-static-dynamic/.
* https://www.tensorflow.org/guide/tensor

RobertLucian:
Fixes #1310.

---

checklist:

### DIFF
--- a/pkg/workloads/cortex/lib/client/tensorflow.py
+++ b/pkg/workloads/cortex/lib/client/tensorflow.py
@@ -249,7 +249,7 @@ def extract_signature(signature_def, signature_key, model_name):
             # a scalar with rank 0 and empty shape
             shape = "scalar"
         elif input_metadata["tensorShape"].get("unknownRank", False):
-            #  unknown rank and shape
+            # unknown rank and shape
             shape = "unknown"
         elif input_metadata["tensorShape"].get("dim", None):
             # known rank and known/unknown shape
--- a/pkg/workloads/cortex/lib/client/tensorflow.py
+++ b/pkg/workloads/cortex/lib/client/tensorflow.py
@@ -280,6 +280,9 @@ def create_prediction_request(signature_def, signature_key, model_name, model_in
         elif signature_def[signature_key]["inputs"][column_name]["tensorShape"].get(
             "unknownRank", False
         ):
+            # unknownRank is set to True if the model input has no rank
+            # it may lead to an undefined behavior if unknownRank is only checked for its presence
+            # so it also gets to be tested against its value
             shape = "unknown"
         else:
             shape = []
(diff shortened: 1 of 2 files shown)


pkg/workloads/cortex/lib/client/tensorflow.py
```diff
@@ -258,19 +275,35 @@ def create_prediction_request(signature_def, signature_key, model_name, model_in
...
 
     for column_name, value in model_input.items():
-        shape = []
-        for dim in signature_def[signature_key]["inputs"][column_name]["tensorShape"]["dim"]:
-            shape.append(int(dim["size"]))
+        if signature_def[signature_key]["inputs"][column_name]["tensorShape"] == {}:
+            shape = "scalar"
+        elif signature_def[signature_key]["inputs"][column_name]["tensorShape"].get(
+            "unknownRank", False
+        ):
```
deliahu, line 282: nit: is this equivalent to `elif "unknownRank" in signature_def[signature_key]["inputs"][column_name]["tensorShape"]:`, and if so, do you prefer it?

RobertLucian: Yes, that is the equivalent. There's one little thing I noticed: `unknownRank` always seems to be set to `True` when no rank is specified, and so, I think it's also good to check the key's value. I'm not sure if there's a case when that's set to `False`.

RobertLucian: I could add a comment to describe the reason.

deliahu: That makes sense, yes, it is better to be safe and check the value too. Thanks!

PR #1332, opened 2020-09-07
Improve CNI supported instances check

### LINKED ISSUE #809: Support c6g and r6g instances
#### Description

* [AWS blog post](https://aws.amazon.com/about-aws/whats-new/2019/12/announcing-new-amazon-ec2-m6g-c6g-and-r6g-instances-powered-by-next-generation-arm-based-aws-graviton2-processors/?trk=ls_card)
* [AWS CNI supported instance types](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/vpc_ip_resource_limit.go) (change to desired release tag or branch)

### LINKED ISSUE #1331: Improve CNI supported instances check
#### Description

Currently, there are instance types that are supported by Cortex which don't pass validation, because validations are running against an older version of the CNI. Simply updating the CNI version for the validation check is not a great solution, because AWS will continuously be updating the default CNI on EKS. A better approach is to maintain a list of the instance types that are supported by the current default CNI version.

deliahu:
closes #1331, closes #809

Note that the compatible instance types for the default CNI version has been uploaded to `https://cortex-public.s3-us-west-2.amazonaws.com/cli-assets/cni_supported_instances.txt`, and should be periodically updated to reflect reality.

---

checklist:


pkg/types/clusterconfig/clusterconfig.go
```diff
@@ -650,14 +657,47 @@ func (cc *Config) Validate(awsClient *aws.Client) error {
...
+	if err := checkCNISupport(instanceMetadata.Type); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Check the instance type against the list of supported instance types for the current default CNI version
+// Returns nil if unable to perform the check successfully
+func checkCNISupport(instanceType string) error {
```
vishalbollu, line 675: This function overloads nil to represent a successful CNI support check and an error check. I think we should return nil only when the CNI support is successful. We should leave error handling to the caller. In this case the caller would swallow any error except for `ErrInstanceTypeNotSupported`.

vishalbollu (approved):

PR #1342, opened 2020-09-07
Display autoscaling group activity history when cluster up fails

### LINKED ISSUE #1340: Display autoscaling group activity history when cluster up fails
#### Description

There are some situations that cause `cortex cluster up` to fail which can only be identified by looking at the autoscaling group activity history. The relevant error information should be shown to the user in the terminal output when possible.

Implementing this would make fixing #1258 less urgent.

deliahu:
closes #1340

---

checklist:


deliahu:
Note that I have not tested this in the real world yet, since by the time I had implemented the code, spot capacity was available again.

cli/cmd/cluster.go
```diff
@@ -173,9 +173,69 @@ var _upCmd = &cobra.Command{
...
-			helpStr += fmt.Sprintf("\n* if your cluster started spinning up, please ensure that your CloudFormation stacks for this cluster have been fully deleted before trying to spin up this cluster again (you can delete your CloudFormation stacks from the AWS console: %s)", getCloudFormationURL(clusterConfig.ClusterName, *clusterConfig.Region))
+
+			eksCluster, err := awsClient.EKSClusterOrNil(clusterConfig.ClusterName)
+			if err != nil {
+				helpStr := "\ndebugging tips (may or may not apply to this error):"
+				helpStr += fmt.Sprintf("\n* if your cluster started spinning up but was unable to provision instances, additional error information may be found in the activity history of your cluster's autoscaling groups (select each autoscaling group and click the \"Activity\" or \"Activity History\" tab): https://console.aws.amazon.com/ec2/autoscaling/home?region=%s#AutoScalingGroups:", *clusterConfig.Region)
```
vishalbollu, line 180: is there anyway to filter for the asgs relevant to this cluster?

pkg/lib/aws/autoscaling.go
```diff
@@ -59,3 +63,37 @@ func (c *Client) AutoscalingGroups(tags map[string]string) ([]*autoscaling.Group
 
 	return asgs, nil
 }
+
+// Return all activities for an ASG which do not have the "Successful" status, sorted from most to least recent
+// Also returns true if the most recent activity for the ASG is "Successful", or if there are no activities for the ASG
+func (c *Client) AutoscalingGroupUnsuccessfulActivities(asgName string) ([]*autoscaling.Activity, bool, error) {
```
vishalbollu, line 69: It looks like this function can return the most recent activity. The caller can inspect the most recent activity and determine if it is successful.

vishalbollu (approved):

cli/cmd/cluster.go
deliahu: I just tried, and I was not able to figure out how to do it

vishalbollu (approved):

PR #1355, opened 2020-09-15
Validate AWS tags

deliahu:
checklist:

### DIFF
--- a/pkg/lib/configreader/errors.go
+++ b/pkg/lib/configreader/errors.go
@@ -34,8 +34,10 @@ const (
 	ErrTooShort                      = "configreader.too_short"
 	ErrAlphaNumericDashUnderscore    = "configreader.alpha_numeric_dash_underscore"
 	ErrAlphaNumericDashDotUnderscore = "configreader.alpha_numeric_dash_dot_underscore"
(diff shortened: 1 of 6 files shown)


pkg/lib/configreader/errors.go
```diff
@@ -126,6 +128,13 @@ func ErrorAlphaNumericDashDotUnderscore(provided string) error {
 	})
 }
 
+func ErrorInvalidAWSTag(provided string) error {
+	return errors.WithStack(&errors.Error{
+		Kind:    ErrInvalidAWSTag,
+		Message: fmt.Sprintf("%s must contain only letters, numbers, spaces, and the following characters: _ . : / = + - @", s.UserStr(provided)),
```
vishalbollu, line 134: Do you think it would be easier to display the regex? Users can copy paste the regex into a regex matcher and test it out more easily.

vishalbollu (approved):

pkg/lib/configreader/errors.go
deliahu: I think this is still clear enough since it's a list of allowed characters; I think it starts to get more confusing when order matters

PR #1360, opened 2020-09-16
Save api spec using json instead of msgpack

### LINKED ISSUE #1295: Serialize API spec using JSON instead of Msgpack
#### Description

API specs are being serialized as msgpack because historically, api configuration needed to support non json compliant structures such as maps with integer keys which. This requirement is no longer necessary. API Specs can be serialized with JSON.

#### Motivation

* Remove serialization logic which marshals api specs to msgpack bytes and then base64 encodes them so to allow structs containing api specs to be marshalled in JSON (e.g. schema.APIResponse)
* The GET job endpoint for a BatchAPI returns a msgpack encoded base64 string of API Spec making it hard to extract the API Spec.

vishalbollu:
closes #1295

---

checklist:

### DIFF
--- a/cli/local/api.go
+++ b/cli/local/api.go
@@ -213,6 +213,12 @@ func FindAPISpec(apiName string) (*spec.API, error) {
 
 	var apiSpec spec.API
 	for _, specPath := range filepaths {
+		if strings.HasSuffix(filepath.Base(specPath), "-spec.msgpack") {
+			apiSpecVersion := GetVersionFromAPISpecFilePath(specPath)
+			if apiSpecVersion != consts.CortexVersion {
+				return nil, ErrorCortexVersionMismatch(apiName, apiSpecVersion)
+			}
+		}
 		if strings.HasSuffix(filepath.Base(specPath), "-spec.json") {
 			apiSpecVersion := GetVersionFromAPISpecFilePath(specPath)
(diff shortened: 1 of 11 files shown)


pkg/lib/configreader/interface_map.go
```diff
@@ -139,6 +139,8 @@ func validateInterfaceMap(val map[string]interface{}, v *InterfaceMapValidation)
 			if err != nil {
 				return nil, errors.Wrap(err, key)
 			}
+
+			val[key] = stringToIntMap
```
deliahu, line 143: Nice! Maybe add a comment on the struct definition that this flag is recursive (even though I presume this was already the case before your fix)

deliahu (approved):

cli/local/api.go
```diff
@@ -213,6 +213,12 @@ func FindAPISpec(apiName string) (*spec.API, error) {
 
 	var apiSpec spec.API
 	for _, specPath := range filepaths {
+		if strings.HasSuffix(filepath.Base(specPath), "-spec.msgpack") {
+			apiSpecVersion := GetVersionFromAPISpecFilePath(specPath)
```
deliahu: Do we need to check the version here, since if it’s msgpack then it’s stale?

PR #1366, opened 2020-09-18
Expose GPUs using device driver

### LINKED ISSUE #1223: Use nvidia-container-toolkit instead of nvidia-docker2 to expose GPUs in Cortex local
#### Description

Cortex local currently relies on setting up a docker runtime with nvidia-docker2 to access gpus. This method is deprecated as of Docker version 19.03. For Docker versions >= 19.03, GPUs should be accessible via `--gpus all` flag after installing nvidia-container-toolkit https://github.com/NVIDIA/nvidia-docker#quickstart.

If it is possible, support both ways of exposing GPUs to Cortex local.

vishalbollu:
closes #1223 

---

checklist:

### DIFF
--- a/cli/local/docker_spec.go
+++ b/cli/local/docker_spec.go
@@ -446,26 +446,25 @@ func deployTensorFlowContainers(api *spec.API, awsClient *aws.Client) error {
 // Retries deploying a container requiring GPU using nvidia runtime, returns original error if isn't relevant, nil if successful and new error if a retry was attempted but failed
 func retryWithNvidiaRuntime(err error, containerConfig *container.Config, hostConfig *container.HostConfig) error {
 	// error message if device driver may look like 'could not select device driver "" with capabilities: [[gpu]]'
-
-	if strings.Contains(err.Error(), "could not select device driver") && strings.Contains(err.Error(), "gpu") {
-		if _, ok := docker.MustDockerClient().Info.Runtimes["nvidia"]; ok {
-			fmt.Println("retrying API deployment using nvidia runtime because device driver for GPU was not found")
-			hostConfig.Runtime = "nvidia"
-			hostConfig.Resources.DeviceRequests = nil
-			containerCreateRequest, err := docker.MustDockerClient().ContainerCreate(context.Background(), containerConfig, hostConfig, nil, "")
-			if err != nil {
-				return errors.Wrap(err, "failed to request a GPU")
-			}
-			err = docker.MustDockerClient().ContainerStart(context.Background(), containerCreateRequest.ID, dockertypes.ContainerStartOptions{})
-			if err != nil {
-				return errors.Wrap(err, "failed to run a container using nvidia runtime; please verify that you've set up your nvidia runtime with Docker correctly")
-			}
-			return nil
-		} else {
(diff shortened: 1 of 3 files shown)


cli/local/docker_spec.go
```diff
@@ -134,7 +134,12 @@ func deployPythonContainer(api *spec.API, awsClient *aws.Client) error {
 			resources.Memory = api.Compute.Mem.Quantity.Value()
 		}
 		if api.Compute.GPU > 0 {
-			runtime = "nvidia"
```
deliahu, line 137: I believe that `runtime := ""` above and `Runtime:   runtime,` below can be deleted

cli/local/docker_spec.go
```diff
@@ -183,7 +188,25 @@ func deployPythonContainer(api *spec.API, awsClient *aws.Client) error {
 
 	err = docker.MustDockerClient().ContainerStart(context.Background(), containerInfo.ID, dockertypes.ContainerStartOptions{})
 	if err != nil {
-		return errors.Wrap(err, api.Identify())
+		if api.Compute.GPU > 0 && strings.Contains(err.Error(), "could not select device driver \"\" with capabilities: [[gpu]]") {
```
deliahu, line 191: I'm a little worried that checking the string so precisely could be brittle, what do you think about shortening the string match requirement?

cli/local/docker_spec.go
```diff
@@ -183,7 +188,25 @@ func deployPythonContainer(api *spec.API, awsClient *aws.Client) error {
...
+			if _, ok := docker.MustDockerClient().Info.Runtimes["nvidia"]; ok {
+				fmt.Println("retrying API deployment using nvidia runtime because device driver for GPU was not found")
+				hostConfig.Runtime = "nvidia"
+				hostConfig.Resources.DeviceRequests = nil
+				containerCreateRequest, err := docker.MustDockerClient().ContainerCreate(context.Background(), containerConfig, hostConfig, nil, "")
+				if err != nil {
+					return errors.Wrap(err, api.Identify(), "failed to allocate GPU")
```
deliahu, line 198: Would it be useful to include the more verbose error message used below for this error and the error on line 202?

cli/local/docker_spec.go
```diff
@@ -205,7 +228,12 @@ func deployONNXContainer(api *spec.API, awsClient *aws.Client) error {
 			resources.Memory = api.Compute.Mem.Quantity.Value()
 		}
 		if api.Compute.GPU > 0 {
-			runtime = "nvidia"
```
deliahu, line 208: Same comment as above regarding removing `runtime`

cli/local/docker_spec.go
```diff
@@ -287,7 +333,12 @@ func deployTensorFlowContainers(api *spec.API, awsClient *aws.Client) error {
 			serveResources.Memory = totalMemory - apiResources.Memory
 		}
 		if api.Compute.GPU > 0 {
-			serveRuntime = "nvidia"
```
deliahu, line 290: Same comment as above regarding removing `serveRuntime `

cli/local/docker_spec.go
```diff
@@ -349,7 +400,25 @@ func deployTensorFlowContainers(api *spec.API, awsClient *aws.Client) error {
 
 	err = docker.MustDockerClient().ContainerStart(context.Background(), containerCreateRequest.ID, dockertypes.ContainerStartOptions{})
 	if err != nil {
-		return errors.Wrap(err, api.Identify())
+		if api.Compute.GPU > 0 && strings.Contains(err.Error(), "could not select device driver \"\" with capabilities: [[gpu]]") {
```
deliahu, line 403: Would it be easy to move this code block to a helper function to be used by each predictor?

deliahu (approved):

PR #1377, opened 2020-09-23
Add a lock around metrics recording for local apis

### LINKED ISSUE #1209: Verify accuracy of request metrics for a local deployment using multiple threads
#### Version

v0.18.*

#### Description

Support for multithreading and multiprocessing was recently added to Cortex local. Request metrics are tracked by increment files to a local directory. Multiprocessing with Cortex local may work as expected because each process writes to files prefixed with their PIDs. However, multithreading on Cortex local may be broken.

Add a mutex to request metrics files to prevent multiple threads from overwriting the same file with outdated values.

vishalbollu:
closes #1209 

---

checklist:

### DIFF
--- a/pkg/workloads/cortex/lib/type/api.py
+++ b/pkg/workloads/cortex/lib/type/api.py
@@ -117,12 +117,11 @@ class API:
             cx_logger().warn("failure encountered while publishing metrics", exc_info=True)
 
     def store_metrics_locally(self, status_code, total_time):
-        self.metrics_file_lock.acquire()
-
         status_code_series = int(status_code / 100)
         status_code_file_name = f"/mnt/workspace/{os.getpid()}.{status_code_series}XX"
         request_time_file = f"/mnt/workspace/{os.getpid()}.request_time"
 
+        self.metrics_file_lock.acquire()
         try:
             self.increment_counter_file(status_code_file_name, 1)
             self.increment_counter_file(request_time_file, total_time)
--- a/pkg/workloads/cortex/lib/type/api.py
+++ b/pkg/workloads/cortex/lib/type/api.py
@@ -17,6 +17,7 @@ import base64
 import time
 from pathlib import Path
 import json
+import threading
 
 import datadog
 
@@ -47,6 +48,8 @@ class API:
             host_ip = os.environ["HOST_IP"]
             datadog.initialize(statsd_host=host_ip, statsd_port="8125")
             self.statsd = datadog.statsd
+        else:
+            self.metrics_file_lock = threading.Lock()
 
     def get_cached_classes(self):
         prefix = os.path.join(self.metadata_root, "classes") + "/"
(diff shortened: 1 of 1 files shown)


pkg/workloads/cortex/lib/type/api.py
```diff
@@ -47,6 +48,8 @@ def __init__(self, provider, storage, model_dir, cache_dir=".", **kwargs):
             host_ip = os.environ["HOST_IP"]
             datadog.initialize(statsd_host=host_ip, statsd_port="8125")
             self.statsd = datadog.statsd
+        else:
```
deliahu, line 51: nit: it might be easier to read as a separate `if provider == "local"`

pkg/workloads/cortex/lib/type/api.py
```diff
@@ -111,13 +114,15 @@ def post_metrics(self, metrics):
...
     def store_metrics_locally(self, status_code, total_time):
+        self.metrics_file_lock.acquire()
         status_code_series = int(status_code / 100)
 
         status_code_file_name = f"/mnt/workspace/{os.getpid()}.{status_code_series}XX"
         self.increment_counter_file(status_code_file_name, 1)
 
         request_time_file = f"/mnt/workspace/{os.getpid()}.request_time"
         self.increment_counter_file(request_time_file, total_time)
+        self.metrics_file_lock.release()
```
deliahu, line 125: I think it would be safer to have `release()` in a `finally`

vishalbollu: Absolutely

deliahu (approved):

pkg/workloads/cortex/lib/type/api.py
```diff
@@ -113,13 +117,17 @@ def post_metrics(self, metrics):
             cx_logger().warn("failure encountered while publishing metrics", exc_info=True)
 
     def store_metrics_locally(self, status_code, total_time):
-        status_code_series = int(status_code / 100)
+        try:
```
RobertLucian, line 120: The `try/except` block isn't required for the following 2 instructions:

```python
self.metrics_file_lock.acquire()
status_code_series = int(status_code / 100)
```

The 2 above will not fail. The only method that might fail though is `self.increment_counter_file`. So we could have something like:

```python
try:
    status_code_file_name = f"/mnt/workspace/{os.getpid()}.{status_code_series}XX"
    self.increment_counter_file(status_code_file_name, 1)
    request_time_file = f"/mnt/workspace/{os.getpid()}.request_time"
    self.increment_counter_file(request_time_file, total_time)
finally:
    self.metrics_file_lock.release()
```

We could also decide to take the 2 string assignments out of the `try` block. I think this could be a little cleaner:

```python
try:
    self.increment_counter_file(status_code_file_name, 1)
    self.increment_counter_file(request_time_file, total_time)
finally:
    self.metrics_file_lock.release()
```

RobertLucian (approved):

PR #1387, opened 2020-09-26
Rollback to previous deployment if update fails [WIP]

### LINKED ISSUE #1250: Perform a complete rollback if one of the update steps fail
#### Description

PR: https://github.com/cortexlabs/cortex/pull/1387/files

There are two steps when updating an API. First the K8s resources are updated, and if that is successful, API gateway is updated. There is a possibility where K8s resources were updated successfully but the API Gateway update failed. If the user were updating the API endpoint in such a scenario, the API would still be considered as live in `cortex get` but it might not be reachable because of the mismatch routes between API Gateway and and the virtual service.

In order to determine whether a deployment of an API is a creation or an update to an existing API, the relevant k8s resources are queried. One solution to recover from partial failure is to do a complete rollback and apply the previous K8s resources and reapply the previous version of API Gateway. 

The error message should contain the status of the rollback.

vishalbollu:
closes #1250 

---

checklist:

### DIFF
--- a/pkg/operator/resources/batchapi/api.go
+++ b/pkg/operator/resources/batchapi/api.go
@@ -77,17 +77,45 @@ func UpdateAPI(apiConfig *userconfig.API, projectID string) (*spec.API, string, 
 
 		err = applyK8sResources(api, prevVirtualService)
 		if err != nil {
-			return nil, "", err
+			rollBackErr := rollBack(prevVirtualService)
+			if rollBackErr != nil {
+				return nil, "", err
+			}
+			return nil, "update failed; rollbacked to previous deployment", nil
 		}
 
 		if err := operator.UpdateAPIGatewayK8s(prevVirtualService, api, true); err != nil {
-			return nil, "", err
+			rollBackErr := rollBack(prevVirtualService)
+			if rollBackErr != nil {
+				return nil, "", err
+			}
+			return nil, "update failed; rollbacked to previous deployment", nil
 		}
 
 		return api, fmt.Sprintf("updated %s", api.Resource.UserString()), nil
 	}
 
 	return api, fmt.Sprintf("%s is up to date", api.Resource.UserString()), nil
+}
+
+func rollBack(prevVirtualService *istioclientnetworking.VirtualService) error {
+	prevAPIID := prevVirtualService.Labels["apiID"]
+	prevAPIName := prevVirtualService.Labels["apiName"]
+
+	prevAPI, err := operator.DownloadAPISpec(prevAPIName, prevAPIID)
+	if err != nil {
+		return err
+	}
+
+	if err := applyK8sResources(prevAPI, prevVirtualService); err != nil {
+		return err
+	}
+
+	if err := operator.UpdateAPIGatewayK8s(prevVirtualService, prevAPI, false); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func DeleteAPI(apiName string, keepCache bool) error {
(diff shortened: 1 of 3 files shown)


pkg/operator/resources/batchapi/api.go
```diff
@@ -77,11 +77,19 @@ func UpdateAPI(apiConfig *userconfig.API, projectID string) (*spec.API, string,
 
 		err = applyK8sResources(api, prevVirtualService)
 		if err != nil {
-			return nil, "", err
+			rollBackErr := rollBack(prevVirtualService)
+			if rollBackErr != nil {
+				return nil, "", err
+			}
+			return nil, "update failed; rollbacked to previous deployment", nil
```
deliahu, line 84: I think it might still be good to return the error, in case the user can do anything about it

pkg/operator/resources/batchapi/api.go
```diff
@@ -77,11 +77,19 @@ func UpdateAPI(apiConfig *userconfig.API, projectID string) (*spec.API, string,
 
 		err = applyK8sResources(api, prevVirtualService)
 		if err != nil {
-			return nil, "", err
+			rollBackErr := rollBack(prevVirtualService)
+			if rollBackErr != nil {
+				return nil, "", err
```
deliahu, line 82: I think it might be good to return both errors, and say something along the lines of “update failed” with the error, and “During rollback, another error occurred” with that error

pkg/operator/resources/realtimeapi/api.go
```diff
@@ -65,7 +65,7 @@ func UpdateAPI(apiConfig *userconfig.API, projectID string, force bool) (*spec.A
 		}
 
 		if err := applyK8sResources(api, prevDeployment, prevService, prevVirtualService); err != nil {
-			go deleteK8sResources(api.Name)
+
```
deliahu, line 68: Should we still delete what may have been created?

pkg/operator/resources/realtimeapi/api.go
```diff
@@ -65,7 +65,7 @@ func UpdateAPI(apiConfig *userconfig.API, projectID string, force bool) (*spec.A
...
 			return nil, "", err
 		}
 		err = operator.AddAPIToAPIGateway(*api.Networking.Endpoint, api.Networking.APIGateway, false)
```
deliahu, line 71: Should we remove from api gateway if this fails, like we do for batch APIs?

pkg/operator/resources/trafficsplitter/api.go
```diff
@@ -59,17 +59,45 @@ func UpdateAPI(apiConfig *userconfig.API, force bool) (*spec.API, string, error)
 			return nil, "", errors.Wrap(err, "upload api spec")
```
deliahu, line 59: (above) Should we remove from API Gateway if adding it failed, like we do in batch?

pkg/operator/resources/batchapi/api.go
```diff
@@ -90,6 +98,26 @@ func UpdateAPI(apiConfig *userconfig.API, projectID string) (*spec.API, string,
...
+	prevAPI, err := operator.DownloadAPISpec(prevAPIName, prevAPIID)
+	if err != nil {
+		return err
+	}
+
+	if err := applyK8sResources(prevAPI, prevVirtualService); err != nil {
+		return err
+	}
+
+	if err := operator.UpdateAPIGatewayK8s(prevVirtualService, prevAPI, false); err != nil {
```
deliahu, line 114: Should the third arg be “true”?

PR #1388, opened 2020-09-26
Enable users to specify CIDR for cluster VPC

vishalbollu:
---

checklist:


pkg/types/clusterconfig/clusterconfig.go
```diff
@@ -1018,6 +1026,19 @@ func validateBucketName(bucket string) (string, error) {
 	return bucket, nil
 }
 
+func validateVPCCIDR(cidr string) (string, error) {
+	if cidr == "" {
```
deliahu, line 1030: Should empty string be allowed? If not, we can remove this check.

deliahu: EDIT: I think empty is probably already blocked by the config reader. So we can probably drop this check

cli/cmd/lib_cluster_config.go
```diff
@@ -515,7 +520,7 @@ func clusterConfigConfirmationStr(clusterConfig clusterconfig.Config, awsCreds A
 	items.Add(clusterconfig.MaxInstancesUserKey, *clusterConfig.MaxInstances)
 	items.Add(clusterconfig.TagsKey, s.ObjFlatNoQuotes(clusterConfig.Tags))
 	if clusterConfig.SSLCertificateARN != nil {
-		items.Add(clusterconfig.SSLCertificateARNKey, *clusterConfig.SSLCertificateARN)
+		items.Add(clusterconfig.SSLCertificateARNUserKey, *clusterConfig.SSLCertificateARN)
```
deliahu, line 523: Good catch. Should we do `TagsUserKey` above too?

deliahu (approved):

PR #1417, opened 2020-10-03
Update image building

deliahu:


Makefile
```diff
@@ -144,6 +144,7 @@ tools:
 	@go get -u -v golang.org/x/lint/golint
 	@go get -u -v github.com/VojtechVitek/rerun/cmd/rerun
 	@python3 -m pip install black
+	@if [[ "$$OSTYPE" == "darwin"* ]]; then brew install parallel; else sudo apt-get install -y parallel; fi
```
RobertLucian, line 147: Shouldn't `$$OSTYPE` actually be `$OSTYPE`?

And 2nd, what do you think of using an `elif` instead of `else` to check if it's a Linux machine? Not necessarily required, but good to have I think.

build/build-image.sh
```diff
@@ -21,51 +21,25 @@ ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/.. >/dev/null && pwd)"
...
-  -f $dir/Dockerfile \
-  -t cortexlabs/${image} \
-  -t cortexlabs/${image}:${CORTEX_VERSION}
-
-if [ "$slim" == "true" ]; then
-  if [ "${image}" == "python-predictor-gpu" ]; then
-    cuda=("10.0" "10.1" "10.2" "11.0")
-    cudnn=("7" "7" "7" "8")
+build_args=""
+if [[ "$image" == *"-slim"* ]]; then
```
RobertLucian, line 32: nit: does it work if the expression is evaluated from `"$image" == *"-slim"`?

dev/registry.sh
```diff
@@ -60,62 +60,31 @@ function ecr_login() {
...
 function build() {
-  local dir=$1
-  local image=$2
-  local tag=$3
+  local image=$1
+  local tag=$2
+  local dir="${ROOT}/images/${image/-slim}"
+
+  build_args=""
+  if [[ "$image" == *"-slim"* ]]; then
```
RobertLucian, line 76: nit: same as above with `"$image" == *"-slim"`.

build/push-images.sh
```diff
@@ -24,10 +24,9 @@ source $ROOT/dev/util.sh
 
 # if parallel utility is installed, the docker push commands will be parallelized
 if command -v parallel &> /dev/null ; then
-  images=$(join_by "," "${ci_images[@]}")
-  ROOT=$ROOT DOCKER_USERNAME=$DOCKER_USERNAME DOCKER_PASSWORD=$DOCKER_PASSWORD SHELL=$(type -p /bin/bash) parallel --halt now,fail=1 --eta -d"," --colsep=" " $ROOT/build/push-image.sh {1} {2} ::: "${images}"
+  ROOT=$ROOT DOCKER_USERNAME=$DOCKER_USERNAME DOCKER_PASSWORD=$DOCKER_PASSWORD SHELL=$(type -p /bin/bash) parallel --will-cite --halt now,fail=1 $ROOT/build/push-image.sh {} ::: "${all_images[@]}"
```
RobertLucian, line 27: nit: no `--eta` tag on this? We do seem to have one for the one above. Thinking this should give us some feedback when the build takes too much time.

dev/registry.sh
```diff
@@ -192,39 +154,34 @@ elif [ "$cmd" = "create" ]; then
...
+  images_to_build+=( "${user_facing_images[@]}" )
+
+  if [ "$flag_include_slim" == "true" ]; then
+    images_to_build+=( "${user_facing_slim_images[@]}" )
+  fi
 
   if command -v parallel &> /dev/null ; then
-    images=$(join_by "," "${images_to_build[@]}")
-    flag_include_slim=$flag_include_slim flag_skip_push=$flag_skip_push ecr_logged_in=$ecr_logged_in ROOT=$ROOT REGISTRY_URL=$REGISTRY_URL SHELL=$(type -p /bin/bash) parallel --halt now,fail=1 --eta -d"," --colsep=" " build_and_push "$ROOT/images/{1} {1} latest {2}" ::: "${images[@]}"
+    flag_skip_push=$flag_skip_push ecr_logged_in=$ecr_logged_in ROOT=$ROOT REGISTRY_URL=$REGISTRY_URL SHELL=$(type -p /bin/bash) parallel --will-cite --halt now,fail=1 build_and_push "{} latest" ::: "${images_to_build[@]}"
```
RobertLucian, line 181: nit: same as the one above regarding the `--eta` tag.

RobertLucian (approved):
I like this more. I think this way it's simpler. I only have a few nits.

dev/registry.sh
deliahu: Yes this was a mistake, thanks!

build/push-images.sh
deliahu: Yes this was a mistake, thanks!

build/build-image.sh
deliahu: I added the `*` at the end since I thought that maybe in the future we could have more suffixes which could come after. But we could also make the change then if that happens, what do you think?

Makefile
deliahu: For some reason, in makefiles `$$` gets converted to just `$`

I added the `elif`

build/build-image.sh
RobertLucian: I see. For strictness purposes, I'd lean towards not having that `*` at the end. I don't exactly have a strong opinion on this, so whichever you think makes more sense.

Makefile
RobertLucian: Got it. That's weird indeed.

PR #1421, opened 2020-10-05
Override the number of parallel jobs for the CI machine

RobertLucian:
This PR reduces the number of parallel jobs for the CI machine from a max of 36 down to 2. This is to prevent getting 137 error codes within Docker (OOM error).

Related to https://github.com/cortexlabs/cortex/pull/1413.

---

checklist:

### DIFF
--- a/build/build-images.sh
+++ b/build/build-images.sh
@@ -24,7 +24,8 @@ source $ROOT/dev/util.sh
 
 # if parallel utility is installed, the docker build commands will be parallelized
 if command -v parallel &> /dev/null ; then
-  ROOT=$ROOT SHELL=$(type -p /bin/bash) parallel --will-cite --halt now,fail=1 --eta $ROOT/build/build-image.sh {} ::: "${all_images[@]}"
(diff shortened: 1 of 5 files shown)


build/build-images.sh
```diff
@@ -24,7 +24,8 @@ source $ROOT/dev/util.sh
 
 # if parallel utility is installed, the docker build commands will be parallelized
 if command -v parallel &> /dev/null ; then
```
deliahu, line 26: I think it might be better to run the `else` even if `NUM_BUILD_PROCS` is defined but it's set to 1; something like this should work:

```bash
if command -v parallel &> /dev/null && [ -n "$NUM_BUILD_PROCS" ] && [ "$NUM_BUILD_PROCS" != "1" ]; then
  ...
else
  ...
fi
```

If we do it this way, we could also use `$NUM_BUILD_PROCS` directly and remove the `num_procs=` line.

what do you think?

RobertLucian: Yes, I agree. I'm making this change.

deliahu (approved):

PR #1444, opened 2020-10-14
Display environment list output in json

vishalbollu:
---

checklist:

### DIFF
--- a/cli/cmd/env.go
+++ b/cli/cmd/env.go
@@ -18,9 +18,12 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/cortexlabs/cortex/cli/types/cliconfig"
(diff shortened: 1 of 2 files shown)


cli/cmd/env.go
```diff
@@ -125,6 +129,15 @@ var _envListCmd = &cobra.Command{
 			exit.Error(err)
 		}
 
+		if _flagOutput == flags.JSONOutputType {
+			bytes, err := libjson.Marshal(cliConfig.Environments)
```
deliahu, line 133: Should we use `MarshalIndent()` here to make it look nicer?

deliahu (approved):

cli/cmd/env.go
vishalbollu: As discussed, for now `-o json` will not pretty print json.

PR #1449, opened 2020-10-15
Python client

### LINKED ISSUE #684: Add Python client
#### Description

Previous version of Python Client docs: https://github.com/cortexlabs/cortex/pull/685

- [ ] update docs
- [ ] create a sample colab notebook

vishalbollu:
closes #684 

---

checklist:

- [ ] update examples


pkg/workloads/cortex/client/setup.py
```diff
@@ -0,0 +1,91 @@
...
+                raise Exception("cortex is only on mac and linux")
+
+            cortex_version = self.config_vars["dist_version"]
+
+            if "dev" in cortex_version:
+                cortex_version = "master"
+
+            download_url = f"https://s3-us-west-2.amazonaws.com/get-cortex/{cortex_version}/cli/{platform}/cortex"
+
+            with open(dest_path, "wb") as f:
```
deliahu, line 47: nit: if this is a clean swap, I'd go for it: https://stackoverflow.com/a/39217788

pkg/workloads/cortex/client/setup.py
```diff
@@ -0,0 +1,91 @@
...
+            f = Path(dest_path)
+            f.chmod(f.stat().st_mode | stat.S_IEXEC)
+
+
+setup(
+    name="cortex",
+    version="0.20.0.dev3",  # CORTEX_VERSION
+    description="Model serving at scale",
+    author="cortex.dev",
+    author_email="dev@cortexlabs.com",
```
deliahu, line 60: this could be `dev@cortex.dev` if you prefer it

pkg/workloads/cortex/client/setup.py
```diff
@@ -0,0 +1,91 @@
...
+                f.write(r.content)
+
+            f = Path(dest_path)
+            f.chmod(f.stat().st_mode | stat.S_IEXEC)
+
+
+setup(
+    name="cortex",
+    version="0.20.0.dev3",  # CORTEX_VERSION
+    description="Model serving at scale",
```
deliahu, line 58: sounds good to me, just run it by Omer

pkg/workloads/cortex/client/setup.py
```diff
@@ -0,0 +1,91 @@
...
+    author="cortex.dev",
+    author_email="dev@cortexlabs.com",
+    license="Apache License 2.0",
+    long_description_content_type="text/markdown",
+    long_description=open("README.md").read(),
```
deliahu, line 63: does this make a file descriptor that needs to be closed?

cli/cmd/deploy.go
```diff
@@ -106,23 +106,31 @@ var _deployCmd = &cobra.Command{
...
-		if _flagOutput == flags.JSONOutputType {
+		switch _flagOutput {
+		case flags.JSONOutputType:
 			bytes, err := libjson.Marshal(deployResults)
 			if err != nil {
 				exit.Error(err)
 			}
 			fmt.Println(string(bytes))
 			return
+		case flags.MixedOutputType:
```
RobertLucian, line 124: Would it be better if we'd call it `base64-encoded-json` output type or something like this? What are your thoughts?

dev/cli_md_template.md
```diff
@@ -2,7 +2,13 @@
...
 _WARNING: you are on the master branch, please refer to the docs on the branch that matches your `cortex version`_
 
-## Installing a specific version of the CLI
+## Install CLI
+
+```bash
+pip install cortex
+```
+
+## Install CLI without Python dependency
```
RobertLucian, line 11: Maybe rename this to `Install CLI without Python Client`.

I think with `Install CLI without Python dependency` we may run the risk of the user thinking of `pip install cortex --no-deps`, and so they'll just jump on using that instead.

pkg/workloads/cortex/client/cortex/binary/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+    args: List[str],
+    hide_output: bool = False,
+    mixed: bool = False,
+) -> str:
+    """
+    Runs the Cortex binary with the specified arguments.
+
+    Args:
+        args: Arguments to use when invoking the Cortex binary.
+        hide_output: Flag to prevent streaming Cortex binary output to stdout. Defaults to False.
```
RobertLucian, line 48: Thinking we could get rid of the default param value comments (i.e. `"Defaults to False"`) as these will show up when hitting `shift+tab` anyway in notebooks and in python scripts.

pkg/workloads/cortex/client/cortex/binary/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+        stdout=subprocess.PIPE,
+        encoding="utf8",
+        env=env,
+    )
+
+    output = ""
+    result = ""
+    result_found = False
+
+    for c in iter(lambda: process.stdout.read(1), ""):
```
RobertLucian, line 72: What do you think of reading all the standard output first and then run a regex on it to extract the base64-encoded data? Would this work?

cli/cmd/deploy.go
vishalbollu: It isn't exactly strictly base64-encoded-json, we can discuss this live.

pkg/workloads/cortex/client/cortex/binary/__init__.py
vishalbollu: We need to stream part of the such as the download image progress output live. We can discuss this one live as well.

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -0,0 +1,240 @@
...
+            api_name,
+            "--env",
+            self.env,
+            "--force",
+            "-o",
+            "json",
+        ]
+
+        if keep_cache:
+            args += "--keep-cache"
```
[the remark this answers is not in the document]
vishalbollu: Good catch

pkg/workloads/cortex/client/setup.py
```diff
@@ -0,0 +1,107 @@
...
+
+        dest_dir = os.path.join(self.install_lib, "cortex", "binary")
+
+        zip_file_path = os.path.join(dest_dir, "cli.zip")
+        cli_file_path = os.path.join(dest_dir, "cli")
+
+        if not os.path.exists(cli_file_path):
+            platform = sys.platform
+
+            if not platform.startswith("darwin") and not platform.startswith("linux"):
```
deliahu, line 41: if it starts with "linux", should we set it to "linux"? otherwise could line 49 fail? (same with darwin)

cli/local/config.go
```diff
@@ -72,4 +73,6 @@ func init() {
 		err := errors.Wrap(err, "unable to write to home directory", _modelCacheDir)
 		exit.Error(err)
 	}
+
+	OutputType = flags.PrettyOutputType
```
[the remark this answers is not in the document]
vishalbollu: No it isn't

pkg/workloads/cortex/client/setup.py
vishalbollu: Yup, good catch.

pkg/workloads/cortex/client/cortex/binary/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+MIXED_CORTEX_MARKER = "~~cortex~~"
+
+
+def run():
+    """
+    Runs the CLI from terminal.
+    """
+    try:
+        process = subprocess.run([get_cli_path()] + sys.argv[1:], cwd=os.getcwd())
+    except KeyboardInterrupt:
```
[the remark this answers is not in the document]
vishalbollu: This error was caught here to prevent throwing a python exception when user ctrl-c's during `cortex logs`. I think it is okay if throw a KeyboardInterrupt exception when executing the python code (likely when someone is running cortex in a Python REPL like a notebook).

deliahu (approved):

pkg/workloads/cortex/client/setup.py
```diff
@@ -0,0 +1,116 @@
...
+    setup_requires=(["setuptools", "requests", "wheel"]),
+    packages=find_packages(),
+    package_data={"cortex.binary": ["cli"]},
+    entry_points={
+        "console_scripts": [
+            "cortex = cortex.binary:run",
+        ],
+    },
+    install_requires=(["importlib-resources; python_version < '3.7'"]),
+    python_requires=">=3.6.1",
```
RobertLucian, line 98: nit: maybe this should be 3.6.0 or 3.6. Is there a reason why 3.6.0 isn't supported?

RobertLucian (approved):
LGTM!

PR #1460, opened 2020-10-18
Add support for private docker image registries

### LINKED ISSUE #1113: Add support for private image registries
#### Description

At the moment, users can only use public images hosted on DockerHub or images hosted on AWS ECR. This feature would add support for private registries as well (i.e. DockerHub), accessible through the use of access tokens.

#### Motivation

For users who want their images to be private and yet not be necessarily hosted on AWS ECR.

deliahu:
closes #1113

---

checklist:


docs/guides/private-docker.md
```diff
@@ -0,0 +1,51 @@
+# Private docker registry
+
+_WARNING: you are on the master branch, please refer to the docs on the branch that matches your `cortex version`_
+
+Until [#1459](https://github.com/cortexlabs/cortex/issues/1459) is addressed, you can use a private docker registry for your Predictor images by following this guide.
```
vishalbollu, line 5: We should mention that this guide is only necessary using images from a private dockerhub registry when deploying to the cluster. Or maybe have a section for local and cluster here.

pkg/lib/docker/docker.go
```diff
@@ -47,13 +48,22 @@ var NoAuth string
...
+}
+
 type Client struct {
 	*dockerclient.Client
 	Info dockertypes.Info
 }
 
-func init() {
-	NoAuth, _ = EncodeAuthConfig(dockertypes.AuthConfig{})
+type K8sAuthSecret struct {
```
vishalbollu, line 60: Could this struct definition be inlined in getDockerAuthStrFromK8s? It looks it is only used for unmarshalling right?

pkg/lib/docker/errors.go
```diff
@@ -56,10 +59,29 @@ func ErrorDockerPermissions(err error) error {
...
+		Message: fmt.Sprintf("%s does not exist locally; download it with `docker pull %s` (if your registry is private, run `docker login` first)", image, image),
+	})
+}
+
+func ErrorImageInaccessible(image string, providerType types.ProviderType, cause error) error {
 	message := fmt.Sprintf("%s is not accessible", image)
+
 	if cause != nil {
-		message += "\n" + errors.Message(cause) // add \n because docker client errors are
+		message += "\n" + errors.Message(cause)
```
vishalbollu, line 73: We can bring back the comment, and also complete the comment? I assume we were going to say that add \n because docker client errors are **useful**'

pkg/lib/k8s/secret.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	if secret == nil {
+		return nil, nil
+	}
+	return secret.Data, nil
+}
+
+func (c *Client) DeleteSecret(name string) (bool, error) {
+	err := c.secretClient.Delete(context.Background(), name, _deleteOpts)
+	if kerrors.IsNotFound(err) {
```
vishalbollu, line 108: I imagine this is handled correctly but shouldn't we check if err is not nill before verifying the error kind? I've seen this implementation in other k8s wrappers so I assume `metav1.StatusReasonUnknown` is returned if it is nil but I just wanted to double-check.

pkg/types/spec/validations.go
```diff
@@ -1282,17 +1303,78 @@ func validateDockerImagePath(image string, providerType types.ProviderType, awsC
...
+		return docker.NoAuth, nil
+	}
+
+	authData, ok := secretData[".dockerconfigjson"]
+	if !ok {
+		return "", ErrorUnexpectedDockerSecretData("should contain \".dockerconfigjson\" key", secretData)
+	}
+
+	var authSecret docker.K8sAuthSecret
+	err = json.Unmarshal(authData, &authSecret)
```
vishalbollu, line 1339: Was the default json package used intentionally here as opposed to libjson?

pkg/lib/docker/errors.go
deliahu: haha my guess was that it was supposed to be "verbose" (since it was referring to the "\n"), but I wasn't sure, so I deleted it :) Should I add "verbose", or something else?

pkg/lib/k8s/secret.go
deliahu: Yes nil is handled, but we can also move this within the `if err != nil {` check. Let's discuss live

pkg/types/spec/validations.go
deliahu: it was not intentional, thanks!

docs/guides/private-docker.md
deliahu: Good call! I went with a note, but could also make a section if you think it's better. We can discuss live

pkg/lib/k8s/secret.go
deliahu: Will resolve separately

vishalbollu (approved):

PR #1468, opened 2020-10-19
Ensure log group is created when multiple prefixes exist for the BatchAPI kind

RobertLucian:
When a batch API is created, if the log group name is a prefix of another existing log group (that's longer in size), then while it's deploying, the log group name for our API won't get created.

---

checklist:

### DIFF
--- a/pkg/operator/resources/batchapi/logging.go
+++ b/pkg/operator/resources/batchapi/logging.go
@@ -44,23 +44,8 @@ func operatorLogStream(jobKey spec.JobKey) string {
 // Checks if log group exists before creating it
 func ensureLogGroupForAPI(apiName string) error {
 	apiNameLogGroup := logGroupNameForAPI(apiName)
-	logGroupExists := false
 
-	err := config.AWS.CloudWatchLogs().DescribeLogGroupsPages(
-		&cloudwatchlogs.DescribeLogGroupsInput{
-			LogGroupNamePrefix: aws.String(apiNameLogGroup),
-		},
-		func(page *cloudwatchlogs.DescribeLogGroupsOutput, lastPage bool) bool {
-			for _, output := range page.LogGroups {
-				if *output.LogGroupName == apiNameLogGroup {
-					logGroupExists = true
-					return false
-				}
-			}
-			return true
-		},
-	)
-
+	logGroupExists, err := config.AWS.DoesLogGroupExist(apiNameLogGroup)
 	if err != nil {
 		return errors.WithStack(err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/operator/resources/batchapi/logging.go
```diff
@@ -43,17 +43,31 @@ func operatorLogStream(jobKey spec.JobKey) string {
...
 // Checks if log group exists before creating it
 func ensureLogGroupForAPI(apiName string) error {
-	output, err := config.AWS.CloudWatchLogs().DescribeLogGroups(&cloudwatchlogs.DescribeLogGroupsInput{
-		Limit:              aws.Int64(1),
-		LogGroupNamePrefix: aws.String(logGroupNameForAPI(apiName)),
-	})
+	apiNameLogGroup := logGroupNameForAPI(apiName)
+	logGroupExists := false
+
+	err := config.AWS.CloudWatchLogs().DescribeLogGroupsPages(
```
vishalbollu, line 49: I know this is the implementation we had discussed during the call a while back. I think we can do an exact match. There is a function called `DoesLogGroupExist` in our `lib/aws` package which can be used to determine if a log group exists or not. That function can be used instead of doing a prefix search.

My apologies!

vishalbollu (changes requested):

pkg/operator/resources/batchapi/logging.go
RobertLucian: No worries! :) With `DoesLogGroupExist` in, the `ensureLogGroupForAPI` function is now simpler.

pkg/operator/resources/batchapi/logging.go
```diff
@@ -43,17 +43,16 @@ func operatorLogStream(jobKey spec.JobKey) string {
...
 func ensureLogGroupForAPI(apiName string) error {
-	output, err := config.AWS.CloudWatchLogs().DescribeLogGroups(&cloudwatchlogs.DescribeLogGroupsInput{
-		Limit:              aws.Int64(1),
-		LogGroupNamePrefix: aws.String(logGroupNameForAPI(apiName)),
-	})
+	apiNameLogGroup := logGroupNameForAPI(apiName)
+
+	logGroupExists, err := config.AWS.DoesLogGroupExist(apiNameLogGroup)
 	if err != nil {
 		return errors.WithStack(err)
```
vishalbollu, line 50: `WithStack` is not necessary here because any unwrapped error would have been wrapped in `DoesLogGroupExist`.

vishalbollu (approved):

pkg/operator/resources/batchapi/logging.go
RobertLucian: Good catch. Thanks!

PR #1492, opened 2020-10-27
Add cluster name/region flags to the CLI

### LINKED ISSUE #1363: Add --name and --region flags to cluster info/down/export commands
#### Description

Allow the user to run e.g. `cortex cluster down --name=cortex --region=us-west-2` to delete a cluster.

Search for all error messages that mention "cortex cluster down", and add these flags to the error messages if possible.

#### Motivation

The user won't need access to the cluster configuration file, and can spin down clusters programmatically.

RobertLucian:
Closes #1363.

---

checklist:


docs/miscellaneous/cli.md
```diff
@@ -170,35 +172,39 @@ Flags:
   -h, --help                        help for configure
 ```
 
-### cluster export
```
deliahu, line 173: We've used `dev/generate_cli_md.sh` to generate this file, and it seems that the order of these two commands is different there. So if you think this order is better, you can update the order in the generator (and might as well rerun it to confirm that there is no diff)

deliahu (approved):

docs/miscellaneous/cli.md
RobertLucian: The current order that we get in `cortex cluster` is `up, info, configure, down, export`. The generator apparently wasn't in the correct order. Modified the order and re-ran to make sure there is no diff.

PR #1521, opened 2020-11-02
Allow overriding of CLI install path and config dir

### LINKED ISSUE #1222: Override the default config directory of the Cortex CLI
#### Description

Currently, all Cortex CLI configs are stored in `~/.cortex` directory. Make this overridable with an environment variable such as `CORTEX_CLI_CONFIG_DIR`.

#### Motivation

Allows the user to set their own config directory, which is useful in the context of having multiple versions of the Cortex CLI. For instance, one could have multiple conda environments set on their development machine that have different versions of the CLI installed, and thus, require different config directories for each environment.

To automatically export the `CORTEX_CLI_CONFIG_DIR` on `conda activate <env>` (_with the assumption that the Cortex CLI is already installed in that environment's path_), the following technique can be employed:
https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables

#### Extra

Might have to modify `get-cli.sh` to accept a new argument that would specify the install path of the CLI executable.

deliahu:
closes #1222

---

checklist:


docs/miscellaneous/cli.md
```diff
@@ -20,6 +20,10 @@ $ bash -c "$(curl -sS https://raw.githubusercontent.com/cortexlabs/cortex/vINSER
 $ bash -c "$(curl -sS https://raw.githubusercontent.com/cortexlabs/cortex/v0.18.1/get-cli.sh)"
 ```
 
+By default, the Cortex CLI is installed at `/usr/local/bin/cortex`. To install the executable elsewhere, export the `CORTEX_INSTALL_PATH` environment variable to your desired location before running the command above.
```
vishalbollu, line 23: `generate_cli_md.sh` needs to be updated accordingly.

vishalbollu (changes requested):

vishalbollu (approved):

PR #1535, opened 2020-11-05
Update cli docs generation script

deliahu:


vishalbollu (changes requested):
Converting the markdown template into multiple lines of echos makes it hard to update the file because now you have to generate the markdown every time to double check that you've updated the echos correctly. What is the motivation for this change?

vishalbollu (approved):

PR #1536, opened 2020-11-05
Disallow ARM-based instances, which are not currently supported

deliahu:
checklist:


vishalbollu (approved):

pkg/lib/aws/ec2.go
```diff
@@ -28,6 +29,13 @@ import (
 	s "github.com/cortexlabs/cortex/pkg/lib/strings"
 )
 
+var _armInstanceRegex = regexp.MustCompile(`^\w+[0-9]+\w*g\w*\.\w+$`)
```
vishalbollu, line 32: A quick comment describing what this regex is trying to match would help here

PR #1537, opened 2020-11-06
Check instance quotas during cluster creation

### LINKED ISSUE #1461: Check that vCPU quota is sufficient for max_instances on cluster up/configure
#### Description

When running `cortex cluster up` or `cortex cluster configure`, the CLI should check that the user's AWS account has enough vCPU quota to support `max_instances` instances of the chosen instance type.

#### Question

What would be the strategy for spot instances (or would this check only be applied when not using spot instances)? With spot, there could be multiple instance types, and the spot quota is a different quota than for on-demand (see [here](https://docs.cortex.dev/cluster-management/spot-instances)).

deliahu:
closes #1461

---

checklist:


pkg/lib/aws/errors.go
```diff
@@ -128,10 +128,11 @@ func ErrorBucketNotFound(bucket string) error {
...
 	return errors.WithStack(&errors.Error{
-		Kind:    ErrInstanceTypeLimitIsZero,
-		Message: fmt.Sprintf(`you don't have access to %s instances in %s; please request access in the appropriate region (https://console.aws.amazon.com/support/cases#/create?issueType=service-limit-increase&limitType=ec2-instances). If you submitted a request and it was recently approved, please allow ~30 minutes for AWS to reflect this change."`, instanceType, region),
+		Kind:    ErrInsufficientInstanceQuota,
+		Message: fmt.Sprintf("your cluster may require up to %d %s %s instances, but your AWS quota for %s %s instances in %s is only %d vCPU (there are %d vCPUs per %s instance); please reduce the maximum number of %s %s instances your cluster may use (e.g. by changing max_instances and/or spot_config if applicable), or request a quota increase to at least %d vCPU here: %s (if your request was recently approved, please allow ~30 minutes for AWS to reflect this change)", requiredInstances, lifecycle, instanceType, lifecycle, instanceType, region, vCPUQuota, vCPUPerInstance, instanceType, lifecycle, instanceType, requiredInstances*vCPUPerInstance, url),
```
deliahu, line 135: note to reviewer: this reads like this:

error: dev/config/cluster.yaml: instance_type: your cluster may require up to 99999 spot t3a.small instances, but your AWS quota for spot t3a.small instances in us-west-2 is only 27000 vCPU (there are 2 vCPUs per t3a.small instance); please reduce the maximum number of spot t3a.small instances your cluster may use (e.g. by changing max_instances and/or spot_config if applicable), or request a quota increase to at least 199998 vCPU here: https://us-west-2.console.aws.amazon.com/servicequotas/home?region=us-west-2#!/services/ec2/quotas/L-34B43A08 (if your request was recently approved, please allow ~30 minutes for AWS to reflect this change)

pkg/types/clusterconfig/clusterconfig.go
```diff
@@ -1078,6 +1079,58 @@ func DefaultAccessConfig() (*AccessConfig, error) {
...
+	var count int64
+
+	// default OnDemandBaseCapacity is 0
+	if cc.SpotConfig.OnDemandBaseCapacity != nil {
+		count += *cc.SpotConfig.OnDemandBaseCapacity
+	}
+
+	// default OnDemandPercentageAboveBaseCapacity is 0
+	if cc.SpotConfig.OnDemandPercentageAboveBaseCapacity != nil {
+		count += int64(math.Ceil(float64(*cc.SpotConfig.OnDemandPercentageAboveBaseCapacity) / 100 * float64(*cc.MaxInstances-count)))
```
vishalbollu, line 1100: Just to prevent bugs, I would use `cc.SpotConfig.OnDemandBaseCapacity` directly instead of using count as a variable that temporarily stores `cc.SpotConfig.OnDemandBaseCapacity` `float64(*cc.MaxInstances-count)`. Maybe in L1095 define a separate variable specifically for `cc.SpotConfig.OnDemandBaseCapacity` and use that variable here.

pkg/types/clusterconfig/clusterconfig.go
```diff
@@ -1078,6 +1079,58 @@ func DefaultAccessConfig() (*AccessConfig, error) {
...
+	}
+
+	// default OnDemandBaseCapacity is 0
+	if cc.SpotConfig.OnDemandBaseCapacity != nil {
+		count -= *cc.SpotConfig.OnDemandBaseCapacity
+	}
+
+	// default OnDemandPercentageAboveBaseCapacity is 0
+	if cc.SpotConfig.OnDemandPercentageAboveBaseCapacity != nil {
+		count -= int64(math.Floor(float64(*cc.SpotConfig.OnDemandPercentageAboveBaseCapacity) / 100 * float64(count)))
```
vishalbollu, line 1128: Same as above.

vishalbollu (approved):
LGTM with a few nits.

vishalbollu (approved):

PR #1557, opened 2020-11-13
Update custom docker image docs

deliahu:


docs/deployments/system-packages.md
```diff
@@ -111,7 +121,7 @@ Update your API configuration file to point to your image:
 - name: my-api
   ...
   predictor:
-    image: <repository_url>:latest
+    image: <your_image_path>:latest
```
RobertLucian, line 124: Should the `image` field still be set to `<repository_url>` as opposed to `<your_image_path>`? We are assuming the use of a custom-built image that's pushed to an image repository of the user's choice.

deliahu: Yeah that sounds good. I changed it since if you are just building and running locally then you won't have a repository url, but I think if you're running locally you'd be able to figure it out based on what I added above

RobertLucian: Got it. I see what you meant.

RobertLucian (approved):

PR #1587, opened 2020-11-23
Accept class definition from Python Client directly

vishalbollu:
---

checklist:

- [ ] update examples


pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -33,7 +38,126 @@ def __init__(self, env: str):
...
+            cls: A Cortex Predictor class implementation. Not required when deploying a traffic splitter.
+                → Realtime API: https://docs.cortex.dev/v/master/deployments/realtime-api/predictors
+                → Batch API: https://docs.cortex.dev/v/master/deployments/batch-api/predictors
+            force: Override any in-progress api updates.
+            wait: Streams logs until the APIs are ready.
+
+        Returns:
+            Deployment status, API specification, and endpoint for each API.
+        """
+        project_dir = Path.home() / ".cortex" / "deployments" / str(int(time.time() * 1000))
```
deliahu, line 121: nit: use uuid here

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -33,7 +38,126 @@ def __init__(self, env: str):
...
+        with util.open_tempdir(str(project_dir)):
+            if cls is None:
```
deliahu, line 123: nit: add comment here

pkg/workloads/cortex/client/setup.py
```diff
@@ -94,7 +94,7 @@ def run(self):
             "cortex = cortex.binary:run",
         ],
     },
-    install_requires=(["importlib-resources; python_version < '3.7'"]),
+    install_requires=(["importlib-resources; python_version < '3.7'", "pyyaml", "dill"]),
```
deliahu, line 97: lock versions (or maybe just dill)

pkg/workloads/cortex/lib/api/predictor.py
```diff
@@ -254,6 +254,8 @@ def class_impl(self, project_dir):
             refresh_logger()
 
         try:
+            if inspect.isclass(impl):  # for pickled classes
+                return impl
             classes = inspect.getmembers(impl, inspect.isclass)
             predictor_class = None
             for class_df in classes:
```
deliahu, line 261: bring back validator for pickled classes

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -33,7 +38,133 @@ def __init__(self, env: str):
...
+            cls: A Cortex Predictor class implementation. Not required when deploying a traffic splitter.
+                → Realtime API: https://docs.cortex.dev/v/master/deployments/realtime-api/predictors
+                → Batch API: https://docs.cortex.dev/v/master/deployments/batch-api/predictors
+            force: Override any in-progress api updates.
+            wait: Streams logs until the APIs are ready.
+
+        Returns:
+            Deployment status, API specification, and endpoint for each API.
+        """
+        project_dir = Path.home() / ".cortex" / "deployments" / str(uuid.uuid4())
```
RobertLucian, line 69: What does the `Path.home() / ".cortex" / "deployments"` do? Does it append the inlined strings on one another? Does the `/` act as an actual `+ "/" +` line of code?

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -33,7 +38,133 @@ def __init__(self, env: str):
...
+
+            # Change if PYTHONVERSION changes
+            expected_version = "3.6"
+            actual_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+            if actual_version < expected_version:
+                raise Exception("cortex is only supported for python versions >= 3.6")  # unexpected
+            if actual_version > expected_version:
+                is_python_set = False
+                for conda_dep in conda_dependencies:
+                    if "::python=" in conda_dep:
```
RobertLucian, line 83: Are we certain that python packages will always have the colons `::` in their package's name? What would happen if they weren't there? If my memory serves me well, you can also specify `python=3.6.x` in `conda-packages.txt` and that will still work, but it will install a different kind of Python (that we don't want) - I might be wrong though.

In such a case, there will be some sort of conflict when conda runs.

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -33,7 +38,133 @@ def __init__(self, env: str):
...
+                return self.deploy_project(config, project_dir, wait=wait, force=force)
+
+            # Change if PYTHONVERSION changes
+            expected_version = "3.6"
+            actual_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+            if actual_version < expected_version:
+                raise Exception("cortex is only supported for python versions >= 3.6")  # unexpected
+            if actual_version > expected_version:
+                is_python_set = False
+                for conda_dep in conda_dependencies:
```
RobertLucian, line 82: I think we can simplify this with
```python
is_python_set = any("::python=" in conda_dep for conda_dep in conda_dependencies)
```

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -33,7 +38,134 @@ def __init__(self, env: str):
...
+
+            if type(config) is not dict:
+                raise ValueError(
+                    "`config` parameter must be a dictionary defining a single Cortex API"
+                )
+
+            if not os.path.exists(project_dir):
+                raise ValueError(f"{project_dir} is not a valid directory")
+
+            cortex_yaml_path = os.path.join(project_dir, f".cortex-{uuid.uuid4()}.yaml")
```
RobertLucian, line 158: If this `deploy_project` is run twice in a row without modifying the API spec and by only specifying the `config` parameter, will this lead to the creation of 2 `yaml` configs? If there are indeed 2 configs created, should we only generate one instead?

Maybe generating a hash based on the input `config` parameter is what we need. What is your opinion?

pkg/workloads/cortex/client/setup.py
```diff
@@ -94,7 +94,7 @@ def run(self):
             "cortex = cortex.binary:run",
         ],
     },
-    install_requires=(["importlib-resources; python_version < '3.7'"]),
+    install_requires=(["importlib-resources; python_version < '3.7'", "pyyaml", "dill==0.3.2"]),
```
RobertLucian, line 97: Is it safe if we don't specify a version for `pyyaml`?

RobertLucian (commented):
A few comments here. Let me know if you want to discuss.

docs/cluster-management/install.md
```diff
@@ -56,7 +56,7 @@ import requests
 local_client = cortex.client("local")
 
 # deploy the model as a realtime api and wait for it to become active
-deployments = local_client.deploy("./cortex.yaml", wait=True)
+deployments = local_client.deploy_project(config_file="./cortex.yaml", wait=True)
```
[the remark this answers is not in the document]
vishalbollu: Agreed, Omer thinks so as well.

RobertLucian (approved):

PR #1589, opened 2020-11-23
Allow local deployments of public models without requiring AWS creds

### LINKED ISSUE #1588: Deployment of local API with public models is not possible without AWS creds
#### Version

Since 0.22.

#### Description

When a local API is deployed that's using the live-reloading/multi-model caching feature and the specified model(s) is/are public, the API will stop abruptly if no AWS credentials have been configured for it (either via the local environment or by including the credentials in the `.env` file).

#### Suggested solution

If a model is public and there are no AWS credentials set, then the S3 client should be anonymous.

RobertLucian:
Fixes #1588.

---

checklist:

### DIFF
--- a/pkg/workloads/cortex/lib/storage/s3.py
+++ b/pkg/workloads/cortex/lib/storage/s3.py
@@ -27,14 +27,16 @@ from cortex.lib.exceptions import CortexException
 
 
 class S3(object):
-    def __init__(self, bucket=None, region=None, client_config={}):
+    def __init__(self, bucket=None, region=None, client_config={}, anonymous=False):
         self.bucket = bucket
         self.region = region
 
         if client_config is None:
             client_config = {}
 
-        if region is not None:
+        if anonymous:
+            client_config["config"] = botocore.client.Config(signature_version=botocore.UNSIGNED)
+        elif region is not None:
             client_config["region_name"] = region
 
         self.s3 = boto3.client("s3", **client_config)
--- a/pkg/workloads/cortex/lib/storage/s3.py
+++ b/pkg/workloads/cortex/lib/storage/s3.py
@@ -27,17 +27,18 @@ from cortex.lib.exceptions import CortexException
 
 
 class S3(object):
-    def __init__(self, bucket=None, region=None, client_config={}, anonymous=False):
+    def __init__(self, bucket=None, region=None, client_config={}):
         self.bucket = bucket
         self.region = region
 
         if client_config is None:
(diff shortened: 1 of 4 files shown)


pkg/workloads/cortex/lib/storage/s3.py
```diff
@@ -27,14 +27,16 @@
...
         self.bucket = bucket
         self.region = region
 
         if client_config is None:
             client_config = {}
 
-        if region is not None:
+        if anonymous:
+            client_config["config"] = botocore.client.Config(signature_version=botocore.UNSIGNED)
+        elif region is not None:
```
deliahu, line 39: can you have anonymous=True and region specified at the same time?

pkg/workloads/cortex/lib/util.py
```diff
@@ -297,3 +297,7 @@ def render_jinja_template(jinja_template_file: str, context: dict) -> str:
 
     template = env.get_template(str(template_path.name))
     return template.render(**context)
+
+
+def available_aws_credentials() -> bool:
```
deliahu, line 302: nit: consider renaming to `are_aws_credentials_available()`

pkg/workloads/cortex/lib/util.py
```diff
@@ -297,3 +297,7 @@ def render_jinja_template(jinja_template_file: str, context: dict) -> str:
...
+    return os.getenv("AWS_ACCESS_KEY_ID") and os.getenv("AWS_SECRET_ACCESS_KEY")
```
deliahu, line 303: Would it be risky to check this in the S3 constructor instead of checking it outside? For example, to leave the calls like `s3_client = S3(bucket_name)`, but in the constructor, set the unsigned config if the creds aren't available? Or could something break if we do it that way?

pkg/workloads/cortex/lib/storage/s3.py
RobertLucian: Yes, apparently, that works okay.

pkg/workloads/cortex/lib/util.py
RobertLucian: Removed function.

pkg/workloads/cortex/lib/util.py
RobertLucian: Moved the process of checking the availability of AWS credentials inside the S3 class' constructor.

deliahu (approved):

PR #1639, opened 2020-11-27
Terminate container when bootloader script fails

vishalbollu:
---

checklist:

- [ ] update examples

### DIFF
--- a/pkg/operator/operator/k8s.go
+++ b/pkg/operator/operator/k8s.go
@@ -388,6 +388,10 @@ func getEnvVars(api *spec.API, container string) []kcore.EnvVar {
 				Name:  "CORTEX_PROJECT_DIR",
 				Value: path.Join(_emptyDirMountPath, "project"),
 			},
+			kcore.EnvVar{
+				Name:  "S6_BEHAVIOUR_IF_STAGE2_FAILS",
+				Value: "2",
+			},
 		)
 
 		if api.Kind == userconfig.RealtimeAPIKind {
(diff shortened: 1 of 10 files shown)


pkg/operator/operator/k8s.go
```diff
@@ -388,6 +388,10 @@ func getEnvVars(api *spec.API, container string) []kcore.EnvVar {
 				Name:  "CORTEX_PROJECT_DIR",
 				Value: path.Join(_emptyDirMountPath, "project"),
 			},
+			kcore.EnvVar{
```
deliahu, line 391: Should we consider baking it into the container image?

pkg/operator/operator/k8s.go
```diff
@@ -388,6 +388,10 @@ func getEnvVars(api *spec.API, container string) []kcore.EnvVar {
...
+				Name:  "S6_BEHAVIOUR_IF_STAGE2_FAILS",
```
deliahu, line 392: With this environment variable, is it possible to remove any `if [ $status -ne 0 ]` checks in `bootloader.sh`?

RobertLucian, line 392: Tying into what David has said, I think we can now remove all `if [ $status -ne 0 ]` checkers in the bootloader. Those ifs were only in because the captured signals did not kill the pod. We should test this.

pkg/operator/operator/k8s.go
RobertLucian: I agree with David. I think this env can be baked into the container image.

RobertLucian (commented):
Cool catch!

RobertLucian (approved):

deliahu (approved):

PR #1651, opened 2020-12-03
Cortex Patch

vishalbollu:
---

checklist:

- [ ] update examples


pkg/operator/resources/resources.go
```diff
@@ -170,6 +169,83 @@ func UpdateAPI(apiConfig *userconfig.API, models []spec.CuratedModelResource, pr
...
+	models := []spec.CuratedModelResource{}
+	err = ValidateClusterAPIs([]userconfig.API{*apiConfig}, &models, projectFiles)
+	if err != nil {
+		err = errors.Append(err, fmt.Sprintf("\n\napi configuration schema can be found here:\n  → Realtime API: https://docs.cortex.dev/v/%s/deployments/realtime-api/api-configuration\n  → Batch API: https://docs.cortex.dev/v/%s/deployments/batch-api/api-configuration\n  → Traffic Splitter: https://docs.cortex.dev/v/%s/deployments/realtime-api/traffic-splitter", consts.CortexVersionMinor, consts.CortexVersionMinor, consts.CortexVersionMinor))
+		return nil, "", err
+	}
+
+	switch deployedResource.Kind {
+	case userconfig.RealtimeAPIKind:
+		return realtimeapi.UpdateAPI(apiConfig, models, prevAPISpec.ProjectID, force)
```
deliahu, line 243: I'm pretty sure that it's the case, but just wanted to confirm that project ID does not include the API config file

pkg/lib/cast/interface_test.go
```diff
@@ -186,6 +186,37 @@ func TestInterfaceToInterfaceInterfaceMap(t *testing.T) {
...
+	var casted map[string]interface{}
+	var expected map[string]interface{}
+
+	in = map[string]interface{}{"test": map[string]interface{}{"testing": []string{}}}
+	expected = map[string]interface{}{"test": map[string]interface{}{"testing": []string{}}}
+	casted, ok = InterfaceToStrInterfaceMapRecursive(in)
+	require.True(t, ok)
+	require.Equal(t, expected, casted)
+
+	in = map[string]interface{}{"test": map[string]interface{}{"testing": []string{}}, "slice": []int{1}}
```
deliahu, line 201: Should the inner map be type `map[interface{}]interface{}`?

pkg/lib/cast/interface.go
```diff
@@ -694,6 +694,37 @@ func InterfaceToStrInterfaceMap(in interface{}) (map[string]interface{}, bool) {
...
+
+	out := map[string]interface{}{}
+
+	for key, value := range inMap {
+		casted, ok := key.(string)
+		if !ok {
+			return nil, false
+		}
+
+		if intToIntMap, ok := value.(map[interface{}]interface{}); ok {
```
deliahu, line 715: will this work when the nested map is a map[string]interface{}? If not, you might consider using `InterfaceToInterfaceInterfaceMap()`

deliahu: update: I tried it in go playground ([code](https://play.golang.org/p/7-cYZDUxFHK)), and the cast seems to not work so I think we might want `InterfaceToInterfaceInterfaceMap()` here

deliahu: update2: I see you have a test for this, so I'm a bit confused as to why it passes. We can discuss live

cli/local/deploy.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/cortexlabs/cortex/pkg/types/userconfig"
 )
 
-func Deploy(env cliconfig.Environment, configPath string, projectFileList []string, deployDisallowPrompt bool) ([]schema.DeployResult, error) {
+func Deploy(env cliconfig.Environment, configPath string, projectFileList []string, _disallowPrompt bool) ([]schema.DeployResult, error) {
```
deliahu, line 35: nit: it should not start with an underscore since it's not a global variable

deliahu (approved):

pkg/workloads/cortex/client/cortex/client.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, env: dict):
         """
         self.env = env
         self.env_name = env["name"]
+        print(self.env)
```
deliahu, line 46: This can be removed

deliahu (approved):

PR #1655, opened 2020-12-04
Add GCP support to Cortex

### LINKED ISSUE #114: Support GCP
### Notes

* Update some docs (e.g. how to install GPUs)

### LINKED ISSUE #1600: GCP - Add cluster-gcp commands and support GCP CLI environment
* Add relevant `cluster-gcp` commands
  * Add validation that if using cluster-gcp commands with cluster config, provider must be gcp (same for aws commands)
* `env configure` (just operator URL)

### LINKED ISSUE #1602: GCP - Support GCS models (e.g. ONNX and TF)

### LINKED ISSUE #1616: GCP - Support GPU

### LINKED ISSUE #1624: GCP - Support live reloading and multi-model caching

RobertLucian:
Closes #114, closes #1600, closes #1602, closes #1616, closes #1624.

---

checklist:


pkg/workloads/cortex/serve/init/bootloader.sh
```diff
@@ -141,4 +141,4 @@ else
 fi
 
 # create the python initialization service
-create_s6_service "py_init" "cd /mnt/project && /opt/conda/envs/env/bin/python /src/cortex/serve/init/script.py"
+create_s6_service "py_init" "cd /mnt/project && $source_env_file_cmd && /opt/conda/envs/env/bin/python /src/cortex/serve/init/script.py"
```
RobertLucian, line 144: We no longer need `$source_env_file_cmd` prepended to `script.py`. This was only for testing purposes.

RobertLucian (commented):
Had some comments, but overall, LGTM! Can't approve the PR because I'm the author of this.

pkg/types/spec/api.go
```diff
@@ -35,19 +35,18 @@ import (
...
+	SpecID           string             `json:"spec_id"`
+	PredictorID      string             `json:"predictor_id"`
+	DeploymentID     string             `json:"deployment_id"`
+	Key              string             `json:"key"`
+	PredictorKey     string             `json:"predictor_key"`
+	LastUpdated      int64              `json:"last_updated"`
+	MetadataRoot     string             `json:"metadata_root"`
+	ProjectID        string             `json:"project_id"`
+	ProjectKey       string             `json:"project_key"`
+	LocalModelCaches []*LocalModelCache `json:"local_model_cache"` // local only
```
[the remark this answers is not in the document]
RobertLucian: @deliahu at a second look, it seems like removing `LocalModelCaches` is still necessary due to one reason: When re-deploying an API of whose model's contents have changed, without the `LocalModelCaches` present in the API spec, the API wouldn't get updated. The function responsible for this is `areAPIsEqual` in the `local/api.go` file.

Assuming the above situation doesn't exist, one small setback which I think falls outside the scope of this PR is that we'd need to start using the model IDs from the running containers to determine if a model can be removed from the Cortex cache or not - right now, we are relying on the API spec to tell us that. I think keeping this in the API spec is better because a container is ephemeral, whereas the API spec will be there after a reboot as well - just as the model will still be there.

Because of the first reason listed in this post, I think the presence of `LocalModelCaches` in the API spec is actually justified.

pkg/operator/resources/validations.go
```diff
@@ -102,7 +102,8 @@ func ValidateClusterAPIs(apis []userconfig.API, models *[]spec.CuratedModelResou
 	for i := range apis {
 		api := &apis[i]
 		if api.Kind == userconfig.RealtimeAPIKind || api.Kind == userconfig.BatchAPIKind {
-			if err := spec.ValidateAPI(api, models, projectFiles, types.AWSProviderType, config.AWS, config.K8s); err != nil {
+			models := []spec.CuratedModelResource{}
```
deliahu, line 105: Is it still necessary to have this here?

pkg/types/spec/validations.go
```diff
@@ -746,33 +762,41 @@ func validatePredictor(
...
 ) error {
 	predictor := api.Predictor
 
-	if predictor.Models != nil && predictor.ModelPath != nil {
-		return ErrorConflictingFields(userconfig.ModelPathKey, userconfig.ModelsKey)
-	}
-	if predictor.Models != nil {
-		if err := validateMultiModelsFields(api); err != nil {
-			return err
+	if providerType == types.AWSProviderType {
```
deliahu, line 770: Can this if statement be removed?

deliahu (approved):

pkg/types/spec/validations.go
RobertLucian: I see this if statement got removed.

pkg/operator/resources/validations.go
RobertLucian: `ValidateAPI` is also called for local deployments. When deploying locally, the `models` argument is required for caching the models locally in the Cortex cache higher up in the call stack.

But here, in `operator/resources/validations.go`, it isn't required. Still, were passing it in, because otherwise, each predictor validation would panic because `*models` can't be set. We could verify if `models != nil` though. What do you think?

PR #1664, opened 2020-12-08
Improve registry.sh script UX

miguelvr:
With the recent GCP merge the `make image-<type>-<provider>` commands require that all provider specific flags are set in `dev/config/env.sh`. They can currently be set as empty, but this seems unnecessary.

This PR improves the UX in 2 ways:
- If building for a specific provider, only that provider's env variables are required (the others do not need to be set)
- Add validation to check if provider specific env variables are set

checklist:

- [x] update examples

### DIFF
--- a/dev/registry.sh
+++ b/dev/registry.sh
@@ -187,6 +187,22 @@ function cleanup_ecr() {
   done
 }
 
+function validate_env() {
+  local provider=$1
+
+  if [ "$provider" = "aws" ]; then
+    if [[ -z ${AWS_REGION} ]] || [[ -z ${AWS_ACCOUNT_ID} ]]; then
+      echo "error: environment variables AWS_REGION and AWS_ACCOUNT_ID should be exported in dev/config/env.sh"
+      exit 1
+    fi
+  elif [ "$provider" = "gcp" ]; then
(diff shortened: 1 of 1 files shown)


dev/registry.sh
```diff
@@ -182,6 +187,22 @@ function cleanup_ecr() {
   done
 }
 
+function validate_env() {
+  local provider=$1
+
+  if [ "$provider" = "aws" ]; then
+    if [[ -z ${AWS_REGION} ]] || [[ -z ${AWS_ACCOUNT_ID} ]]; then
```
RobertLucian, line 194: Let's go with single-bracket conditions here since we really only use single-bracket conditions. That's for keeping it consistent.

dev/registry.sh
```diff
@@ -25,6 +25,11 @@ source $ROOT/dev/util.sh
 
 GCR_HOST=${GCR_HOST:-"gcr.io"}
 
+# set variables to empty strings if they do not exist
```
RobertLucian, line 28: I don't think we really need this comment since this script is for us and it's obvious what it does. What do you think?

RobertLucian (approved):
LGTM with the exception of 2 nits.

miguelvr:
@RobertLucian addressed your comments on my last commit

PR #1592, opened 2020-11-23
Translator example

caleb-kaiser:
closes #<issue ID>

---

checklist:

- [x] run `make test` and `make lint`
- [x] test manually (i.e. build/push all images, restart operator, and re-deploy APIs)

PR #1690, opened 2020-12-12
Optimise API model discovery speed

### LINKED ISSUE #1663: Speed up the searching/validating processes of models on the serving container
#### Description

When an API with many models is deployed (when using the `predictor.models.dir` field), the searching and validating processes of the models is very slow - so slow, that for a large number of models (i.e. 1000), it takes > 10 minutes to run it once.

The 2 culprit functions for that are `validate_models_dir_paths` and `find_all_cloud_models`, with the former taking the most amount of time.

A good example of this is the translator example https://github.com/cortexlabs/cortex/pull/1592.

#### Motivation

Apart from taking a lot of time to update the models' tree once, it's also the CPU that's wasted in the process. This needs to be faster.

#### Reason

Most likely a complexity problem.

RobertLucian:
Closes #1663.

As an example, for https://github.com/cortexlabs/cortex/pull/1592, it reduces the discovery time of models from ~300 seconds down to about 16 seconds (`s3.search` + model dir validation). 9 seconds are for `s3.search` and about 7 seconds are for validating the models. This is while factoring in the fact that there are ~10k paths with the specified prefix in `cortex.yaml#predictor.models.dir`.

- The validating functions can be further optimized.
- A profiling tool has been used for this.

---

checklist:


pkg/workloads/cortex/lib/util.py
```diff
@@ -116,23 +126,32 @@ def remove_non_empty_directory_paths(paths: List[str]) -> List[str]:
     Then after calling this function, it will look like:
     models/tensorflow/iris/1569001258/saved_model.pb
```
RobertLucian, line 127: Assume that paths are either all absolutes or none of them are. Also error if there are mixed paths (absolutes + relatives).

deliahu (approved):

RobertLucian (commented):
A mod requested.

PR #1700, opened 2020-12-15
Update locale environment variables

deliahu:
Setting `LANGUAGE=en_US.UTF-8` seems to be necessary for some applications

Setting `LC_MESSAGES=en_US.UTF-8` is just for good measure


images/onnx-predictor-cpu/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN cd /tmp/ && \
 ENV S6_BEHAVIOUR_IF_STAGE2_FAILS 2
 
 RUN locale-gen en_US.UTF-8
-ENV LANG=en_US.UTF-8 LANGUAGE=en_US.en LC_ALL=en_US.UTF-8
+ENV LANG=en_US.UTF-8 LANGUAGE=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8
```
RobertLucian, line 23: In this case, would it be safer if we set the other locale variables as exemplified [here](https://unix.stackexchange.com/a/199306)? Ignore the SO question, just notice the output of `locale` in that case.

RobertLucian (approved):
Overall, LGTM, just one question which might change things.

images/onnx-predictor-cpu/Dockerfile
deliahu: Thanks for bringing this up! I confirmed that even without setting `LC_MESSAGES`, all of these variables are set properly (via `LC_ALL`), so I've removed it.

PR #1706, opened 2020-12-16
Initial implementation for task runner in python [WIP]

vishalbollu:


pkg/workloads/cortex/serve/start/batch.py
```diff
@@ -41,9 +41,6 @@
     "job_spec": None,
     "provider": None,
     "predictor_impl": None,
-    "predict_route": None,
-    "client": None,
-    "class_set": set(),
```
miguelvr, line 46: i noticed you are removing these fields on a BatchAPI file, is this intentional?

PR #1702, opened 2020-12-16
Move model-related fields in a subfield for the Python Predictor

RobertLucian:
Closes cortexlabs/internal#20, closes cortexlabs/internal#39.

---

checklist:

- [x] run `make test` and `make lint`
- [x] test manually (i.e. build/push all images, restart operator, and re-deploy APIs)
- [x] update examples
- [x] update docs and add any new files to `summary.md` (view in [gitbook](https://docs.cortex.dev/v/master) after merging)

PR #1711, opened 2020-12-18
Support file paths for ONNX predictor

### LINKED ISSUE #1686: Support file paths for ONNX predictor
Ideally, with live reloading enabled (if it's simple). If so, then this should also extend to the Python Predictor. Otherwise, it's only going to be limited to the ONNX predictor, and the error message shown when specifying a file path with the Python predictor should be improved

RobertLucian:
Closes #1686.

Blocked on #1702.

---

checklist:


pkg/operator/operator/k8s.go
```diff
@@ -578,20 +579,41 @@ func pythonDownloadArgs(api *spec.API) string {
...
+			ItemName:         "the project code",
+			HideFromLog:      true,
+			HideUnzippingLog: true,
 		},
 	}
 
+	if api.Predictor.Models.Path != nil && strings.HasSuffix(*api.Predictor.Models.Path, ".onnx") {
+		downloadContainerArs = append(downloadContainerArs, downloadContainerArg{
+			From: *api.Predictor.Models.Path,
+			To:   path.Join(_modelDir, consts.SingleModelName, "1"),
```
deliahu, line 596: nit: I think if we add `ItemName: "the onnx model"` to the struct, it will print a nice message in the downloader container

pkg/operator/operator/k8s.go
```diff
@@ -578,20 +579,41 @@ func pythonDownloadArgs(api *spec.API) string {
...
+	}
+
+	for _, model := range api.Predictor.Models.Paths {
+		if model == nil {
+			continue
+		}
+		if strings.HasSuffix(model.Path, ".onnx") {
+			downloadContainerArs = append(downloadContainerArs, downloadContainerArg{
+				From: model.Path,
+				To:   path.Join(_modelDir, model.Name, "1"),
```
deliahu, line 607: nit: I think if we add `ItemName: "the onnx model"` to the struct, it will print a nice message in the downloader container

pkg/types/spec/validations.go
```diff
@@ -1083,18 +1083,31 @@ func validateONNXPredictor(api *userconfig.API, models *[]CuratedModelResource,
...
+		if strings.HasSuffix(*predictor.Models.Path, ".onnx") {
+			if err := validateONNXModelFilePath(*predictor.Models.Path, projectFiles.ProjectDir(), awsClient, gcpClient); err != nil {
+				return modelWrapError(err)
+			}
+			modelFileResources = []userconfig.ModelResource{
+				{
+					Name: consts.SingleModelName,
+					Path: *predictor.Models.Path,
+				},
+			}
```
deliahu, line 1098: nit: if lines 1093-1098 are the same as 1100-1105, should they be moved to before the `if` on 1089?

pkg/types/spec/validations.go
```diff
@@ -1150,6 +1187,58 @@ func validateONNXPredictor(api *userconfig.API, models *[]CuratedModelResource,
...
+		if err != nil {
+			return err
+		}
+
+		bucket, modelPrefix, err := aws.SplitS3Path(modelPath)
+		if err != nil {
+			return err
+		}
+
+		yes, err := awsClientForBucket.IsS3Prefix(bucket, modelPrefix)
```
deliahu, line 1206: nit: consider renaming `yes` to `isS3Prefix` (also below)

deliahu, line 1206: Should be checking if the prefix exists or if the file exists? E.g. if `modelPrefix` is 1/2, and there is no file named `1/2` but there is a file named `1/22` or `1/2/3` this will return true. If that is not desired, check out `IsS3File()`. If it is desired, disregard this comment :)

pkg/operator/operator/k8s.go
RobertLucian: Thought of not doing it because adding this log would have been misleading since we advertise the `predictor.models` predictor as having the LR active. That being said, I actually think this is a good idea because it highlights to the user that a model gets downloaded at deploy time (suggestive that it's not LR) and it's also potentially easier for us to debug if the user shares their logs.

pkg/types/spec/validations.go
RobertLucian: Moved it after the if/else block.

pkg/types/spec/validations.go
RobertLucian: The intention was to check the file, not the prefix. Fixed this! Thanks!

RobertLucian: Makes sense.

PR #1741, opened 2020-12-25
Configure log level within the Cortex API spec config

### LINKED ISSUE #1484: Configure log level within the Cortex API spec config
#### Description

Add a field in the `cortex.yaml` API spec that allows the user to set the log level for their API deployments:

```yaml
- name: <string>  # API name (required)
  predictor:
    ...
    log_level: info # allow CRITICAL, ERROR, WARNING, INFO, DEBUG, NOTSET
    ... 
```

The default log level set would be `INFO`.

Consider adding a guide to showcase using `cx_logger()` in predictor implementations.

#### Motivation

To prevent spamming the log lines with unnecessary information. As pointed out by @vackosar.

#### Issues to sort out

1. Looks like uvicorn ignores our log levels from `log_config.yaml`, and for that, we'd have to have something like:
https://github.com/tiangolo/fastapi/issues/1508#issuecomment-638365277
1. The log levels set in `log_config.yaml` don't seem to apply to the `cx_logger` - `info` is still the default log level regardless of which level is set in `log_config.yaml`.

RobertLucian:
Closes #1484.

---

checklist:


pkg/cortex/serve/log_config.yaml
```diff
@@ -19,31 +19,54 @@
 version: 1
 disable_existing_loggers: False
 formatters:
-  default:
+  default_without_pid:
```
vishalbollu, line 22: This doesn't look like it is being used, let's remove it.

pkg/lib/tensorflow/logs.go
```diff
@@ -0,0 +1,33 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tensorflow
+
+// Value to set for TF_CPP_MIN_LOG_LEVEL environment variable. Default is 0.
+func NumericLogLevelFromLogLevel(logLevel string) int {
```
vishalbollu, line 20: This function can be added to userconfig/log_level as a helper right?

pkg/cortex/serve/init/bootloader.sh
```diff
@@ -52,44 +52,68 @@ if [ "$CORTEX_PROVIDER" != "local" ]; then
...
     fi
-fi
 
-# install pip packages
-if [ -f "/mnt/project/requirements.txt" ]; then
-    pip --no-cache-dir install -r /mnt/project/requirements.txt
+}
+
+# install user dependencies
+tempf=$(mktemp)
```
vishalbollu, line 100: As discussed on call, reorder this to stream output to stdout if cortex log level is debug or info and only buffer output if its warning or higher.

vishalbollu (changes requested):

pkg/lib/tensorflow/logs.go
RobertLucian: Yes, that's a helper function. Not 100% if that should be in `userconfig/log_level` though.

pkg/cortex/serve/init/bootloader.sh
```diff
@@ -52,43 +52,67 @@ if [ "$CORTEX_PROVIDER" != "local" ]; then
...
-if [ -f "/mnt/project/requirements.txt" ]; then
-    pip --no-cache-dir install -r /mnt/project/requirements.txt
+}
+
+# install user dependencies
+if [ "$CORTEX_LOG_LEVEL" = "DEBUG" ] || [ "$CORTEX_LOG_LEVEL" = "INFO" ]; then
+    install_deps
+else
+    tempf=$(mktemp)
+    set +e
```
vishalbollu, line 104: A commenting describing this behaviour would be helpful

vishalbollu (approved):

PR #1746, opened 2020-12-29
Catch operator panics in go routines with sentry

RobertLucian:
Fixes cortexlabs/internal#47.

---

checklist:


pkg/lib/routines/routines.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/cortexlabs/cortex/pkg/lib/exit"
+)
+
+func RunWithPanicHandler(f func()) {
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				err := ErrorUnexpectedError(fmt.Sprintf("%v", r))
+				exit.Error(err)
```
deliahu, line 30: I'm not sure if we always want to exit; I think in the operator it would be good to just send the telemetry, print the stack trace, and not exit; that is what we do with the operator API server [here](https://github.com/cortexlabs/cortex/blob/master/pkg/operator/endpoints/middleware.go#L43) ([implementation](https://github.com/cortexlabs/cortex/blob/master/pkg/operator/endpoints/respond.go#L71)) and in the cron ([here](https://github.com/cortexlabs/cortex/blob/master/pkg/lib/cron/cron.go#L72)). Perhaps in the CLI exiting makes sense. What do you think?

pkg/lib/routines/routines.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+
+	"github.com/cortexlabs/cortex/pkg/lib/exit"
+)
+
+func RunWithPanicHandler(f func()) {
+	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				err := ErrorUnexpectedError(fmt.Sprintf("%v", r))
```
deliahu, line 29: I think this error can be removed like the other one, and instead we can use `err := errors.CastRecoverError(r)`

pkg/lib/routines/routines.go
RobertLucian: As per our last conversation, we are no longer exiting if a panic occurs and instead we recover and report the error.

deliahu (approved):

PR #1756, opened 2021-01-04
Allow users to configure shared memory for IPC

### LINKED ISSUE #1638: Allow setting the shared memory (shm) size for the API container
Hello.

I'm looking for the way to set shm size of cortex workers.

According to the aws documentation(https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_linuxparameters), it seems possible to set shared memory size of containers.

I've implemented pytorch dataloader (num_workers > 0) to use prefetch feature to get images from s3. Because it takes much time to download the files. 

btw only I got was this message, "error: unexpected bus error encountered in worker. this might be caused by insufficient shared memory (shm)".

Is it possible to set shm size or any workarounds?

Many thanks in advance.

vishalbollu:
closes #1638 

---

checklist:


docs/workloads/batch/configuration.md
```diff
@@ -14,6 +14,7 @@
     image: <string> # docker image to use for the Predictor (default: quay.io/cortexlabs/python-predictor-cpu:master or quay.io/cortexlabs/python-predictor-gpu:master based on compute)
     env: <string: string>  # dictionary of environment variables
     log_level: <string>  # log level that can be "debug", "info", "warning" or "error" (default: "info")
+    shm_size: <string>  # size of shared memory (/dev/shm) for sharing data between multiple processes, e.g. 64Mi or 1Gi (default: Null)
```
RobertLucian, line 17: What do you think of moving `shm_size` to the compute section instead

pkg/operator/operator/k8s.go
```diff
@@ -294,10 +325,12 @@ func TensorFlowPredictorContainers(api *spec.API) ([]kcore.Container, []kcore.Vo
 	return containers, volumes
 }
 
-func ONNXPredictorContainers(api *spec.API) []kcore.Container {
+func ONNXPredictorContainers(api *spec.API) ([]kcore.Container, []kcore.Volume) {
 	resourceList := kcore.ResourceList{}
 	resourceLimitsList := kcore.ResourceList{}
```
RobertLucian, line 330: nit: rename `resourceList` and `resourceLimitsList` to `apiResourceList` and `apiResourceLimitsList`.

pkg/operator/operator/k8s.go
```diff
@@ -294,10 +325,12 @@ func TensorFlowPredictorContainers(api *spec.API) ([]kcore.Container, []kcore.Vo
...
 	containers := []kcore.Container{}
```
RobertLucian, line 331: nit: move `containers` after `volumes` (as it's the case for the other predictor types).

pkg/operator/operator/k8s.go
```diff
@@ -294,10 +325,12 @@ func TensorFlowPredictorContainers(api *spec.API) ([]kcore.Container, []kcore.Vo
...
+	apiPodVolumeMounts := defaultVolumeMounts()
+	volumes := DefaultVolumes()
```
RobertLucian, line 333: nit: not exactly the purpose of this PR, but could we rename `DefaultVolumes` to `defaultVolumes`? It doesn't appear to be used outside this package and we can always open it up if we want.

pkg/types/spec/errors.go
```diff
@@ -228,6 +231,14 @@ func ErrorSurgeAndUnavailableBothZero() error {
 	})
 }
 
+func ErrorShmSizeCannotExceedMem(shmSize k8s.Quantity, mem k8s.Quantity) error {
+	const maxNumProcesses int32 = 1
```
RobertLucian, line 235: `maxNumProcesses` doesn't appear to be used anywhere. Can we remove it?

pkg/types/userconfig/api.go
```diff
@@ -383,6 +384,10 @@ func (predictor *Predictor) UserStr() string {
 	sb.WriteString(fmt.Sprintf("%s: %s\n", ProcessesPerReplicaKey, s.Int32(predictor.ProcessesPerReplica)))
 	sb.WriteString(fmt.Sprintf("%s: %s\n", ThreadsPerProcessKey, s.Int32(predictor.ThreadsPerProcess)))
 
+	if predictor.ShmSize != nil {
```
RobertLucian, line 387: Should we place this after `log_level` since this comes before it in the docs?

RobertLucian (approved):

PR #1789, opened 2021-01-15
Support GCP instances with multiple GPUs

### LINKED ISSUE #1784: GCP - support instances with multiple GPUs
### Description

Allow configuring instances to have multiple GPUs

deliahu:
closes #1784

---

checklist:

- [ ] update docs and add any new files to `summary.md` (view in [gitbook](https://cortex-labs.gitbook.io/staging/-MOmCGMADSRNQahK3Kox/) after merging)


pkg/types/clusterconfig/cluster_config_gcp.go
```diff
@@ -136,6 +137,20 @@ var UserGCPValidation = &cr.StructValidation{
 				AllowExplicitNull: true,
 			},
 		},
+		{
+			StructField: "AcceleratorsPerInstance",
+			Int64PtrValidation: &cr.Int64PtrValidation{
+				AllowExplicitNull: true,
+			},
+			DefaultDependentFields: []string{"AcceleratorType"},
```
RobertLucian, line 145: off: I take this field `DefaultDependentFields` (alongside `DefaultDependentFieldsFunc`) can be used for `preemptible` and `on_demand_backup` in https://github.com/cortexlabs/cortex/pull/1791.

pkg/lib/configreader/reader.go
```diff
@@ -1130,11 +1130,24 @@ func setField(val interface{}, destStruct interface{}, fieldName string) error {
 		debug.Ppg(destStruct)
 		return errors.Wrap(ErrorCannotSetStructField(), fieldName)
 	}
+
+	if val == nil {
```
RobertLucian, line 1134: What is this for? Is this whole if block implemented because we now return `nil` interfaces from `DefaultDependentFieldsFunc`?

RobertLucian (approved):
LGTM! For reference, in the GCP validation on master, we don't yet validate for the number of accelerators, so this PR is good in this regard.

pkg/types/clusterconfig/cluster_config_gcp.go
deliahu: Possibly, we can discuss this live

pkg/lib/configreader/reader.go
deliahu: Yes! In general, this block is necessary to set fields in a struct to nil (which I guess we had never done, but now we do in the `DefaultDependentFieldsFunc` I added)

PR #1812, opened 2021-01-20
Build images and update operator before executing e2e tests

RobertLucian:
checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -265,8 +265,12 @@ test-e2e:
 	@$(MAKE) test-e2e-aws
 	@$(MAKE) test-e2e-gcp
 test-e2e-aws:
+	@$(MAKE) images-all-aws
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -265,8 +265,12 @@ test-e2e:
 	@$(MAKE) test-e2e-aws
 	@$(MAKE) test-e2e-gcp
 test-e2e-aws:
+	@$(MAKE) images-all-aws
+	@$(MAKE) operator-update-aws
```
deliahu, line 269: nit: I think this can be `operator-restart-aws` since the operator image will be built and pushed in the previous line (also for GCP)

deliahu (approved):

Makefile
RobertLucian: That makes sense!

PR #1819, opened 2021-01-22
Prevent the user from specifying non-s3/gcs model paths

### LINKED ISSUE #1817: Non-S3/GCS paths are not properly handled in the API spec when deploying
### Description

When a non S3/GCS path is used in the API spec (in the `predictor.models` section), the CLI doesn't error with a message saying that only S3/GCS paths are accepted.

### Solution

When non S3/GCS paths are used, the CLI deploy command should fail with a command like "only S3/GCS paths are accepted".

RobertLucian:
Fixes #1817.

---

checklist:

### DIFF
--- a/pkg/types/spec/validations.go
+++ b/pkg/types/spec/validations.go
@@ -606,18 +606,16 @@ func multiModelValidation(fieldName string) *cr.StructFieldValidation {
 				{
 					StructField: "Path",
 					StringPtrValidation: &cr.StringPtrValidation{
-						Required:        false,
-						InvalidPrefixes: []string{".", "./", "/"},
-						Validator:       checkForInvalidBucketProvider,
+						Required:  false,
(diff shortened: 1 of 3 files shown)


pkg/types/spec/validations.go
```diff
@@ -606,14 +606,18 @@ func multiModelValidation(fieldName string) *cr.StructFieldValidation {
 				{
 					StructField: "Path",
 					StringPtrValidation: &cr.StringPtrValidation{
-						Required: false,
+						Required:        false,
+						InvalidPrefixes: []string{".", "./", "/"},
```
deliahu, line 610: Is it necessary to have this check, since it would get caught on the line below (and that seems like it might be a nicer error message in that case)?

RobertLucian: No, it isn't required. Just thought it would be nice. What do you think?

PR #1820, opened 2021-01-22
Prevent panic when k8s job doesn't exist

RobertLucian:
To discuss with @vishalbollu.


miguelvr:
@RobertLucian just as a note, this can probably be simplified into

```go
if k8sJob == nil {
    return nil
}
```

as opposed to checking if it's different than nil everywhere

vishalbollu (approved):

PR #1856, opened 2021-02-03
Nightly build for E2E Tests

### LINKED ISSUE #1768: Run nightly e2e tests

miguelvr:
closes #1768 

---

checklist:


.circleci/config.yml
```diff
@@ -68,8 +72,77 @@ jobs:
...
+    environment:
+      CORTEX_TEST_REALTIME_DEPLOY_TIMEOUT: 60
+      CORTEX_TEST_BATCH_DEPLOY_TIMEOUT: 30
+      CORTEX_TEST_BATCH_JOB_TIMEOUT: 200
+      NUM_BUILD_PROCS: 2
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - install-go
```
deliahu, line 87: Is go necessary?

.circleci/config.yml
```diff
@@ -68,8 +72,77 @@ jobs:
...
+      - run:
+          name: Generate Cluster Config
+          command: |
+            cat \<< EOF > ./cluster-aws.yaml
```
deliahu, line 91: Should we add a test for GCP?

.circleci/config.yml
```diff
@@ -68,8 +72,77 @@ jobs:
...
+            image_nvidia: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/nvidia:latest
+            image_fluent_bit: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/fluent-bit:latest
+            image_istio_proxy: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/istio-proxy:latest
+            image_istio_pilot: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/istio-pilot:latest
+            image_prometheus: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus:latest
+            image_prometheus_config_reloader: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus-config-reloader:latest
+            image_prometheus_operator: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus-operator:latest
+            image_prometheus_statsd_exporter: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus-statsd-exporter:latest
+            image_prometheus_to_cloudwatch: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus-to-cloudwatch:latest
```
deliahu, line 116: What do you think of using the `master` tag from quay.io, which is built/pushed on each build, e.g. `image_operator: quay.io/cortexlabs/operator:master`

.circleci/config.yml
```diff
@@ -68,8 +72,77 @@ jobs:
...
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - install-go
+      - run:
+          name: Generate Cluster Config
+          command: |
+            cat \<< EOF > ./cluster-aws.yaml
+            cluster_name: cortex
+            provider: aws
+            region: ${AWS_REGION}
```
deliahu, line 94: I don't mind hardcoding the region and the bucket, or if you prefer to use the environment variables, maybe they should be more clearly just for this purpose, e.g. `NIGHTLY_AWS_REGION`, what do you think?

.circleci/config.yml
```diff
@@ -68,8 +72,77 @@ jobs:
...
+            image_prometheus: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus:latest
+            image_prometheus_config_reloader: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus-config-reloader:latest
+            image_prometheus_operator: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus-operator:latest
+            image_prometheus_statsd_exporter: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus-statsd-exporter:latest
+            image_prometheus_to_cloudwatch: ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/cortexlabs/prometheus-to-cloudwatch:latest
+            EOF
+      - run:
+          name: Install Dependencies
+          command: |
+            pip install awscli
```
deliahu, line 121: Do we need the aws cli?

.circleci/config.yml
```diff
@@ -68,8 +72,77 @@ jobs:
...
+            pip install -e ./test/e2e
+            make cli
```
deliahu, line 123: Would it work to `pip install` the master version of the CLI that we push to S3, e.g. `pip install https://s3-us-west-2.amazonaws.com/get-cortex/master/python/cortex-master.tar.gz`

deliahu (commented):
This is awesome! We can discuss tomorrow whether it makes sense to build the assets or use the pre-built ones from the latest CI build

.circleci/config.yml
miguelvr: to build the CLI only, we can drop it if we use it the one in S3

.circleci/config.yml
miguelvr: I am adding it soon

.circleci/config.yml
miguelvr: that makes sense and it will save a ton of time 👍  good idea

.circleci/config.yml
miguelvr: that makes sense

.circleci/config.yml
miguelvr: for pushing the images only, can be removed if we use the pre-built ones

.circleci/config.yml
miguelvr: will try it

miguelvr:
@deliahu addressed all the comments

let's sync in order to add the env vars / create the test environments

.circleci/config.yml
```diff
@@ -29,6 +32,42 @@ commands:
...
+        description: cluster config file path
+        type: string
+        default: ./cluster.yaml
+    steps:
+      - run:
+          name: Create Cluster
+          command: cortex cluster up --config << parameters.config >> --configure-env << parameters.provider >> -y
+      - run:
+          name: Run E2E Tests
+          command: pytest -v test/e2e/tests -k << parameters.provider >> --<< parameters.provider >>-env << parameters.provider >>
```
deliahu, line 61: nit: would it be possible to combine `--aws-env`/`--gcp-env` into one flag `--env`?

.circleci/config.yml
```diff
@@ -29,6 +32,42 @@ commands:
...
+        type: enum
+        enum: [gcp, aws]
+      config:
+        description: cluster config file path
+        type: string
+        default: ./cluster.yaml
+    steps:
+      - run:
+          name: Create Cluster
+          command: cortex cluster up --config << parameters.config >> --configure-env << parameters.provider >> -y
```
deliahu, line 58: Should this be `cortex cluster-gcp up` for GCP?

deliahu, line 58: I'm thinking it might be good to explicitly use AWS credentials, e.g. for AWS we could set env vars like `NIGHTLY_AWS_ACCESS_KEY_ID` and `NIGHTLY_AWS_SECRET_ACCESS_KEY`, what do you think? I'm not as familiar with the best way to do this for GCP.

miguelvr: indeed, it should

.circleci/config.yml
miguelvr: potentially, but I think that is out of the scope of this PR

.circleci/config.yml
miguelvr: sure, I can do that.

For GCP, we need to save the whole JSON file as an env var, and then create a file from its content

deliahu (approved):

PR #1862, opened 2021-02-03
Add script for installing the core dependencies of Cortex

### LINKED ISSUE #1779: Improve experience when Dockerfile modifications uninstall cortex_internal package
## Option 1

Pack installation procedures of core Cortex dependencies into a script users can easily run when building custom Docker images

### Example

```Dockerfile
FROM cortexlabs/python-predictor-cpu:0.26

# run a bunch of stuff for the creation of the custom image
# in the process, the python packages get removed (i.e. due to upgrading the Python runtime)

# reinstall core Cortex dependencies
RUN /usr/local/cortex/install-core-dependencies.sh
```

## Option 2 (could be in place of or in addition to Option 1)

At runtime, detect that the cortex_internal package has been uninstalled, and reinstall it

## Additional tasks

* Add docs for how to change the python version:

```Dockerfile
FROM quay.io/cortexlabs/python-predictor-gpu-slim:0.26.0

# update python runtime version
RUN conda install conda-forge::python=3.7
# and reinstall core cortex runtime dependencies
RUN pip --no-cache-dir install \
    -r /src/cortex/serve/serve.requirements.txt \
    /src/cortex/serve/
```

RobertLucian:
Closes #1779

---

checklist:

- [x] update docs and add any new files to `summary.md` (view in [gitbook](https://cortex-labs.gitbook.io/staging/


pkg/cortex/serve/init/install-core-dependencies.sh
```diff
@@ -0,0 +1,47 @@
...
+function module_exists() {
+    module_name=$1
+    python -c "import importlib, sys; loader = importlib.util.find_spec('$module_name'); sys.exit(1) if loader is None else sys.exit(0)"
+}
+
+temp_freeze=$(mktemp)
+pip freeze > "${temp_freeze}"
+
+if ! grep -q "cortex-internal" "${temp_freeze}"; then
+    pip install --no-cache-dir -U \
```
miguelvr, line 28: why the `-U` flag? Couldn't this cause problems if a dependency has breaking changes?

pkg/cortex/serve/init/install-core-dependencies.sh
```diff
@@ -0,0 +1,47 @@
...
+            tensorflow-cpu==2.3.0 \
+            tensorflow-serving-api==2.3.0
+    fi
+elif [ "${CORTEX_IMAGE_TYPE}" = "onnx-predictor-cpu" ] && ! module_exists "onnxruntime"; then
+    pip install --no-cache-dir -U \
+        onnxruntime==1.4.0
+elif [ "${CORTEX_IMAGE_TYPE}" = "onnx-predictor-gpu" ] && ! module_exists "onnxruntime"; then
+    pip install --no-cache-dir -U \
+        onnxruntime-gpu==1.4.0
+fi
```
miguelvr, line 47: why is this necessary if the base images already have these requirements?

pkg/cortex/serve/init/install-core-dependencies.sh
RobertLucian: If the dependencies' versions of `cortex-internal` don't match with those that already are in the environment, then upgrade them to the right versions. We need to ensure that the `cortex-internal` package is running on the right set of dependencies. If it weren't for that `-U` flag, then there could be a version of `uvicorn` that's far too old for `cortex-internal` to work correctly.

That being said, this is not ideal, because we don't have ranges associated with the dependencies. This is something that needs to be addressed in a subsequent PR.

pkg/cortex/serve/init/install-core-dependencies.sh
RobertLucian: The base images have these requirements indeed, but this avoids the situation where a user changes/removes the conda environment, and then they are left with an environment that doesn't have `cortex-internal` package installed. We've had that happen to our users. And it's better to give them a single script to run rather than give them _this_.

Let me know what you think!

miguelvr: sounds good

miguelvr (approved):
LGTM

docs/workloads/dependencies/images.md
```diff
@@ -44,6 +44,18 @@ RUN pip install --no-cache-dir pandas \
     && conda clean -a
 ```
 
+If you need to re-install the core pip dependencies for Cortex onto your custom image in the existing conda environment, run the following command:
```
deliahu, line 47: Do you mind adding documentation for how to change the python version?

images/python-predictor-inf/Dockerfile
```diff
@@ -103,7 +104,12 @@ RUN test "${SLIM}" = "true" || ( \
             torchvision==0.6.1 \
     )
 
+COPY pkg/cortex/serve/init/install-core-dependencies.sh /usr/local/cortex/install-core-dependencies.sh
+RUN chmod +x /usr/local/cortex/install-core-dependencies.sh
+
 COPY pkg/cortex/serve/ /src/cortex/serve
+ENV CORTEX_LOG_CONFIG_FILE /src/cortex/serve/log_config.yaml
```
deliahu, line 111: This can be separated to a new PR or commit

pkg/cortex/serve/init/install-core-dependencies.sh
```diff
@@ -0,0 +1,47 @@
...
+
+function module_exists() {
+    module_name=$1
+    python -c "import importlib, sys; loader = importlib.util.find_spec('$module_name'); sys.exit(1) if loader is None else sys.exit(0)"
+}
+
+temp_freeze=$(mktemp)
+pip freeze > "${temp_freeze}"
+
+if ! grep -q "cortex-internal" "${temp_freeze}"; then
```
deliahu, line 27: nit: we can use `module_exists()` here instead of pip freeze (and delete lines 24-25 and 33)

docs/workloads/dependencies/images.md
```diff
@@ -52,7 +52,9 @@ If you need to upgrade the Python Runtime version from its current 3.6.9 on your
 FROM quay.io/cortexlabs/python-predictor-cpu-slim:master
 
 # upgrade python runtime version
+RUN conda update -n base -c defaults conda
```
deliahu, line 57: What is the purpose of this line? We can discuss it live

docs/workloads/dependencies/images.md
```diff
@@ -44,6 +44,25 @@ RUN pip install --no-cache-dir pandas \
     && conda clean -a
 ```
 
+If you need to upgrade the Python Runtime version from its current 3.6.9 on your custom image using the existing conda environment, follow this procedure:
```
deliahu, line 47: Do you think we should also mention how to update the python version when we talk about `dependencies.sh` (in `system-packages.md`)?

PR #1863, opened 2021-02-04
Delete unused prometheus volume on cluster down

miguelvr:
checklist:


cli/cmd/cluster_gcp.go
```diff
@@ -263,6 +265,15 @@ var _clusterGCPDownCmd = &cobra.Command{
 
 		fmt.Print("￮ spinning down the cluster ")
 
+		uninstallCmd := "/root/uninstall.sh"
+		if _flagClusterGCPDownKeepVolumes {
+			uninstallCmd += " --keep-volumes"
+		}
+		_, _, err = runGCPManagerAccessCommand(uninstallCmd, *accessConfig, nil, nil)
+		if err != nil {
+			exit.Error(err)
```
vishalbollu, line 274: For cluster down, it is generally preferable to have best-effort deletes. 

Consider the following scenario: the first `cluster-gcp down` request fails on the delete cluster step. In the next retry, the `cluster-gcp down` command fails at the removing the volumes step because the previous cluster down attempt prevented us from getting kubectl config. The user can't retry `cluster-gcp down` because it fails at an unrelated step.

cli/cmd/cluster.go
```diff
@@ -97,6 +98,7 @@ func clusterInit() {
 	addClusterRegionFlag(_clusterDownCmd)
 	addAWSCredentialsFlags(_clusterDownCmd)
 	_clusterDownCmd.Flags().BoolVarP(&_flagClusterDisallowPrompt, "yes", "y", false, "skip prompts")
+	_clusterDownCmd.Flags().BoolVar(&_flagClusterDownKeepVolumes, "keep-volumes", false, "keep cortex provisioned persistent volumes")
```
vishalbollu, line 101: This flag might be renamed to --keep-cloud-resources so that it can be used to preserver volumes, buckets, policies and log groups. For now I think it is fine.

manager/uninstall.sh
```diff
@@ -18,8 +18,51 @@ set -e
...
+  kubectl get configmap cluster-config -o jsonpath='{.data.cluster\.yaml}' > ./cluster.yaml
+
+  # delete resources to detach disk
+  python render_template.py ./cluster.yaml manifests/prometheus-monitoring.yaml.j2 | kubectl delete -f - >/dev/null
+  kubectl delete pvc --namespace default prometheus-prometheus-db-prometheus-prometheus-0 >/dev/null
+}
+
+function uninstall_grafana() {
+  kubectl delete statefulset --namespace default grafana >/dev/null
+  kubectl delete pvc --namespace default grafana-storage >/dev/null
```
vishalbollu, line 65: In a scenario where cluster-down fails, it may not be possible to get kube config to delete the volumes via kubectl. We should use cloud specific apis to find and delete volumes without having to rely on kubectl.

cli/cmd/cluster_gcp.go
```diff
@@ -269,9 +269,21 @@ var _clusterGCPDownCmd = &cobra.Command{
...
+		if (exitCode != nil && *exitCode != 0) || err != nil {
+			var errorMessage string
+			if err != nil {
+				errorMessage = err.Error()
+			} else {
+				errorMessage = output
+			}
+
+			fmt.Print("\n\n")
+			fmt.Print(fmt.Sprintf("￮ failed to uninstall components from cluster: %s", errorMessage))
```
vishalbollu, line 282: `failed to uninstall components from cluster` -> `failed to delete persistent disks from storage, please visit https://console.cloud.google.com/compute/disks?project=%s to manually delete the disks`

RobertLucian: Did it like this instead: `failed to delete persistent disks from storage, please visit https://console.cloud.google.com/compute/disks?project=%s to manually delete the disks starting with the %s prefix: %s`.

vishalbollu (approved):
The two outstanding comments can be addressed in follow up tickets

PR #1896, opened 2021-02-17
Allow custom dependency paths

### LINKED ISSUE #1777: Customize dependency search path in cortex.yaml
### Description

Add the ability to specify the search path for dependency files such as `requirements.txt`, `conda-packages.txt` or `dependencies.sh` inside the `cortex.yaml` configuration.

### Motivation

We want to be able to use different `cortex.yaml` configurations each with their own dependencies in the same code base. Our plan is to setup the active at the root of the repository and point to the predictor via the `predictor.path` setting. However, this would require that we also copy all the dependencies files to the root.

Adding an additional config option inside `cortex.yaml` to specify the search path relative to the cortex root similar to `python_path` would work for us. The new option could be called `dependency_path`.

miguelvr:
closes #1777 

---

checklist:


docs/workloads/batch/configuration.md
```diff
@@ -23,6 +23,10 @@ predictor:
   env: <string: string>  # dictionary of environment variables
   log_level: <string>  # log level that can be "debug", "info", "warning" or "error" (default: "info")
   shm_size: <string>  # size of shared memory (/dev/shm) for sharing data between multiple processes, e.g. 64Mi or 1Gi (default: Null)
+  dependency_path: # (optional)
```
deliahu, line 26: I think we should consider also adding a note about this to `python-packages.md` and `system-packages.md`, what do you think?

docs/workloads/batch/configuration.md
```diff
@@ -23,6 +23,10 @@ predictor:
...
+    pip: <string>  # path to requirements.txt (default: requirements.txt)
+    conda: <string>  # path to conda-packages.txt (default: conda-packages.txt)
+    shell: <string>  # path to a shell script for system package installation (default: dependencies.sh)
```
deliahu, line 29: I think we should also consider adding `.cortexignore` to the list of things that can be configured, since otherwise, a change in a different API's `requirements.txt` would cause an API to be considered updated. What do you think?

docs/workloads/batch/configuration.md
deliahu, line 26: nit: consider `dependency_paths` (although I don't have a strong preference either way).

Another option could be:

```yaml
dependencies:
  pip_path:
  conda_path:
  shell_path:
```

miguelvr: maybe it doesn't even need the `_path` suffix

RobertLucian: I think I'd prefer `dependencies` too. Shorter is better. And `dependencies` will make the user think of the right thing. I think going without the `_path` suffix also makes sense. We'll have the documentation to back this up.

docs/workloads/batch/configuration.md
miguelvr: i am not following this

deliahu (approved):

PR #1898, opened 2021-02-19
Fix Task example in docs

deliahu:


docs/workloads/task/example.md
```diff
@@ -56,39 +58,57 @@ scikit-learn==0.23.2
...
+# > {"job_id":"69b183ed6bdf3e9b","api_name":"trainer",...}
+```
+
+Or, using Python `requests`:
+
 ```python
 import cortex
 import requests
 
 cx = cortex.client("aws")
```
RobertLucian, line 94: nit: add a comment like "aws is the name of the environment used in this example".

docs/workloads/task/example.md
```diff
@@ -7,34 +7,36 @@ Create APIs that can perform arbitrary tasks like training or fine-tuning a mode
...
-    s3.upload_file("model.pkl", bucket, os.path.join(key, "model.pkl"))
+import os
+import boto3
+import pickle
+from sklearn.datasets import load_iris
+from sklearn.model_selection import train_test_split
+from sklearn.linear_model import LogisticRegression
+
+
+class Task:
```
RobertLucian, line 18: Since we show how to submit a job using the Python client, should we also show how one can deploy the API using the client? The only downside to this is that the example page gets longer.

RobertLucian (approved):

docs/workloads/task/example.md
deliahu: That is a good question. I think we should have a separate discussion about this. Is it ok to merge it in as is for now?

PR #1943, opened 2021-03-08
Temporarily lock torch to 1.7 for pytorch text-generator gpu tests

vishalbollu:
Torch 1.8.0 came out recently and certain operations don't seem to be supported for CUDA 10.2 cudnn 7/8 on a Tesla T4. It appears to not work for cuda 10.1 cudnn 7 as well.

Torch 1.7 seems to be working as expected. Temporarily set the Torch version to 1.7 till issues on Torch 1.8 are resolved. Here is a related issue: https://github.com/pytorch/pytorch/issues/53336.

---

checklist:

### DIFF
--- a/test/apis/pytorch/text-generator/requirements.txt
+++ b/test/apis/pytorch/text-generator/requirements.txt
@@ -1,2 +1,2 @@
-torch
+torch==1.7.*
 transformers==3.0.*


test/apis/pytorch/text-generator/requirements.txt
```diff
@@ -1,2 +1,2 @@
-torch
+torch==1.7.*
```
RobertLucian, line 1: What do you think of doing this for all the other `requirements.txt` for all of our examples?

RobertLucian (approved):
LGTM!

test/apis/pytorch/text-generator/requirements.txt
vishalbollu: Since we plan on reverting this change once 1.8 becomes more stable, I think it is okay if we do enough to get the tests passing.

RobertLucian: Yeah, that sounds good.

PR #1954, opened 2021-03-15
Use AWS SDK to delete volumes during cluster down

deliahu:


pkg/lib/aws/ec2.go
```diff
@@ -329,3 +330,54 @@ func (c *Client) DescribeVpcs() ([]ec2.Vpc, error) {
...
+			return false
+		}
+		for _, volume := range output.Volumes {
+			if volume == nil {
+				continue
+			}
+			for _, tag := range volume.Tags {
+				if tag.Key != nil && *tag.Key == fmt.Sprintf("kubernetes.io/cluster/%s", clusterName) {
+					volumes = append(volumes, *volume)
+					continue
```
miguelvr, line 370: unnecessary `continue` statement

pkg/lib/aws/ec2.go
```diff
@@ -329,3 +330,54 @@ func (c *Client) DescribeVpcs() ([]ec2.Vpc, error) {
...
+func (c *Client) ListPVCVolumesForCluster(clusterName string) ([]ec2.Volume, error) {
+	var volumes []ec2.Volume
+	err := c.EC2().DescribeVolumesPages(&ec2.DescribeVolumesInput{}, func(output *ec2.DescribeVolumesOutput, lastPage bool) bool {
+		if output == nil {
+			return false
+		}
+		for _, volume := range output.Volumes {
+			if volume == nil {
+				continue
+			}
```
miguelvr, line 366: these two can be merged into a single function, or at least `ListPVCVolumesForCluster` can call `ListVolumes`

The single signature would look like:

```go
func (c *Client) ListVolumes(tags... *ec2.Tag) ([]ec2.Volume, error) {
    // (...)
}
```

pkg/lib/aws/ec2.go
deliahu: Good catch! I meant for this to be `break`

vishalbollu (approved):

PR #1955, opened 2021-03-15
Increase eksctl version to 0.40.0

vishalbollu:
Update eksctl version to address Spot GPUs not joining the cluster.

---

checklist:


RobertLucian (approved):
Tested on MITC and it works!

PR #2006, opened 2021-03-29
Filter eksctl ouput when reporting errors

miguelvr:
checklist:


pkg/lib/strings/operations.go
```diff
@@ -226,3 +227,25 @@ func PluralCustom(singular string, plural string, count interface{}) string {
 	}
 	return plural
 }
+
+// RemoveDuplicates returns a filtered string slice without repeated entries.
+// The ignoreRegex parameter can optionally be used to ignore repeated patters in each slice entry.
```
vishalbollu, line 232: nit: `patters` -> `patterns`

pkg/lib/strings/operations.go
```diff
@@ -226,3 +227,25 @@ func PluralCustom(singular string, plural string, count interface{}) string {
...
+func RemoveDuplicates(strs []string, ignoreRegex *regexp.Regexp) []string {
+	var result []string
+	counter := map[string]int64{}
```
vishalbollu, line 235: `"github.com/cortexlabs/cortex/pkg/lib/sets/strset"` can be used here.

cli/cmd/cluster.go
```diff
@@ -61,6 +62,8 @@ var (
 	_flagClusterDownKeepVolumes bool
 )
 
+var _eksctlPrefixRegex = regexp.MustCompile("^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} \\[.+] {2}")
```
vishalbollu, line 65: I am not sure but you may be able to avoid `\\` if you use backticks instead of `"` around the regex

pkg/lib/strings/operations.go
miguelvr: what are the wins here?

cli/cmd/cluster.go
miguelvr: the IDE escaped it for me, I'm not sure either... I'm not a regex expert, at all 😅

pkg/lib/strings/operations.go
miguelvr: I've looked at the code, and I don't see how I can use the library to replace the counter

vishalbollu: We can not use strset if we want to reduce dependencies in the lib

cli/cmd/cluster.go
vishalbollu: `"^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} \\[.+] {2}"` -> 
```
`^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} \[.+] {2}`
```

vishalbollu (approved):

PR #2007, opened 2021-03-29
Fix async delete

miguelvr:
checklist:

### DIFF
--- a/pkg/operator/resources/asyncapi/api.go
+++ b/pkg/operator/resources/asyncapi/api.go
@@ -147,16 +147,18 @@ func UpdateAPI(apiConfig userconfig.API, projectID string, force bool) (*spec.AP
 func DeleteAPI(apiName string, keepCache bool) error {
 	err := parallel.RunFirstErr(
 		func() error {
-			deployment, err := config.K8s.GetVirtualService(operator.K8sName(apiName))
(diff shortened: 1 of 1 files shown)


pkg/operator/resources/asyncapi/api.go
```diff
@@ -147,16 +147,18 @@ func UpdateAPI(apiConfig userconfig.API, projectID string, force bool) (*spec.AP
 func DeleteAPI(apiName string, keepCache bool) error {
 	err := parallel.RunFirstErr(
 		func() error {
-			deployment, err := config.K8s.GetVirtualService(operator.K8sName(apiName))
```
vishalbollu, line 150: I think you can try to delete the queue by prefix `config.AWS.DeleteQueuesWithPrefix` to potentially cleanup any partial deletes where the vs has been deleted but the queue hasn't.

miguelvr: doesn't that risk deleting queues from other apis?

vishalbollu: Good point, I have created a ticket to address that scenario https://github.com/cortexlabs/cortex/issues/2009. I believe batch apis currently have the problem of delete queues from other apis.

vishalbollu (approved):

RobertLucian (approved):

PR #2011, opened 2021-03-29
Allow using images from other accounts and regions

### LINKED ISSUE #1988: Allow the use of any ECR images that the AWS account has access to
#### Description

Deploying an API that uses an ECR image from a different AWS account that the current AWS account has access to is currently not permitted. That's because we only allow for ECR images that only belong to the current AWS account. It doesn't take into consideration ECR images that were shared with another AWS account.

This can lead to errors like this:
`cortex.exceptions.CortexBinaryException: error: .cortex.yaml: iris-classifier (RealtimeAPI): predictor: image: registry account ID (XXXXX) doesn't match your AWS account ID (YYYYY), and using an ECR registry in a different AWS account is not supported`

#### Solution

At the very least, eliminate the check that verifies if the ECR image is under the same AWS account or not. The best solution is to check if the AWS user really does have access to the ECR image (checking policies, trying to inspect the image, etc).

vishalbollu:
closes #1988 

Operator and worker instances now have the same access. Users can attach polices to enable their cluster to access images from other ecr accounts and regions. This PR removes image validations that prevents users from accessing images from other regions and accounts.

---

checklist:

### DIFF
--- a/pkg/types/spec/validations.go
+++ b/pkg/types/spec/validations.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/cortexlabs/cortex/pkg/consts"
 	"github.com/cortexlabs/cortex/pkg/lib/aws"
 	"github.com/cortexlabs/cortex/pkg/lib/cast"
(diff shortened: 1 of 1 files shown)


RobertLucian (approved):
LGTM!

PR #2013, opened 2021-03-29
Fix downloader daemonsets

RobertLucian:
checklist:

### DIFF
--- a/manager/install.sh
+++ b/manager/install.sh
@@ -397,25 +397,18 @@ function await_pre_download_images() {
   done
 
   echo -n "￮ downloading docker images "
-  printed_dot_count=0
   for ds_name in ${daemonsets[@]}; do
-    printed_dot="false"
     i=0
     until [ "$(kubectl get daemonset $ds_name -n=default -o 'jsonpath={.status.numberReady}')" == "$(kubectl get daemonset $ds_name -n=default -o 'jsonpath={.status.desiredNumberScheduled}')" ]; do
       if [ $i -eq 120 ]; then break; fi  # give up after 6 minutes
       echo -n "."
-      printed_dot="true"
       ((i=i+1))
       sleep 3
     done
     kubectl -n=default delete --ignore-not-found=true daemonset $ds_name &>/dev/null
-    if [ "$printed_dot" == "true" ]; then
-      printed_dot_count=$(($printed_dot_count+1))
-    fi
   done
 
-
-  if [ "$printed_dot_count" == "${#daemonsets[@]}" ]; then echo " ✓"; else echo "✓"; fi
+  echo " ✓"
 }
 
 function validate_cortex() {
--- a/manager/install.sh
+++ b/manager/install.sh
@@ -378,26 +378,13 @@ function start_pre_download_images() {
 }
 
 function await_pre_download_images() {
-  daemonsets=( "image-downloader-cpu" )
+  daemonsets=( "image-downloader-cpu" "image-downloader-gpu" "image-downloader-inf" )
 
-  has_gpu="false"
-  has_inf="false"
-
-  cluster_config_len=$(cat /in/cluster_${CORTEX_CLUSTER_NAME}_${CORTEX_REGION}.yaml | yq -r .node_groups | yq -r length)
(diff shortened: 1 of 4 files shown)


manager/install.sh
```diff
@@ -352,30 +352,70 @@ function start_pre_download_images() {
...
-  if [[ "$CORTEX_INSTANCE_TYPE" == p* ]] || [[ "$CORTEX_INSTANCE_TYPE" == g* ]] || [ -n "$CORTEX_ACCELERATOR_TYPE" ]; then
+  envsubst < manifests/image-downloader-cpu.yaml | kubectl apply -f - &>/dev/null
+
+  has_gpu="false"
+  has_inf="false"
+
+  cluster_config_len=$(cat /in/cluster_${CORTEX_CLUSTER_NAME}_${CORTEX_REGION}.yaml | yq -r .node_groups | yq -r length)
+  for idx in $(seq 0 $(($cluster_config_len-1))); do
+    ng_instance_type=$(cat /in/cluster_${CORTEX_CLUSTER_NAME}_${CORTEX_REGION}.yaml | yq -r .node_groups[$idx].instance_type)
+    if [[ "$has_gpu" == "false" && ( "$ng_instance_type" == p* || "$ng_instance_type" == g* ) ]]; then
```
deliahu, line 363: nit: I think this would work too if you prefer:

```bash
if [[ "$ng_instance_type" == p* || "$ng_instance_type" == g* ]]; then
```

manager/install.sh
```diff
@@ -352,30 +352,70 @@ function start_pre_download_images() {
...
+      has_gpu="true"
+    fi
+    if [[ "$has_inf" == "false" && "$ng_instance_type" == inf* ]]; then
```
deliahu, line 366: nit: I think this would work too if you prefer:

```bash
if [[ "$ng_instance_type" == inf* ]]; then
```

manager/install.sh
```diff
@@ -352,30 +352,70 @@ function start_pre_download_images() {
...
+  echo -n "￮ downloading docker images "
+  printed_dot_count=0
+  for ds_name in ${daemonsets[@]}; do
     printed_dot="false"
     i=0
-    until [ "$(kubectl get daemonset image-downloader -n=default -o 'jsonpath={.status.numberReady}')" == "$(kubectl get daemonset image-downloader -n=default -o 'jsonpath={.status.desiredNumberScheduled}')" ]; do
+    until [ "$(kubectl get daemonset $ds_name -n=default -o 'jsonpath={.status.numberReady}')" == "$(kubectl get daemonset $ds_name -n=default -o 'jsonpath={.status.desiredNumberScheduled}')" ]; do
       if [ $i -eq 120 ]; then break; fi  # give up after 6 minutes
       echo -n "."
       printed_dot="true"
```
deliahu, line 407: nit: I don't think it's necessary to keep track of `printed_dot` and `printed_dot_count`; instead, we can drop `printed_dot` and increment `printed_dot_count` in the loop

deliahu: EDIT: we could drop both `printed_dot` and `printed_dot_count` if we print at least one dot no matter what, or if we print a dot each time a daemonset is deleted (so we are guaranteed to have at least one).

manager/install.sh
```diff
@@ -352,30 +352,70 @@ function start_pre_download_images() {
...
 
 function await_pre_download_images() {
-  if kubectl get daemonset image-downloader -n=default &>/dev/null; then
-    echo -n "￮ downloading docker images "
+  daemonsets=( "image-downloader-cpu" )
+
+  has_gpu="false"
+  has_inf="false"
+
+  cluster_config_len=$(cat /in/cluster_${CORTEX_CLUSTER_NAME}_${CORTEX_REGION}.yaml | yq -r .node_groups | yq -r length)
```
deliahu, line 386: consider splitting this out into two helper functions, `has_gpu()` and `has_inf()` (which can also be used above)

deliahu: EDIT: alternatively, it might be preferred to get the daemonsets that should be deleted from the cluster state, either by checking that the name starts with `image-downloader-`, or by adding/checking a label. What do you think?

deliahu (approved):

PR #2016, opened 2021-03-30
Handle undefined default environment in python client

deliahu:


miguelvr (approved):

dev/deploy_test_iris.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cortex
+
+cx = cortex.client()
```
vishalbollu, line 17: There is already a python test, `deploy_test.py` and `python_version_test.sh`. I think we can just keep one of them.

PR #2018, opened 2021-03-30
Docs for Realtime gRPC

RobertLucian:
Need to check if the following command is valid (from the `example_grpc.md` page):

```bash
grpcurl -plaintext -proto predictor.proto -d '{"text": "hello-world"}' ***.elb.us-west-2.amazonaws.com:80 text_generator.Predictor/Predict
```

---

checklist:


miguelvr:
> This seems extremely confusing to me. Now it's not clear what is related to HTTP or to gRPC. I would strongly suggest creating sections for HTTP and gRPC separately, where the specifics are addressed separately.

I'm referring to the `predictor.md` and `configuration.md` sections.

I think gRPC should be split in a similar fashion to the different predictor types

deliahu (approved):

PR #2053, opened 2021-04-07
Fix tensorflow batch job exit code on success

miguelvr:
checklist:


pkg/types/clusterconfig/cluster_config.go
```diff
@@ -75,6 +75,7 @@ type CoreConfig struct {
 	ImageOperator                   string `json:"image_operator" yaml:"image_operator"`
 	ImageManager                    string `json:"image_manager" yaml:"image_manager"`
 	ImageDownloader                 string `json:"image_downloader" yaml:"image_downloader"`
+	ImageKubexit                    string `json:"image_kubexit" yaml:"image_kubexit"`
```
deliahu, line 78: There may be a few additional places where this should be added (search the code base for e.g. "istio-pilot", "istio_pilot", and "IstioPilot" to find them)

miguelvr: I think those were removed

deliahu: I think it's still missing in `create.md` (but that's the only remaining one I could find)

pkg/operator/operator/k8s.go
```diff
@@ -971,6 +1006,28 @@ func getEnvVars(api *spec.API, container string) []kcore.EnvVar {
...
+		},
+		{
+			Name:  "KUBEXIT_GRAVEYARD",
+			Value: _kubexitGraveyardMountPath,
+		},
+	}
+
+	if deathDeps != nil {
+		envVars = append(envVars, kcore.EnvVar{
+			Name:  "KUBEXIT_DEATH_DEPS",
```
RobertLucian, line 1023: robert: what is this one for?

pkg/operator/operator/k8s.go
```diff
@@ -369,10 +384,22 @@ func pythonPredictorContainers(api *spec.API, envVars []kcore.EnvVar) ([]kcore.C
...
+	return tensorFlowPredictorContainers(api, envVars, tfServingEnvVars, false)
 }
 
-func tensorFlowPredictorContainers(api *spec.API, envVars []kcore.EnvVar) ([]kcore.Container, []kcore.Volume) {
+func TensorFlowPredictorJobContainers(api *spec.API) ([]kcore.Container, []kcore.Volume) {
+	envVars := getEnvVars(api, APIContainerName)
+	envVars = append(envVars, getKubexitEnvVars(APIContainerName)...)
+
+	tfServingEnvVars := getEnvVars(api, _tfServingContainerName)
+	tfServingEnvVars = append(tfServingEnvVars, getKubexitEnvVars(_tfServingContainerName, APIContainerName)...)
```
RobertLucian, line 397: Should we also add `_neuronRTDContainerName` to `getKubexitEnvVars` as a param? If `compute.inf` > 0 and the tensorflow predictor is used, then the `_neuronRTDContainerName` will also be present as a sidecar.

images/tensorflow-serving-cpu/run.sh
```diff
@@ -28,4 +28,9 @@ if [[ -n ${TF_MAX_BATCH_SIZE} && -n ${TF_BATCH_TIMEOUT_MICROS} ]]; then
 fi
 
 # launch TFS
-tensorflow_model_server "$@"
+
+if [ -f "/mnt/kubexit" ]; then
+  /mnt/kubexit tensorflow_model_server "$@"
```
RobertLucian, line 33: We should also do something like this to `images/neuron-rtd`.

pkg/operator/operator/k8s.go
```diff
@@ -971,6 +1006,28 @@ func getEnvVars(api *spec.API, container string) []kcore.EnvVar {
...
 }
 
+func getKubexitEnvVars(containerName string, deathDeps ...string) []kcore.EnvVar {
+	envVars := []kcore.EnvVar{
+		{
+			Name:  "KUBEXIT_NAME",
+			Value: containerName,
+		},
+		{
+			Name:  "KUBEXIT_GRAVEYARD",
```
RobertLucian, line 1016: And these `KUBEXIT_*` vars are specific to what `kubexit` is expecting, right?

RobertLucian (commented):
Generally LGTM except for an uncertainty about neuron-rtd.

pkg/operator/operator/k8s.go
miguelvr: for defining the death dependencies (e.g. tf serving has death dependency on the API container)

pkg/operator/operator/k8s.go
miguelvr: makes sense

pkg/operator/operator/k8s.go
miguelvr: yes, they are

PR #2069, opened 2021-04-09
Improve the coverage of the e2e test suite

RobertLucian:
checklist:

- [x] self-review


.circleci/config.yml
```diff
@@ -58,7 +58,10 @@ commands:
           command: cortex cluster up << parameters.config >> --configure-env aws -y
       - run:
           name: Run E2E Tests
-          command: pytest -v test/e2e/tests --env aws
+          command: |
+            pytest -v test/e2e/tests --env aws --skip-autoscaling --skip-load --skip-long-running
+            pytest -v test/e2e/tests --env aws -k test_autoscaling
+            pytest -v test/e2e/tests --env aws -k test_load
```
miguelvr, line 64: this can be summed up to a single command

test/e2e/e2e/tests.py
```diff
@@ -290,5 +356,591 @@ def test_task_api(
...
+        while time.time() - start_time <= time_to_run:
+            response = request_prediction(client, api_name, payload)
+
+            assert (
+                response.status_code == HTTPStatus.OK
+            ), f"status code: got {response.status_code}, expected {HTTPStatus.OK}"
+
+            if expectations and "response" in expectations:
+                assert_response_expectations(response, expectations["response"])
+
```
deliahu, line 934: Check that the number of 200s matches what is expected

test/e2e/tests/conftest.py
```diff
@@ -67,14 +82,55 @@ def pytest_configure(config):
...
+            "autoscaling_test_config": {
+                "max_replicas": 20,
+            },
+            "load_test_config": {
+                "realtime": {
+                    "total_requests": 10 ** 6,
+                    "desired_replicas": 50,
+                    "concurrency": 50,
+                    "min_rtt": 0.005,  # measured in seconds
+                    "max_rtt": 1.500,  # measured in seconds
```
deliahu, line 105: check if this can be reduced

deliahu (approved):

test/e2e/tests/conftest.py
RobertLucian: Reduced it to 1s.

.circleci/config.yml
RobertLucian: That would be ideal - and it is possible, but we'd have to create dependencies (within pytest) between the single-request tests and the load/autoscaling tests. There's no point in letting a load test run if the single-request test failed. I saw this plugin https://pytest-dependency.readthedocs.io/en/stable/usage.html that we could use, but from what I read, this only works for tests that sit in the same module.

The simplest next thing is to just create these "dependencies" by separating the pytest command into multiple ones.

PR #2071, opened 2021-04-09
Whitelist IP addresses

### LINKED ISSUE #2003: Investigate ways to whitelist ip addresses for APIs
Restrict access to APIs by whitelisting ip addresses/cidr. Based on this [documentation](https://kubernetes.io/docs/concepts/services-networking/service/#aws-nlb-support) it might be possible to do it at the service level.

- [ ] Investigate `.spec.loadBalancerSourceRanges` although this ticket isn't related, it has information about how istio can be [setup](https://github.com/istio/istio/issues/26507) with loadBalancerSourceRanges
- [ ] Expose ip address ranges list to the user in cluster config for both operator and api loadbalancer

Verify that this approach works for both internet facing and internal loadbalancers.

Additional Information:
Here is a [gist](https://gist.github.com/vishalbollu/59615369067a4ddd17d1752fcd951381) with instructions that seemed to have worked for an internet-facing operator loadbalancer.

vishalbollu:
closes #2003 

---

checklist:


manager/manifests/istio.yaml.j2
```diff
@@ -107,6 +110,10 @@ spec:
             {% endif %}
           service:
             type: LoadBalancer
+            {% if config.get('api_load_balancer_cidr_white_list', [])|length > 0 %}
+            loadBalancerSourceRanges: {{ config['api_load_balancer_cidr_white_list'] }}
+            {% endif %}
+            selector:
```
deliahu, line 116: Is this line intentional?

vishalbollu: removed

deliahu (approved):

PR #2072, opened 2021-04-10
Fix async api responses' statuses (status would be stuck in in_queue state)

RobertLucian:
checklist:


async-gateway/service.go
```diff
@@ -81,14 +81,30 @@ func (s *service) GetWorkload(id string) (GetWorkloadResponse, error) {
...
+			break
+		}
+		if fileStatus == StatusInProgress {
+			status = fileStatus
+		}
+		if fileStatus == StatusCompleted || fileStatus == StatusFailed {
+			status = fileStatus
+			break
+		}
+	}
```
miguelvr, line 106: nit: make this a private method to improve readability

async-gateway/service.go
```diff
@@ -81,14 +81,30 @@ func (s *service) GetWorkload(id string) (GetWorkloadResponse, error) {
...
 
-	status := Status(statusBuf[:])
+	status := StatusInQueue
+	for _, file := range files {
+		fileStatus := Status(file)
+
+		if !fileStatus.Valid() {
+			status = fileStatus
+			break
+		}
```
miguelvr, line 98: I don't get this part. Shouldn't an error be returned if the status is not valid?

async-gateway/service.go
```diff
@@ -66,7 +66,7 @@ func (s *service) CreateWorkload(id string, payload io.Reader, contentType strin
 		return "", err
 	}
 
-	statusPath := fmt.Sprintf("%s/%s/status", prefix, id)
+	statusPath := fmt.Sprintf("%s/%s/%s", prefix, id, StatusInQueue)
 	log.Debug(fmt.Sprintf("setting status to %s", StatusInQueue))
 	if err := s.storage.Upload(statusPath, strings.NewReader(string(StatusInQueue)), "text/plain"); err != nil {
```
deliahu, line 71: Nit: if we aren't checking the contents of the file, it might be more clear to the reader if we create an empty file rather than having the file contain the status

async-gateway/service.go
```diff
@@ -81,14 +81,30 @@ func (s *service) GetWorkload(id string) (GetWorkloadResponse, error) {
 	log := s.logger.With(zap.String("id", id))
 
 	// download workload status
-	statusPath := fmt.Sprintf("%s/%s/status", prefix, id)
-	log.Debug("downloading status file", zap.String("path", statusPath))
-	statusBuf, err := s.storage.Download(statusPath)
+	statusPath := fmt.Sprintf("%s/%s/*", prefix, id)
```
deliahu, line 84: Nit: if this variable is only used for logging, it might be clearer to the reader to remove the variable and log a different variable which is used later in the code, or in-line the path in the log statement

async-gateway/storage.go
```diff
@@ -80,6 +82,25 @@ func (s *s3) Download(key string) ([]byte, error) {
 	return buff.Bytes(), nil
 }
 
+// List lists a set of files from a given S3 path
+func (s *s3) List(key string) ([]string, error) {
```
deliahu, line 86: Nit: it would be good to update the comment and/or the name of the function to indicate that this only works if all sub-paths are only one level deep. Also it might be worth ensuring that the input prefix ends in a slash (assuming it's not empty string)

RobertLucian: Done. Also, for ensuring the key ends with a "/", I added the logic inside the function for that without creating a utility function like `EnsureSuffix`. If we end up needing that in multiple places, that's when we could add the utility instead.

async-gateway/service.go
RobertLucian: I think I agree. The reason I let it pass was that it would have gotten caught in the switch case for the `default` case.

async-gateway/service.go
```diff
@@ -77,32 +77,24 @@ func (s *service) CreateWorkload(id string, payload io.Reader, contentType strin
...
 			ID:     id,
 			Status: status,
 		}, nil
-	case StatusCompleted: // continues execution after switch/case, below
-	default:
-		return GetWorkloadResponse{}, fmt.Errorf("invalid workload status: %s", status)
 	}
 
 	// attempt to download user result
+	prefix := s.workloadStoragePrefix()
```
miguelvr, line 95: if the prefix is required below, compute it once and pass it as argument to `getStatus`

async-gateway/service.go
```diff
@@ -127,6 +119,38 @@ func (s *service) GetWorkload(id string) (GetWorkloadResponse, error) {
...
 
+func (s *service) getStatus(id string) (Status, error) {
+	prefix := s.workloadStoragePrefix()
+	log := s.logger.With(zap.String("id", id))
+
+	// download workload status
+	log.Debug("checking status", zap.String("path", fmt.Sprintf("%s/%s/*", prefix, id)))
+	files, err := s.storage.List(fmt.Sprintf("%s/%s", prefix, id))
+	if err != nil {
+		return Status(""), err
```
miguelvr, line 130: nit: return Status{}, err

miguelvr (approved):
Two minor comments, but otherwise LGMT

async-gateway/service.go
RobertLucian: I think that would be an invalid syntax. It's not a struct.

miguelvr: right, in that case: 
return "", err

async-gateway/service.go
RobertLucian: Since `getStatus` has `s` as the receiver, and `s` has all the required information to compute the prefix, then I don't think we should pass the prefix into `getStatus` - it should be computed from the information that it has been given. Plus, I think that would make it a bit more convoluted.

Let me know what you think.

miguelvr: that's double the allocations, but shouldn't be a problem, I guess.

`getStatus` is a helper method, so I'm not too worried about what it receives as argument

PR #2074, opened 2021-04-12
Misc cleanup

deliahu:


pkg/operator/operator/k8s.go
```diff
@@ -1246,28 +1246,24 @@ func NodeSelectors() map[string]string {
...
+		Key:      "nvidia.com/gpu",
+		Operator: kcore.TolerationOpExists,
+		Effect:   kcore.TaintEffectNoSchedule,
+	},
+	{
+		Key:      "aws.amazon.com/neuron",
+		Operator: kcore.TolerationOpEqual,
+		Value:    "true",
+		Effect:   kcore.TaintEffectNoSchedule,
+	},
```
miguelvr, line 1266: I think it is better to revert this because in the case we need to mutate the tolerations somewhere it will mutate this single variable instead of a value copy

miguelvr (approved):
Just the one single comment, otherwise LGTM

PR #2079, opened 2021-04-12
Add HPA to Async API Gateway to handle different loads of traffic

### LINKED ISSUE #2078: Attach HPA to the async gateway to scale based on the CPU/mem usage

RobertLucian:
Closes #2078.

Increased the load size for the async test to also give a bit of time for the HPA to scale up the gateway.

---

checklist:

- [x] run e2e tests


pkg/lib/k8s/hpa.go
```diff
@@ -63,6 +64,16 @@ func HPA(spec *HPASpec) *kautoscaling.HorizontalPodAutoscaler {
 						},
```
deliahu, line 64: Would unexpected things happen if only cpu or mem are specified in the `spec`? Would it be safer to only add to the `[]kautoscaling.MetricSpec` array if the resource type is not 0?

pkg/lib/k8s/hpa.go
```diff
@@ -179,20 +190,35 @@ func IsHPAUpToDate(hpa *kautoscaling.HorizontalPodAutoscaler, minReplicas int32,
 		return false
 	}
 
-	if len(hpa.Spec.Metrics) != 1 {
+	if len(hpa.Spec.Metrics) != 2 {
+		return false
+	}
```
deliahu, line 195: Update this (and below in this function) if you make changes due to my comment above

deliahu (approved):

PR #2126, opened 2021-04-27
Rename predict/predictor

deliahu:


RobertLucian (approved):
All good (with the exclusion of the compiled protobufs as we have discussed).

PR #2132, opened 2021-04-29
Add option to set shared memory (shm) size for Task API

### LINKED ISSUE #2115: Allow setting the shared memory (shm) size for TaskAPI
### Description

Allow setting shared memory for TaskAPI similar to other inference APIs.

### Motivation

Currently it is not possible to use more than one PyTorch dataloader to train a model because of the shared memory limits, they fail with the error:

```
ERROR: Unexpected bus error encountered in worker. This might be caused by insufficient shared memory (shm).
```

### Additional context

This was previously discussed in "Allow setting the shared memory (shm) size for the API container" #1638.

RobertLucian:
Closes #2115.

---

checklist:

### DIFF
--- a/pkg/types/spec/errors.go
+++ b/pkg/types/spec/errors.go
@@ -232,10 +232,10 @@ func ErrorSurgeAndUnavailableBothZero() error {
 	})
 }
 
-func ErrorShmSizeCannotExceedMem(shmSize k8s.Quantity, mem k8s.Quantity) error {
+func ErrorShmSizeCannotExceedMem(topSection string, shmSize k8s.Quantity, mem k8s.Quantity) error {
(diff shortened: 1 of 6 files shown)


pkg/types/spec/errors.go
```diff
@@ -232,10 +232,10 @@ func ErrorSurgeAndUnavailableBothZero() error {
 	})
 }
 
-func ErrorShmSizeCannotExceedMem(shmSize k8s.Quantity, mem k8s.Quantity) error {
+func ErrorShmSizeCannotExceedMem(topSection string, shmSize k8s.Quantity, mem k8s.Quantity) error {
```
deliahu, line 235: `parentFieldName`

deliahu (approved):

PR #2143, opened 2021-05-06
Move python code into it's own folder

miguelvr:
`pkg` should be for go packages only

checklist:


miguelvr:
the pipeline fail if unavoidable

deliahu (approved):

PR #2145, opened 2021-05-06
Rename python client functions

miguelvr:
checklist:

- [x] update examples


docs/clients/python.md
```diff
@@ -76,36 +81,130 @@ Delete an environment configured on this machine.
...
+
+```python
+ | deploy_realtime_api(api_spec: Dict[str, Any], handler, requirements: Optional[List] = None, conda_packages: Optional[List] = None, force: bool = True, wait: bool = False) -> Dict
+```
+
+Deploy a Realtime API.
+
+**Arguments**:
+
+- `api_spec` - A dictionary defining a single Cortex API. See https://docs.cortex.dev/v/master/workloads/realtime-apis/configuration for schema.
```
deliahu, line 114: we should add a comment in the vicinity of this line and for each docs link below like we have for the other one (`<!-- CORTEX_VERSION_MINOR -->`) - this helps me find which changes to make when doing releases. It's a little hacky, but we do it in `dev/generate_python_client_md.sh`.

python/client/cortex/client.py
```diff
@@ -144,11 +339,11 @@ def create_api(
             if not is_python_set:
                 conda_packages = [f"python={actual_version}", "pip=19.*"] + conda_packages
 
-        if len(requirements) > 0:
+        if requirements is not None:
```
deliahu, line 342: Should we take the branch if it's not none and has length > 0? e.g. `if requirements`? (same question for conda packages)

deliahu (approved):

vishalbollu (approved):

PR #2152, opened 2021-05-07
Decrease file size and project size limits

deliahu:


cli/cmd/deploy.go
```diff
@@ -44,8 +44,8 @@ var (
 	_warningProjectBytes = 1024 * 1024 * 10
 	_warningFileCount    = 1000
 
-	_maxFileSizeBytes    int64 = 1024 * 1024 * 512
-	_maxProjectSizeBytes int64 = 1024 * 1024 * 512
+	_maxFileSizeBytes    int64 = 1024 * 1024 * 50 // 50mb
+	_maxProjectSizeBytes int64 = 1024 * 1024 * 50 // 50mb
```
RobertLucian, line 48: Let's make it 32MiB.

vishalbollu: Changing these limits would cause the CLI to throw an explicit error. The deploy request won't even be sent to the operator.

vishalbollu (approved):

cli/cmd/deploy.go
deliahu: @maskani-moh When you were running into the OOM issue caused by your large files, did the CLI show you a warning about the file size? If not, were you by chance using the `--yes` flag or the Python client?

maskani-moh: I was indeed using the `--yes` flag, so the CLI wasn't complaining about the large files.

PR #2164, opened 2021-05-11
Set default environment name to the cluster name

### LINKED ISSUE #1546: Create an env with cluster name after cluster up
#### Description

Currently, after `cortex cluster up`, the an environment named `aws` is created by default. Consider creating an environment with the cluster name as the default.

An advantage of this change is that consecutive `cortex cluster up` commands won't override the the pervious cluster's environment because now each cluster will have a unique environment.

A disadvantage is that the cluster name can be long so typing out CLI commands may become more tedious (this can be overcome by using the default).

deliahu:
closes #1546


cli/cmd/cluster.go
```diff
@@ -144,26 +144,31 @@ var _clusterUpCmd = &cobra.Command{
...
+		accessConfig, err := getNewClusterAccessConfig(clusterConfigFile)
+		if err != nil {
 			exit.Error(err)
 		}
 
-		accessConfig, err := getNewClusterAccessConfig(clusterConfigFile)
+		envName := _flagClusterUpEnv
+		if envName == "" {
+			envName = accessConfig.ClusterName
+		}
```
deliahu, line 159: note to reviewer: this is a weird diff since the code was rearranged, these are the only lines that were added

PR #2171, opened 2021-05-13
Fix make operator-local command

RobertLucian:
Running the `make operator-local` command is failing because the bucket is not available.

---

checklist:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,6 +24,7 @@ import (
 	"github.com/cortexlabs/cortex/pkg/consts"
 	batch "github.com/cortexlabs/cortex/pkg/crds/apis/batch/v1alpha1"
 	"github.com/cortexlabs/cortex/pkg/lib/aws"
+	cr "github.com/cortexlabs/cortex/pkg/lib/configreader"
 	"github.com/cortexlabs/cortex/pkg/lib/errors"
 	"github.com/cortexlabs/cortex/pkg/lib/hash"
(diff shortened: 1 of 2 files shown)


deliahu (commented):
This looks good to me (@miguelvr do you mind taking a look too?)

Another option could be to download the configmap in `operator_local.sh` to somewhere in the local file system, and then exporting `$CORTEX_CLUSTER_CONFIG_PATH` before starting the local operator. Just a thought for you to consider in case you like that approach better.

pkg/config/config.go
```diff
@@ -60,6 +61,20 @@ func InitConfigs(clusterConfig *clusterconfig.Config, operatorMetadata *clusterc
...
+func getClusterConfigFromConfigMap() (clusterconfig.Config, error) {
+	configMapData, err := K8s.GetConfigMapData("cluster-config")
+	if err != nil {
+		return clusterconfig.Config{}, err
+	}
+	clusterConfig := clusterconfig.Config{}
+	errs := cr.ParseYAMLBytes(&clusterConfig, clusterconfig.FullManagedValidation, []byte(configMapData["cluster.yaml"]))
+	if errors.FirstError(errs...) != nil {
+		return clusterconfig.Config{}, errors.FirstError(errs...)
+	}
```
miguelvr, line 73: I would prefer to have the `cr.ParseYAMLBytes` function to return a single error. You can move the `errors.FirstError` inside

miguelvr (approved):
LGTM after you address the comment I left

PR #2172, opened 2021-05-14
HTTP Reverse Proxy implementation

miguelvr:
checklist:


cmd/proxy/main.go
```diff
@@ -0,0 +1,95 @@
...
+		port              int
+		userPort          int
+		targetConcurrency int
+		maxConcurrency    int
+	)
+
+	flag.IntVar(&port, "port", 8000, "port where the proxy will be served")
+	flag.IntVar(&userPort, "user-port", 8080, "port where the proxy will redirect to the traffic to")
+	flag.IntVar(&targetConcurrency, "target-concurrency", 0, "target concurrency for user container")
+	flag.IntVar(&maxConcurrency, "max-concurrency", 0, "max concurrency for user container")
```
deliahu, line 45: It's not clear to me what these are based on the descriptions, but that's ok, I think we can merge it in for now, since I wanted to discuss this behavior as a team anyways.

pkg/proxy/handler.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func Handler(breaker *Breaker, next http.Handler) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if isKubeletProbe(r) {
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		if breaker != nil {
```
deliahu, line 33: nit: It might be a little nicer to go with this (to match the above):

```go
if breaker == nil {
    next.ServeHTTP(w, r)
    return
}

// add the other code here
```

It would also be an option to merge the two if you prefer, e.g. `if isKubeletProbe(r) || breaker == nil {`

pkg/proxy/handler.go
```diff
@@ -0,0 +1,52 @@
...
+			if err := breaker.Maybe(r.Context(), func() {
+				next.ServeHTTP(w, r)
+			}); err != nil {
+				if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, ErrRequestQueueFull) {
```
deliahu, line 37: I just want to confirm that for now, there is no timeout on the request (assuming that it goes through and wasn't blocked due to e.g. the queue being full).

pkg/proxy/handler.go
```diff
@@ -0,0 +1,52 @@
...
+					http.Error(w, err.Error(), http.StatusServiceUnavailable)
```
deliahu, line 38: Should it `return` after this line just to be safe (since the response has been sent)?

Would the same apply to `w.WriteHeader()` below?

pkg/proxy/breaker.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// Do the thing.
+	thunk()
+	// Report success
+	return nil
+}
+
+// InFlight returns the number of requests currently in flight in this breaker.
+func (b *Breaker) InFlight() int {
+	return int(b.inFlight.Load())
```
deliahu, line 164: Is there a reason to convert this to `int` from `int64`?

deliahu (approved):

pkg/proxy/handler.go
miguelvr: there's none implemented

PR #2176, opened 2021-05-18
Add readiness probe to realtime cortex proxy

miguelvr:
checklist:


pkg/proxy/probe/probe.go
```diff
@@ -0,0 +1,184 @@
...
+		for _, header := range p.HTTPGet.HTTPHeaders {
+			req.Header.Add(header.Name, header.Value)
+		}
+
+		defer func() {
+			// Ensure body is both read _and_ closed so it can be reused for keep-alive.
+			// No point handling errors, connection just won't be reused.
+			_, _ = io.Copy(ioutil.Discard, res.Body)
+			_ = res.Body.Close()
+		}()
```
deliahu, line 163: nit: consider moving this to right after the `if err != nil {` check, just to ensure that no errors can happen in between

pkg/proxy/probe/probe.go
```diff
@@ -0,0 +1,184 @@
...
+		res, err := httpClient.Do(req)
+		if err != nil {
+			return err
+		}
+
+		req.Header.Add(proxy.UserAgentKey, proxy.KubeProbeUserAgentPrefix)
+
+		for _, header := range p.HTTPGet.HTTPHeaders {
+			req.Header.Add(header.Name, header.Value)
+		}
```
deliahu, line 156: Should this be above `httpClient.Do(req)`?

pkg/proxy/probe/probe.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/cortexlabs/cortex/pkg/proxy"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+	kcore "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	_defaultInitialDelaySeconds = 1
```
deliahu, line 38: Is this currently being used?

cmd/proxy/main.go
```diff
@@ -63,7 +66,7 @@ func main() {
 		maxQueueLength = maxConcurrency * 10
```
deliahu, line 66: nit: I think this flag should be required too (since we will set a default from outside the container)

pkg/proxy/probe/probe.go
miguelvr: that won't work because the defer is executed once it exits the current context

pkg/proxy/probe/probe.go
miguelvr: fixed

pkg/proxy/probe/probe.go
miguelvr: I just realized it is not, but maybe it does not make sense to use it (?)

miguelvr: I can add it, it shouldn't be hard, but I need to add a mutex to check if it is the first time the probe is executed

cmd/proxy/main.go
```diff
@@ -79,7 +80,21 @@ func main() {
 	)
 
 	promStats := proxy.NewPrometheusStatsReporter()
-	readinessProbe := probe.NewDefaultProbe(log, target) // TODO: initialize custom probe from flags
+
+	readinessProbe := probe.NewDefaultProbe(log, target)
```
deliahu, line 84: replace with `var readinessProbe` and initialize once

pkg/proxy/probe/probe.go
```diff
@@ -27,24 +27,17 @@ import (
...
-	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 const (
-	_defaultInitialDelaySeconds = 1
-	_defaultTimeoutSeconds      = 3
-	_defaultPeriodSeconds       = 1
-	_defaultSuccessThreshold    = 1
-	_defaultFailureThreshold    = 5
+	_defaultTimeoutSeconds = 3
```
deliahu, line 36: consider making this 1

deliahu (approved):

PR #2181, opened 2021-05-20
Dequeuer proxy for Async and Batch APIs

miguelvr:
checklist:


pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,206 @@
...
+func (h *AsyncMessageHandler) handleFailure(requestID string) error {
+	err := h.updateStatus(requestID, status.AsyncStatusFailed)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to update status to %s", status.AsyncStatusFailed))
+	}
+	return nil
+}
+
+func (h *AsyncMessageHandler) updateStatus(requestID string, status status.AsyncStatus) error {
+	key := fmt.Sprintf("%s/%s/status/%s", h.storagePath, requestID, status)
```
RobertLucian, line 130: These `fmt.Sprintf("...")` keys can be converted to functions and moved to a new file. That way, there's only a single obvious place where the paths are defined. What do you think?

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,206 @@
...
+		Body:        output.Body,
+		ContentType: contentType,
+	}, nil
+}
+
+func (h *AsyncMessageHandler) deletePayload(requestID string) {
+	key := fmt.Sprintf("%s/%s/payload", h.storagePath, requestID)
+	err := h.aws.DeleteS3File(h.config.Bucket, key)
+	if err != nil {
+		h.log.Errorw("failed to delete user payload", "error", err) // TODO: telemetry
```
RobertLucian, line 161: Telemetry todo?

pkg/dequeuer/async_handler_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	_testBucket = "test"
+)
+
+func TestAsyncMessageHandler_Handle(t *testing.T) {
+	log := newLogger(t)
+	awsClient := testAWSClient(t)
```
RobertLucian, line 40: Nice!

pkg/dequeuer/batch_handler.go
```diff
@@ -0,0 +1,216 @@
...
+		metrics:                   statsdClient,
+		log:                       log,
+	}
+}
+
+func (h *BatchMessageHandler) Handle(message *sqs.Message) error {
+	if isOnJobCompleteMessage(message) {
+		err := h.onJobComplete(message)
+		if err != nil {
+			// TODO
```
RobertLucian, line 72: What are these 3 TODOs for? Do we need to address them?

pkg/dequeuer/batch_handler.go
```diff
@@ -0,0 +1,216 @@
...
+func (h *BatchMessageHandler) recordTimePerBatch(elapsedTime time.Duration) error {
+	err := h.metrics.Histogram("cortex_time_per_batch", elapsedTime.Seconds(), h.tags, 1.0)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	return nil
+}
+
+func (h *BatchMessageHandler) submitRequest(messageBody string) error {
+	response, err := http.Post(h.config.TargetURL, "application/json", bytes.NewBuffer([]byte(messageBody)))
```
RobertLucian, line 115: I think it would be good to also set a `"X-Cortex-BatchID"` header. That way, the batch ID can also be retrieved from the request without having to read the `cortex/spec/job.json` file (assuming the user wants the batch ID in the first place).

pkg/dequeuer/dequeuer.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (d *SQSDequeuer) Shutdown() {
+	d.done <- struct{}{}
+}
+
+func (d *SQSDequeuer) handleMessage(message *sqs.Message, messageHandler MessageHandler, done chan struct{}) {
+	messageErr := messageHandler.Handle(message)
+	// if messageErr != nil {
+	// 	// TODO: add to dead letter queue
+	// }
```
RobertLucian, line 148: Is this incomplete? (Same for the other TODOs in this function). If not, then we should add comments explaining why they are there and mention that it doesn't break the dequeuer.

.circleci/config.yml
```diff
@@ -75,18 +76,17 @@ commands:
 
 jobs:
   test:
-    docker:
-      - image: circleci/python:3.6
+    machine:
+      image: ubuntu-1604:202104-01  # machine executor necessary to run go integration tests
     steps:
       - checkout
-      - setup_remote_docker
```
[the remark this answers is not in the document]
miguelvr: yes, that's correct

pkg/lib/aws/credentials.go
```diff
@@ -18,11 +18,11 @@ package aws
 
 // access key ID may be unavailable depending on how the client was instantiated
 func (c *Client) AccessKeyID() *string {
-	if c.sess.Config.Credentials == nil {
+	if c.Sess.Config.Credentials == nil {
```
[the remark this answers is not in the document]
miguelvr: I made sess public to allow for mocking with localstack. What I can do differently is to have a new function to create a client with a custom session. Something on the likes of `awslib.NewForSession(sess *aws.Session)`

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,206 @@
...
+		storagePath: awslib.S3Path(config.Bucket, fmt.Sprintf("%s/workloads/%s", config.ClusterUID, config.APIName)),
+	}
+}
+
+func (h *AsyncMessageHandler) Handle(message *sqs.Message) error {
+	err := h.handleMessage(message)
+	if err != nil {
+		var requestID string
+		if message.Body != nil {
+			requestID = *message.Body
```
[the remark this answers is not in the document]
miguelvr: the requestID is the message body

pkg/dequeuer/batch_handler.go
miguelvr: this is done already (I just didn't have pushed when you commented :) )

pkg/dequeuer/async_handler.go
miguelvr: we need to report the logged errors in general

pkg/dequeuer/batch_handler.go
```diff
@@ -0,0 +1,231 @@
...
+
+func NewBatchMessageHandler(config BatchMessageHandlerConfig, awsClient *awslib.Client, statsdClient statsd.ClientInterface, log *zap.SugaredLogger) *BatchMessageHandler {
+	tags := []string{
+		"api_name:" + config.APIName,
+		"job_id" + config.JobID,
+	}
+
+	return &BatchMessageHandler{
+		config:                    config,
+		jobCompleteMessageRenewal: _jobCompleteMessageRenewal,
```
deliahu, line 65: nit: "renewal" -> "delay"

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,208 @@
...
+
+const (
+	// CortexRequestIDHeader is the header containing the workload request id for the user container
+	CortexRequestIDHeader = "X-Cortex-Request-ID"
+)
+
+type AsyncMessageHandler struct {
+	config      AsyncMessageHandlerConfig
+	aws         *awslib.Client
+	metrics     statsd.ClientInterface
```
deliahu, line 45: this can be removed

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,208 @@
...
+		config:      config,
+		aws:         awsClient,
+		metrics:     metrics,
+		log:         logger,
+		storagePath: awslib.S3Path(config.Bucket, fmt.Sprintf("%s/workloads/%s", config.ClusterUID, config.APIName)),
+	}
+}
+
+func (h *AsyncMessageHandler) Handle(message *sqs.Message) error {
+	err := h.handleMessage(message)
```
deliahu, line 73: add a check for message != nil and not empty (before calling handleMessage())

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,208 @@
...
+		h.log.Errorw("failed processing request", "id", requestID, "error", err)
+		err = h.handleFailure(requestID) // FIXME: should only handle failure if user error (?)
+		if err != nil {
+			return errors.Wrap(err, "failed to handle message failure")
+		}
+	}
+	return nil
+}
+
+func (h *AsyncMessageHandler) handleMessage(message *sqs.Message) error {
```
deliahu, line 89: nit: can accept `requestID string`

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,208 @@
...
+func (h *AsyncMessageHandler) Handle(message *sqs.Message) error {
+	err := h.handleMessage(message)
+	if err != nil {
+		var requestID string
+		if message.Body != nil {
+			requestID = *message.Body
+		}
+
+		h.log.Errorw("failed processing request", "id", requestID, "error", err)
+		err = h.handleFailure(requestID) // FIXME: should only handle failure if user error (?)
```
deliahu, line 81: same as batch regarding the handleFailure, etc

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,208 @@
...
+func (h *AsyncMessageHandler) getPayload(requestID string) (*userPayload, error) {
+	key := fmt.Sprintf("%s/%s/payload", h.storagePath, requestID)
+	output, err := h.aws.S3().GetObject(
+		&s3.GetObjectInput{
+			Key:    aws.String(key),
+			Bucket: aws.String(h.config.Bucket),
+		},
+	)
+	if err != nil {
+		return nil, err
```
deliahu, line 144: nit: `errors.WithStack()`

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,208 @@
...
+	key := fmt.Sprintf("%s/%s/payload", h.storagePath, requestID)
+	err := h.aws.DeleteS3File(h.config.Bucket, key)
+	if err != nil {
+		h.log.Errorw("failed to delete user payload", "error", err)
+		telemetry.Error(errors.Wrap(err, "failed to delete user payload"))
+	}
+}
+
+func (h *AsyncMessageHandler) submitRequest(payload *userPayload, requestID string) (interface{}, error) {
+	httpClient := &http.Client{}
```
deliahu, line 168: nit: add this to the handler (also for batch)

pkg/dequeuer/batch_handler.go
```diff
@@ -0,0 +1,231 @@
...
+		targetURL = urls.Join(targetURL, "/on_job_complete")
+	}
+
+	httpClient := &http.Client{}
+	req, err := http.NewRequest(http.MethodPost, h.config.TargetURL, bytes.NewBuffer([]byte(messageBody)))
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
deliahu: I just confirmed this with Vishal

deliahu (approved):

pkg/dequeuer/batch_handler.go
```diff
@@ -0,0 +1,226 @@
...
+	req.Header.Set(CortexJobIDHeader, h.config.JobID)
+	response, err := h.httpClient.Do(req)
+	if err != nil {
+		return ErrorUserContainerNotReachable(err)
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return ErrorUserContainerResponseStatusCode(response.StatusCode)
```
deliahu, line 133: consider: if `isOnJobComplete == true` and if the error code is 404, this could be just logged as as info rather than returning an error

pkg/dequeuer/async_handler.go
```diff
@@ -0,0 +1,206 @@
...
+	if !strings.HasPrefix(response.Header.Get("Content-Type"), "application/json") {
+		return nil, ErrorUserContainerResponseMissingJSONHeader()
+	}
+
+	var result interface{}
+	if err = json.NewDecoder(response.Body).Decode(&result); err != nil {
+		return nil, ErrorUserContainerResponseNotJSONDecodable()
+	}
+
+	return result, nil
```
vishalbollu, line 200: Since it has been verified that response.Body is json, we can just return the bytes. That way, we don't need to marshal the result in uploadResult.

pkg/dequeuer/message_handler.go
```diff
@@ -0,0 +1,35 @@
...
+	return &handleFuncMessageHandler{HandleFunc: handleFunc}
+}
+
+type handleFuncMessageHandler struct {
+	HandleFunc func(message *sqs.Message) error
+}
+
+func (h *handleFuncMessageHandler) Handle(msg *sqs.Message) error {
+	return h.HandleFunc(msg)
+}
```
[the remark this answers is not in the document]
miguelvr: the interface makes the most sense indeed. The HandleFuncMessageHandler is for testing purposes only. It was implemented in one of the test files, but I decided to move it here

pkg/dequeuer/dequeuer.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (d *SQSDequeuer) Start(messageHandler MessageHandler) error {
+	noMessagesInPreviousIteration := false
+
+loop:
+	for {
+		select {
+		case <-d.done:
+			break loop
```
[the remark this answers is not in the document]
miguelvr: I think so, yeah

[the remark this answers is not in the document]
miguelvr: I prefer the break for aesthetic reasons: 
without the break, the `return nil` statement only happens in the select statement, while the bottom one becomes unreachable. The happy path of a function should reach the end of the function block

pkg/dequeuer/async_handler.go
miguelvr: I'm not following. I have to attempt to decode the result to json to check if it is valid json, and then I just return it to later upload it. How would we skip a step here?

vishalbollu: When it is being uploaded, I believe there is another json.marshal step that marshal's the result. If this function returned byte array, then you we can use `UploadBytesToS3` instead of `UploadJSONToS3`.

PR #2206, opened 2021-05-31
CaaS - healthcheck probes for Async/Batch/Realtime

RobertLucian:
checklist:

- [x] update examples


pkg/dequeuer/probes_test.go
```diff
@@ -0,0 +1,98 @@
...
+					Host: "localhost",
+				},
+			},
+		}, log),
+	}
+
+	require.False(t, HasTCPProbeTargetingUserPod(probes, userPodPort))
+}
+
+func TestDefaultTCPProbPresent(t *testing.T) {
```
miguelvr, line 64: nit: TestDefaultTCPProbPresent → TestDefaultTCPProbePresent

pkg/dequeuer/dequeuer.go
```diff
@@ -92,7 +95,7 @@ func (d *SQSDequeuer) ReceiveMessage() (*sqs.Message, error) {
 	return output.Messages[0], nil
 }
 
-func (d *SQSDequeuer) Start(messageHandler MessageHandler) error {
+func (d *SQSDequeuer) Start(messageHandler MessageHandler, userPodsHealthy func() bool) error {
```
miguelvr, line 98: nit: userPodsHealthy → readinessProbeFunc

pkg/lib/probe/handler.go
```diff
@@ -14,14 +14,26 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package proxy
+package probe
```
miguelvr, line 17: probe package should not be in lib, move to pkg/probe

cmd/proxy/main.go
```diff
@@ -163,6 +163,11 @@ func main() {
 		}(name, server)
 	}
 
+	probeStopper := readinessProbe.StartProbing()
```
deliahu, line 166: We can replace the proxy's probe with a simple TCP check on the /healthz handler

pkg/lib/probe/probe.go
```diff
@@ -62,13 +77,72 @@ func NewDefaultProbe(target string, logger *zap.SugaredLogger) *Probe {
...
+			TimeoutSeconds:      _defaultTimeoutSeconds,
+			PeriodSeconds:       _defaultPeriodSeconds,
+			SuccessThreshold:    _defaultSuccessThreshold,
+			FailureThreshold:    _defaultFailureThreshold,
 		},
 		logger: logger,
 	}
 }
 
-func (p *Probe) ProbeContainer() bool {
```
miguelvr, line 71: this needs to be brought back and be called by `StartProbing`, because `IsHealthy` is currently not thread safe (will have race conditions, which can't happen)

pkg/lib/probe/probe.go
```diff
@@ -62,13 +77,72 @@ func NewDefaultProbe(target string, logger *zap.SugaredLogger) *Probe {
...
+
+	ticker := time.NewTicker(time.Duration(p.PeriodSeconds) * time.Second)
+	time.AfterFunc(time.Duration(p.InitialDelaySeconds)*time.Second, func() {
+		successCount := int32(0)
+		failureCount := int32(0)
+
+		for {
+			select {
+			case <-stop:
+				return
```
miguelvr, line 101: this does not make a lot of sense to me, because we are returning a nil channel

pkg/dequeuer/http_handler.go
```diff
@@ -14,14 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package probe
+package dequeuer
 
 import "net/http"
 
-func Handler(pb *Probe) http.HandlerFunc {
+func HandlerWithConditionalFunc(isHealthy func() bool) http.HandlerFunc {
```
deliahu, line 21: `func HealthCheckHandler`

pkg/lib/probe/probe.go
```diff
@@ -62,13 +77,72 @@ func NewDefaultProbe(target string, logger *zap.SugaredLogger) *Probe {
...
+			}
+		}
+	})
+
+	return stop
+}
+
+func (p *Probe) IsHealthy() bool {
+	return p.healthy
+}
```
miguelvr, line 127: this will have race conditions, needs to be thread safe for realtime. The realtime probe should stay as it was

pkg/dequeuer/probes.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func ProbesFromFile(probesPath string, logger *zap.SugaredLogger) ([]*probe.Probe, error) {
+	fileBytes, err := files.ReadFileBytes(probesPath)
+	if err != nil {
+		return nil, err
+	}
+
+	probesMap := map[string]kcore.Probe{}
+	if err := json.Unmarshal(fileBytes, &probesMap); err != nil {
```
deliahu, line 36: nit: use `libjson` or add the stack trace to the error

pkg/lib/probe/probe.go
```diff
@@ -62,13 +77,72 @@ func NewDefaultProbe(target string, logger *zap.SugaredLogger) *Probe {
...
+	for _, probe := range probes {
+		if probe == nil {
+			continue
+		}
+		if !probe.IsHealthy() {
+			return false
+		}
+	}
+	return true
+}
```
miguelvr, line 143: same, needs to be thread safe or only used in single thread usages

miguelvr (changes requested):
there are a few assumptions that simply won't work for this PR.

the realtime probe should work as it was before, where the handler called the function to probe the container, while the probes for Async/Batch can call this function in a ticker (maybe even in the dequeuer main)

this is due to thread safety on realtime API. Async/Batch do not have this requirement, but they might have it in the future as well. 

I also wouldn't simply slap a mutex on it because that will block all goroutines in the probe, and it will occur quite often.leading to increased latencies.

pkg/lib/probe/handler.go
```diff
@@ -14,14 +14,26 @@ See the License for the specific language governing permissions and
...
+		if !pb.IsHealthy() {
+			w.WriteHeader(http.StatusInternalServerError)
+			_, _ = w.Write([]byte("unhealthy"))
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte("healthy"))
+	}
+}
```
deliahu, line 35: This can be removed

pkg/lib/probe/probe.go
```diff
@@ -26,19 +26,34 @@ import (
...
+
+	// kubeProbeUserAgentPrefix is the user agent header prefix used in k8s probes
+	// Since K8s 1.8, prober requests have
+	//   User-Agent = "kube-probe/{major-version}.{minor-version}".
+	_kubeProbeUserAgentPrefix = "kube-probe/"
+)
+
 const (
-	_defaultTimeoutSeconds = 1
+	_initialDelaySeconds     = int32(1)
```
deliahu, line 45: nit: `_defaultInitialDelaySeconds`

pkg/lib/probe/probe.go
```diff
@@ -62,13 +77,72 @@ func NewDefaultProbe(target string, logger *zap.SugaredLogger) *Probe {
...
+			SuccessThreshold:    _defaultSuccessThreshold,
+			FailureThreshold:    _defaultFailureThreshold,
 		},
 		logger: logger,
 	}
 }
 
-func (p *Probe) ProbeContainer() bool {
+func (p *Probe) StartProbing() chan bool {
+	stop := make(chan bool)
```
deliahu, line 91: nit: `stop := make(chan struct{})` (can signal via `stop <- struct{}{}`

pkg/lib/probe/probe.go
```diff
@@ -62,13 +77,72 @@ func NewDefaultProbe(target string, logger *zap.SugaredLogger) *Probe {
...
+					successCount = 0
+				}
+
+				if successCount >= p.SuccessThreshold {
+					p.healthy = true
+				} else if failureCount >= p.FailureThreshold {
+					p.healthy = false
+				}
+				p.hasRunOnce = true
+			}
```
miguelvr, line 118: this ticker does not need to be implemented by us. The apimachinery package has the implementation for the kubernetes probes

pkg/lib/probe/probe.go
```diff
@@ -62,13 +77,72 @@ func NewDefaultProbe(target string, logger *zap.SugaredLogger) *Probe {
...
+
+func (p *Probe) IsHealthy() bool {
+	return p.healthy
+}
+
+func (p *Probe) HasRunOnce() bool {
+	return p.hasRunOnce
+}
+
+func AreProbesHealthy(probes []*Probe) bool {
```
deliahu, line 133: nit: try to replace all `[]*Probe` with `[]Probe`

deliahu (approved):

PR #2211, opened 2021-06-01
Prevent nil pointer and improve logging on failed health checks

vishalbollu:


deliahu (approved):

cmd/proxy/main.go
```diff
@@ -208,13 +208,13 @@ func readinessTCPHandler(port int, logger *zap.SugaredLogger) http.HandlerFunc {
 		address := net.JoinHostPort("localhost", strconv.FormatInt(int64(port), 10))
 
 		conn, err := net.DialTimeout("tcp", address, timeout)
-		_ = conn.Close()
 		if err != nil {
-			logger.Warn(err)
+			logger.Warn(errors.Wrap(err, "health check to user provided containers failed"))
```
RobertLucian, line 212: We could replace the string with something more specific like "TCP probe to user-provided container port failed".

RobertLucian (approved):
Good catch! LGTM! Just one nit!

PR #2220, opened 2021-06-03
CaaS - batch probe fixes

RobertLucian:
checklist:


deliahu (approved):

pkg/workloads/k8s.go
```diff
@@ -189,6 +190,19 @@ func batchDequeuerProxyContainer(api spec.API, jobID, queueURL string) (kcore.Co
 				},
 			},
 		}),
+		ReadinessProbe: &kcore.Probe{
+			Handler: kcore.Handler{
+				HTTPGet: &kcore.HTTPGetAction{
+					Path: "/healthz",
+					Port: intstr.FromInt(int(consts.AdminPortInt32)),
```
miguelvr, line 197: intstr.FromInt32(adminPort)

RobertLucian: There doesn't seem to be a `FromInt32` method. It's just `FromInt`.

miguelvr: in any case it the port shouldn't be hardcoded to the const

PR #2247, opened 2021-06-11
Move statsd agent to a deployment on the operator nodegroup

### LINKED ISSUE #2041: Metrics during a batch job are lost if the underlying instances are removed
Even if a batch job were to complete successfully, it might be marked as `completed with failures` if an instance running one of the workers is removed. Instances can be removed for any number reasons, the primary one being AWS retracting spot instances. If an instance is removed, the metrics stored on the statsd agent on that node will be lost and therefore will not be available to be scrapped by prometheus.

Proposed solution:

Deploy a statsd agent dedicated for the batch jobs scheduled on the operator nodegroup. Make it available to other pods via a service. All batch workers will push their metrics to this service.

vishalbollu:
closes #2041 

---

checklist:

### DIFF
--- a/manager/manifests/prometheus-statsd-exporter.yaml
+++ b/manager/manifests/prometheus-statsd-exporter.yaml
@@ -23,17 +23,13 @@ data:
       observer_type: histogram
 
 ---
-
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
   name: prometheus-statsd-exporter
   namespace: default
 spec:
-  updateStrategy:
-    type: RollingUpdate
-    rollingUpdate:
-      maxUnavailable: 1
+  replicas: 1
   selector:
     matchLabels:
       name: prometheus-statsd-exporter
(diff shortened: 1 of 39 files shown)


manager/manifests/prometheus-statsd-exporter.yaml
```diff
@@ -23,17 +23,13 @@ data:
       observer_type: histogram
 
 ---
-
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
```
deliahu, line 27: `_cortexCPUReserve` and `_cortexMemReserve` should be updated accordingly (see the comment above their definitions). It's also worth checking that `cortex cluster info` displays the right info for capacity.

deliahu (approved):

miguelvr (approved):

miguelvr:
did you check this works (i.e. that you are able to write metrics to statsd)?

manager/manifests/prometheus-statsd-exporter.yaml
vishalbollu: Good thinking, updated.

PR #2248, opened 2021-06-15
Permissive network validations

vishalbollu:
Print a warning indicate that the quota check has bee skipped if unable to find quotas.

---

checklist:

### DIFF
--- a/pkg/lib/aws/servicequotas.go
+++ b/pkg/lib/aws/servicequotas.go
@@ -152,13 +152,47 @@ func (c *Client) VerifyInstanceQuota(instances []InstanceTypeRequests) error {
 	return nil
 }
 
+func (c *Client) ListServiceQuotas(quotaCodes []string, serviceCodes []string) (map[string]int, error) {
+	desiredQuotaCodes := strset.New(quotaCodes...)
+	quotaCodeToValueMap := map[string]int{}
+
+	for _, serviceCode := range serviceCodes {
+		err := c.ServiceQuotas().ListServiceQuotasPages(
+			&servicequotas.ListServiceQuotasInput{
+				ServiceCode: aws.String(serviceCode),
+			},
+			func(page *servicequotas.ListServiceQuotasOutput, lastPage bool) bool {
+				if page == nil {
+					return false
+				}
+				for _, quota := range page.Quotas {
+					if quota == nil || quota.QuotaCode == nil || quota.Value == nil {
+						continue
+					}
+					if desiredQuotaCodes.Has(*quota.QuotaCode) {
+						quotaCodeToValueMap[*quota.QuotaCode] = int(*quota.Value)
+					}
+				}
+				return true
+			},
+		)
+		if err != nil {
+			return nil, errors.Wrap(err, serviceCode)
+		}
+	}
+
+	return quotaCodeToValueMap, nil
+}
+
 func (c *Client) VerifyInternetGatewayQuota(internetGatewayQuota int, requiredInternetGateways int) error {
 	internetGatewaysInUse, err := c.ListInternetGateways()
 	if err != nil {
 		return err
 	}
-	if internetGatewayQuota-len(internetGatewaysInUse)-requiredInternetGateways < 0 {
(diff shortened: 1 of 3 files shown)


pkg/types/clusterconfig/network_validations.go
```diff
@@ -0,0 +1,170 @@
...
+				}
+			}
+			return true
+		},
+	)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	skippedValidations := ""
```
deliahu, line 98: nit: consider `var skippedValidations []string`, and then joining by `\n` in the deferred function

pkg/types/clusterconfig/network_validations.go
```diff
@@ -0,0 +1,170 @@
...
+	defer func() {
+		if len(skippedValidations) > 0 {
+			fmt.Println(skippedValidations)
+		}
+	}()
```
deliahu, line 103: Is it necessary to print skipped validations if there was an error that was returned? If not, this can be moved to the bottom of the function (not deferred). We can discuss this live

deliahu (approved):

pkg/types/clusterconfig/network_validations.go
```diff
@@ -0,0 +1,170 @@
...
+	awsClient *aws.Client,
+	requiredInternetGateways int,
+	natGatewayRequired bool,
+	highlyAvailableNATGateway bool,
+	requiredVPCs int,
+	availabilityZones strset.Set,
+	numNodeGroups int,
+	longestCIDRWhiteList int) error {
+
+	desiredQuotaCodes := strset.New(_elasticIPsQuotaCode, _internetGatewayQuotaCode, _natGatewayQuotaCode, _vpcQuotaCode, _securityGroupsQuotaCode, _securityGroupRulesQuotaCode)
```
RobertLucian, line 48: Can we list them line by line? I think that would be easier to read.

pkg/types/clusterconfig/network_validations.go
```diff
@@ -0,0 +1,170 @@
...
+	quotaCodeToValueMap := map[string]int{}
+
+	err := awsClient.ServiceQuotas().ListServiceQuotasPages(
```
RobertLucian, line 51: I think that creating a function like `GetServiceQuotas` that takes in the desired quota codes and the services to list from (`ec2` and `vpc`) that then returns a map of these might be simpler to follow. This function would then be called from here.

i.e.
```go
quotaCodeToValueMap := awsClient.GetServiceQuotas(
	[]string{
		_elasticIPsQuotaCode,
		_internetGatewayQuotaCode,
		_natGatewayQuotaCode,
		_vpcQuotaCode,
		_securityGroupsQuotaCode,
		_securityGroupRulesQuotaCode
	},
	[]string{
		"ec2",
		"vpc",
	}
)
```

pkg/lib/aws/servicequotas.go
```diff
@@ -159,181 +152,126 @@ func (c *Client) VerifyInstanceQuota(instances []InstanceTypeRequests) error {
...
-		},
-	)
+func (c *Client) VerifyInternetGatewayQuota(internetGatewayQuota int, requiredInternetGateways int) error {
+	internetGatewaysInUse, err := c.ListInternetGateways()
 	if err != nil {
-		return errors.WithStack(err)
+		return err
+	}
+	if internetGatewayQuota-len(internetGatewaysInUse)-requiredInternetGateways < 0 {
+		additionalQuotaRequired := len(internetGatewaysInUse) + requiredInternetGateways - internetGatewayQuota
```
RobertLucian, line 161: Since `additionalQuotaRequired` is the negative of `internetGatewayQuota-len(internetGatewaysInUse)-requiredInternetGateways`, we can just move `additionalQuotaRequired` outside of the if branch and do `if additionalQuotaRequired > 0`.

Same for `VerifyEIPQuota`, `VerifyVPCQuota` and `VerifySecurityGroupQuota` functions (where applicable).

pkg/types/clusterconfig/network_validations.go
vishalbollu: We discussed live

PR #2253, opened 2021-06-17
Prevent `"unexpected_error"` status on the Task API from surfacing

### LINKED ISSUE #2215: Task job gets set to `"unexpected_error"` status sometimes

RobertLucian:
Closes #2215.

---

checklist:

### DIFF
--- a/test/e2e/tests/conftest.py
+++ b/test/e2e/tests/conftest.py
@@ -125,8 +125,8 @@ def pytest_configure(config):
                 },
                 "task": {
                     "jobs": 10 ** 2,
-                    "concurrency": 4,
-                    "submit_timeout": 200,  # measured in seconds
+                    "concurrency": 1,
+                    "submit_timeout": 400,  # measured in seconds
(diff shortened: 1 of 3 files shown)


vishalbollu (approved):

test/e2e/tests/conftest.py
```diff
@@ -125,8 +125,8 @@ def pytest_configure(config):
                 },
                 "task": {
                     "jobs": 10 ** 2,
-                    "concurrency": 4,
-                    "submit_timeout": 200,  # measured in seconds
+                    "concurrency": 4,  # task-retrieving concurrency
```
deliahu, line 128: Would it be preferred to set this to 1 here rather than hard coding it in the other file?

RobertLucian: If we set this to 1, then the task-retrieving process will be sequential and not parallel (that should stay that way). I think we can add another field though. One for the task-submitting part and another for the task-retrieving part.

Then again, not too concerned about this.

PR #2259, opened 2021-06-17
Fix EndTime nil pointer when persisting metrics

vishalbollu:
Previously we used to get a snapshot of the batch metrics based on the job EndTime. The EndTime field would be nil if the job didn't complete due to an unhandled worker error. In such cases, controller would encounter a nil pointer error.

The controller has been updated to get a snapshot of the metrics in prometheus based on the current time rather than the EndTime so that even in situations where EndTime is nil, it tries to get the most accurate metrics.

---

checklist:

### DIFF
--- a/pkg/crds/controllers/batch/batchjob_controller_helpers.go
+++ b/pkg/crds/controllers/batch/batchjob_controller_helpers.go
@@ -672,10 +672,11 @@ func getMetrics(r *BatchJobReconciler, batchJob batch.BatchJob) (metrics.BatchMe
 		ID:      batchJob.Name,
 		APIName: batchJob.Spec.APIName,
 		Kind:    userconfig.BatchAPIKind,
-	}, batchJob.Status.EndTime.Time)
+	}, time.Now())
 	if err != nil {
 		return metrics.BatchMetrics{}, err
 	}
+
 	return jobMetrics, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/crds/controllers/batch/batchjob_controller_helpers.go
```diff
@@ -46,6 +46,7 @@ import (
 	kbatch "k8s.io/api/batch/v1"
 	kcore "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
deliahu, line 49: nit: consider calling it `kmeta` or `metav1` or something similar (it seems like we have it as `kmeta` in our other files)

miguelvr: let's stick with `kmeta` 👍

pkg/crds/controllers/batch/batchjob_controller_helpers.go
```diff
@@ -672,10 +685,11 @@ func getMetrics(r *BatchJobReconciler, batchJob batch.BatchJob) (metrics.BatchMe
 		ID:      batchJob.Name,
 		APIName: batchJob.Spec.APIName,
 		Kind:    userconfig.BatchAPIKind,
-	}, batchJob.Status.EndTime.Time)
+	}, time.Now())
```
miguelvr, line 688: shouldn't this be something like:

```go
endTime := time.Now
if batchJob.Status.EndTime != nil {
    endTime = batchJob.Status.EndTime.Time
}
```
and then we use the `endTime` variable in the call to `GetMetrics`

miguelvr (approved):
LGTM after the suggested change

pkg/crds/controllers/batch/batchjob_controller_helpers.go
vishalbollu: Initially, I was thinking `time.Now()` is generally a good option. Semantically, using endTime makes sense. Agreed.

PR #2260, opened 2021-06-18
Use cortexlabs's version kubexit to address death dependencies

### LINKED ISSUE #2217: Death dep is not honoured by kubexit if the container died before kubexit started
Kubexit only watches for changes to the graveyard folder. If a container starts and fails before the second container starts, kubexit will not terminate the second container even if it had a death dependency on the first container because the graveyard was updated before the second container even started.

To reproduce the issue, terminate the first container before the second container starts. An easy way to do this is by having an incorrect CMD or ARGs to force the first container to quit as early as possible.

Potential solution:
- update kubexit to watch for changes and then do a pass on all of the tombstones to determine if the graveyard has already been marked with death timestamps.

vishalbollu:
closes #2217 

---

checklist:


images/kubexit/Dockerfile
```diff
@@ -1 +1,14 @@
-FROM miguelvr/kubexit:0.3.2-patch
+FROM golang:1.14.1-alpine3.11 AS builder
```
deliahu, line 1: Nit: consider adding a link to where this was copied from

deliahu (approved):

images/kubexit/Dockerfile
```diff
@@ -1 +1,15 @@
...
+RUN apk update && apk upgrade && \
+    apk add bash git openssh
+
+RUN mkdir /tmp/kubexit
+RUN git clone -b v0.1.0-cortex --depth 1  https://github.com/cortexlabs/kubexit.git /tmp/kubexit
+
+WORKDIR /tmp/kubexit
+RUN CGO_ENABLED=0 GOOS=linux go build -o kubexit ./cmd/kubexit
+
+FROM alpine:3.11
```
miguelvr, line 12: nit: use the distroless base image

PR #2262, opened 2021-06-18
Add call to logger.Sync() on the test loggers and fix dequeuer test

miguelvr:
checklist:

### DIFF
--- a/pkg/dequeuer/dequeuer_test.go
+++ b/pkg/dequeuer/dequeuer_test.go
@@ -171,12 +171,15 @@ func TestSQSDequeuer_ReceiveMessage(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	logger := newLogger(t)
+	defer func() { _ = logger.Sync() }()
+
 	dq, err := NewSQSDequeuer(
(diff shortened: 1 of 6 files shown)


pkg/dequeuer/dequeuer_test.go
```diff
@@ -351,7 +351,7 @@ func TestSQSDequeuer_Start_HandlerError(t *testing.T) {
 
 	dq.waitTimeSeconds = aws.Int64(0)
 	dq.notFoundSleepTime = 0
-	dq.renewalPeriod = time.Second
+	dq.renewalPeriod = 500 * time.Millisecond
```
RobertLucian, line 354: Why the 500ms renewal? And should we do this the same for the one in `TestSQSDequeuer_StartMessageRenewer`?

miguelvr: it doesn't matter that much, I just wanted to make sure it was fast enough

RobertLucian: Cool!

RobertLucian (approved):

PR #2265, opened 2021-06-19
Support refresh CLI command for Async APIs

### LINKED ISSUE #2237: Cortex refresh for AsyncAPI
`cortex refresh` is already supported for RealtimeAPIs to force a rolling update. It is useful when images are set to latest tag and new versions of the images need to be pulled.

Revisit deployment ID.

deliahu:
closes #2237

---

checklist:


cmd/operator/main.go
```diff
@@ -65,9 +65,10 @@ func main() {
 
 	for i := range deployments {
 		deployment := deployments[i]
-		apiKind := deployment.Labels["apiKind"]
-		if userconfig.KindFromString(apiKind) == userconfig.RealtimeAPIKind ||
-			userconfig.KindFromString(apiKind) == userconfig.AsyncAPIKind {
+		apiKind := userconfig.KindFromString(deployment.Labels["apiKind"])
+		if apiKind == userconfig.RealtimeAPIKind ||
+			(apiKind == userconfig.AsyncAPIKind && deployment.Labels["cortex.dev/async"] != "gateway") {
```
deliahu, line 70: Note to reviewer: I think this should have been checking `!= "gateway"` already (not related to this PR), but not 100% sure. Lmk what you think.

pkg/operator/resources/asyncapi/status.go
```diff
@@ -297,6 +297,7 @@ func addPodToReplicaCounts(pod *kcore.Pod, deployment *kapps.Deployment, counts
 }
 
 func isPodSpecLatest(deployment *kapps.Deployment, pod *kcore.Pod) bool {
```
deliahu, line 299: note to reviewer: this is being called with the api deployment and the gateway deployment. I believe the gateway deployment can never be out of date, so therefore it's ok that these labels don't exist on the gateway deployment (which means this will always return `true`).

cmd/operator/main.go
vishalbollu: I believe this is correct. The gateway is currently scaled by HPA. There is an if statement in UpdateMetricsCron and UpdateAutoscalerCron that ignores the gateway deployments. Perhaps `UpdateAutoscalerCron` and `UpdateMetricsCron` should be renamed to `UpdateAPIAutoscalerCron` and  `UpdateAPIMetricsCron` respectively to make it clear. If it were renamed, I would be comfortable with dropping the if statements in both functions

pkg/operator/resources/asyncapi/status.go
vishalbollu: That makes sense.

pkg/operator/resources/asyncapi/queue.go
```diff
@@ -52,8 +53,10 @@ func createFIFOQueue(apiName string, deploymentID string, tags map[string]string
 	return *output.QueueUrl, nil
 }
 
-func apiQueueName(apiName string, deploymentID string) string {
-	return config.ClusterConfig.SQSNamePrefix() + apiName + clusterconfig.SQSQueueDelimiter + deploymentID + ".fifo"
+func apiQueueName(apiName string, createdTime int64) string {
```
vishalbollu, line 56: Add a comment explaining why the api queue name should include createdTime

pkg/types/spec/api.go
```diff
@@ -41,12 +41,14 @@ type API struct {
 
 	Key string `json:"key"`
 
+	CreatedTime  int64  `json:"created_time"`
```
vishalbollu, line 44: I would call this field InitialDeploymentTime or something to indicate that this is not just when this particular spec was created.

vishalbollu (approved):
LGTM after adding a few comments and some renaming.

pkg/operator/resources/asyncapi/api.go
```diff
@@ -64,25 +64,31 @@ func getGatewayK8sName(apiName string) string {
...
 	if err != nil {
 		return nil, "", err
 	}
 
-	deployID := deploymentID()
-	if prevK8sResources.apiDeployment != nil && prevK8sResources.apiDeployment.Labels["deploymentID"] != "" {
-		deployID = prevK8sResources.apiDeployment.Labels["deploymentID"]
+	createdTime := time.Now().UnixNano()
+	deploymentID := generateDeploymentID()
+	if prevK8sResources.gatewayVirtualService != nil && prevK8sResources.gatewayVirtualService.Labels["createdTime"] != "" {
```
miguelvr, line 79: this label is unnecessary. All kubernetes resources have a `creationTimestamp` field

vishalbollu: The creationTimestamp is required to initialize the SQS queue. The SQS queue is created before the k8s resources are applied. Once there is a CRD for the async api, we wouldn't need to have a separate field like `creationTimestamp`.

PR #2298, opened 2021-06-29
Scale to zero

### LINKED ISSUE #445: Enable Replica Auto Scaling down to zero
It would be really useful especially for smaller applications to be able to scale GPU's down to 0 when there is no traffic.

## Possible approach

* To trigger scaling 1 -> 0, check CloudWatch metrics for no requests for a certain amount of time (user-configurable?).
* Scale 1 -> 0 by setting `deployment.spec.replicas` to 0.
* When scaling 1 -> 0, also update the Istio Virtual Service to route requests to that API to a new deployment running in the Cortex node (or use the existing operator)
* 0 -> 1 scaling is triggered when a request comes in to that service
* Scale 0 -> 1 by setting deployment.spec.replicas to 0
* Either the service holds onto the request until the pod is ready, forwards it, and replies with the response, or responds immediately with a message saying e.g. "0 -> 1 scaling has been triggered, please try again in a few minutes"

miguelvr:
closes #445 

---

checklist:


pkg/autoscaler/realtime_scaler.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/cortexlabs/cortex/pkg/workloads"
+	promv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+	"go.uber.org/zap"
+	kapps "k8s.io/api/apps/v1"
+	kmeta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	_waitForReadyReplicasTimeout = 10 * time.Minute
```
deliahu, line 37: Consider making this 15 or 20 minutes

pkg/types/clusterconfig/cluster_config.go
```diff
@@ -1621,6 +1637,15 @@ func (cc *CoreConfig) TelemetryEvent() map[string]interface{} {
...
 	}
+	if !strings.HasPrefix(cc.ImageActivator, "cortexlabs/") {
+		event["image_activator._is_custom"] = true
+	}
+	if !strings.HasPrefix(cc.ImageAutoscaler, "cortexlabs/") {
+		event["image_activator._is_custom"] = true
+	}
+	if !strings.HasPrefix(cc.ImageProxy, "cortexlabs/") {
+		event["image_proxy._is_custom"] = true
+	}
```
deliahu, line 1648: This is duplicated

pkg/types/spec/validations.go
```diff
@@ -473,7 +473,8 @@ func computeValidation() *cr.StructFieldValidation {
 
 func autoscalingValidation(kind userconfig.Kind) *cr.StructFieldValidation {
```
deliahu, line 474: This func can be converted to a var, and the `minReplicas` variable can be replaced by `0`

pkg/activator/activator.go
```diff
@@ -0,0 +1,244 @@
...
+
+	act, err := a.getOrCreateAPIActivator(ctx, apiName)
+	if err != nil {
+		return err
+	}
+
+	tracker := a.getOrCreateDeploymentTracker(apiName)
+
+	if act.inFlight() == 0 {
+		go a.awakeAPI(apiName)
```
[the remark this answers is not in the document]
miguelvr: the queue would have different values, and the awaken call is a no-op if it was already called within the stabilization period

pkg/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,294 @@
...
+		zap.String("apiKind", api.Kind.String()),
+	)
+
+	log.Infof("autoscaling awake event")
+	if err := scaler.Scale(api.Name, 1); err != nil {
+		return err
+	}
+
+	a.Lock()
+	a.awakenMap[api.Name] = time.Now()
```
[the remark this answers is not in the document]
miguelvr: no, because it should only be registered if the scale call succeeds

pkg/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,294 @@
...
+	if autoscalerCron, ok := a.crons[api.Name]; ok {
+		autoscalerCron.Cancel()
+		delete(a.crons, api.Name)
+	}
+
+	a.Lock()
+	delete(a.awakenMap, api.Name)
+	a.Unlock()
+
+	log.Info("autoscaler stop")
```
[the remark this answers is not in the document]
miguelvr: it's already included

cmd/autoscaler/main.go
```diff
@@ -0,0 +1,236 @@
...
+	"go.uber.org/zap"
+	istioclient "istio.io/client-go/pkg/clientset/versioned"
+	istioinformers "istio.io/client-go/pkg/informers/externalversions"
+	"k8s.io/apimachinery/pkg/api/meta"
+	kmeta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+)
+
+func main() {
```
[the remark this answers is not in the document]
miguelvr: yes

pkg/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,295 @@
...
+	awakenStabilizationPeriod time.Duration
+}
+
+func New(logger *zap.SugaredLogger) *Autoscaler {
+	return &Autoscaler{
+		logger:                    logger,
+		crons:                     make(map[string]cron.Cron),
+		scalers:                   make(map[userconfig.Kind]Scaler),
+		lastAwakenTimestamp:       make(map[string]time.Time),
+		awakenStabilizationPeriod: _awakenStabilizationPeriod,
```
deliahu, line 55: I think both of these will be able to be removed

pkg/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+}
+
+func (a *Autoscaler) AddScaler(scaler Scaler, kind userconfig.Kind) {
+	a.scalers[kind] = scaler
+}
+
+func (a *Autoscaler) Awaken(api userconfig.Resource) error {
+	a.Lock()
```
deliahu, line 64: Remove the use of awaken timestamp, instead lock the function, get the deployment, and check the current value of requested replicas

pkg/autoscaler/autoscaler.go
```diff
@@ -0,0 +1,295 @@
...
+		// awaken state: was scaled from zero
+		// This needs to be protected by a Mutex because an Awaken call will also modify it
+		a.Lock()
+		awoke := time.Since(a.lastAwakenTimestamp[api.Name]) <= _awakenStabilizationPeriod
+		a.Unlock()
+
+		// do now allow downscale bellow 1 if API was awoke with the awaken stabilization period
+		if awoke {
+			request = libmath.MaxInt32(currentReplicas, 1)
+		}
```
deliahu, line 261: Let's delete this for now

pkg/activator/activator.go
```diff
@@ -0,0 +1,272 @@
...
+	maxQueueLength, maxConcurrency, err := concurrencyFromAnnotations(vs.Annotations)
+	if err != nil {
+		return nil, err
+	}
+
+	apiAct := newAPIActivator(maxQueueLength, maxConcurrency)
+
+	a.activatorsMux.Lock()
+	a.apiActivators[apiName] = apiAct
+	a.activatorsMux.Unlock()
```
deliahu, line 133: It might be better to place the lock around this entire function rather than just the beginning and end, to prevent multiple threads from sneaking in (like we discussed for `(a *Autoscaler) Awaken()`

pkg/activator/activator.go
```diff
@@ -0,0 +1,272 @@
...
+	if apiMetadata.apiKind != userconfig.RealtimeAPIKind {
+		return
+	}
+
+	apiName := apiMetadata.apiName
+
+	a.logger.Debugw("adding new api activator", zap.String("apiName", apiName))
+
+	a.activatorsMux.Lock()
+	a.apiActivators[apiName] = newAPIActivator(apiMetadata.maxQueueLength, apiMetadata.maxConcurrency)
```
deliahu, line 171: Do you think it would be safer to check if an activator already exists in the `a.apiActivators` map before creating a new one?

pkg/activator/helpers.go
```diff
@@ -0,0 +1,82 @@
...
+		annotations:    annotations,
+		maxConcurrency: maxConcurrency,
+		maxQueueLength: maxQueueLength,
+	}, nil
+}
+
+func concurrencyFromAnnotations(annotations map[string]string) (int, int, error) {
+	maxQueueLength, err := strconv.Atoi(annotations[userconfig.MaxQueueLengthAnnotationKey])
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to parse \"%s\" annotation", userconfig.MaxQueueLengthAnnotationKey) // FIXME: proper error
```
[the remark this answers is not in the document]
deliahu: Marking this as resolved, since I the k8s meta libraries didn't line up

pkg/activator/activator.go
```diff
@@ -168,7 +164,9 @@ func (a *activator) addAPI(obj interface{}) {
 	a.logger.Debugw("adding new api activator", zap.String("apiName", apiName))
```
deliahu, line 164: Should we use the `a.activatorsMux` lock for this entire function?

pkg/activator/activator.go
```diff
@@ -168,7 +164,9 @@ func (a *activator) addAPI(obj interface{}) {
...
 
 	a.activatorsMux.Lock()
-	a.apiActivators[apiName] = newAPIActivator(apiMetadata.maxQueueLength, apiMetadata.maxConcurrency)
+	if a.apiActivators[apiName] != nil {
```
deliahu, line 167: Should this be `== nil`?

Also is it possible to check this (and return) at the top of the function? e.g.

```go
if a.apiActivators[apiName] != nil {
    return
}
```

pkg/activator/activator.go
```diff
@@ -267,6 +265,6 @@ func (a *activator) removeReadinessTracker(obj interface{}) {
 	}
 
 	a.trackersMux.Lock()
+	defer a.trackersMux.Unlock()
```
deliahu, line 268: If you'd like, there are a few other places in this file where we can use this pattern (search for `.Lock()`)

pkg/autoscaler/autoscaler.go
```diff
@@ -240,6 +238,17 @@ func (a *Autoscaler) autoscaleFn(api userconfig.Resource) (func() error, error)
...
+			// awaken state: was scaled from zero
+			// This needs to be protected by a Mutex because an Awaken call will also modify it
+			a.Lock()
+			lastAwakenTimestamp := a.lastAwakenTimestamp[api.Name]
+
+			// Make sure we don't scale below zero if API was recently awaken
+			if time.Since(lastAwakenTimestamp) < autoscalingSpec.DownscaleStabilizationPeriod {
+				request = libmath.MaxInt32(request, 1)
+			}
+			a.Unlock()
```
deliahu, line 251: Do you know why this is necessary, i.e. that the lines of code above which consider `DownscaleStabilizationPeriod` are not sufficient?

pkg/activator/activator.go
miguelvr: not necessary, it is better as it is, in terms of performance

pkg/autoscaler/autoscaler.go
miguelvr: it is not sufficient. I have checked it with the unit tests

pkg/activator/handler.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func NewHandler(act Activator, logger *zap.SugaredLogger) *Handler {
+	return &Handler{
+		activator: act,
+		logger:    logger,
+	}
+}
+
+func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
miguelvr: done!

miguelvr:
@deliahu this should be good to go now

deliahu (approved):

PR #2299, opened 2021-06-29
Increase cluster autoscaler limits

### LINKED ISSUE #2297: Raise cluster autoscaler resource limits
Cluster autoscaler is running out of memory at 200+ nodes. Raise the limits of cluster-autoscale to allow the cluster-autoscaler to use more memory than requested if available. It might get evicted from the node but the operator nodegroup will scale up and will be provisioned on the new instance.

miguelvr:
closes #2297 

---

checklist:


deliahu:
This is addressed by #2307

PR #2305, opened 2021-07-02
Allow ssl and ip address whitelisting fields to be updated on configure

### LINKED ISSUE #2107: Make the SSL certificate ARN configurable on an existing cluster
#### Implementation

Can be done just by editing the API service's respective annotation: `service.beta.kubernetes.io/aws-load-balancer-ssl-cert`.
To make it simpler to implement, we could only allow this to be changed if the user has initially provided an ACM.

As for where this could fit, we could add another generic CLI command that would only allow changing the SSL ARN for now.
The command could be `cortex cluster update <field-to-update> <value-to-update-it-to>`.

#### Context

Requested by [Oldřich Šafář](https://cortex-dot-dev.slack.com/archives/C01KHQRUWNT/p1618931053300900) from the community Slack.

#### Temporary workaround

Install [kubectl](https://docs.cortex.dev/clusters/advanced/kubectl), run `kubectl edit service ingressgateway-apis -n istio-system`, and update the `service.beta.kubernetes.io/aws-load-balancer-ssl-cert annotation` accordingly.

#### Notes

* Update the networking docs: the SSL certificate section of the custom domain guide can be moved into the HTTPS guide.

#### Relevant code

`istio.yaml.j2`

```yaml
{% if config.get('ssl_certificate_arn', '') != '' %}
service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "{{ config['ssl_certificate_arn'] }}"
{% endif %}
```

`apis.yaml.j2`

```yaml
{% if config.get('ssl_certificate_arn', '') == '' %}
- port:
    number: 443
    name: https
    protocol: HTTPS
  hosts:
    - "*"
  tls:
    mode: SIMPLE
    serverCertificate: /etc/istio/customgateway-certs/tls.crt
    privateKey: /etc/istio/customgateway-certs/tls.key
{% else %}
- port:
    number: 443
    name: https
    protocol: HTTP
  hosts:
    - "*"
{% endif %}
```

`cluster_config.go`

```go
if cc.SSLCertificateARN != nil {
	exists, err := awsClient.DoesCertificateExist(*cc.SSLCertificateARN)
	if err != nil {
		return errors.Wrap(err, SSLCertificateARNKey)
	}

	if !exists {
		return errors.Wrap(ErrorSSLCertificateARNNotFound(*cc.SSLCertificateARN, cc.Region), SSLCertificateARNKey)
	}
}
```

vishalbollu:
closes #2107 

---

checklist:

### DIFF
--- a/manager/install.sh
+++ b/manager/install.sh
@@ -90,6 +90,8 @@ function cluster_configure() {
   add_nodegroups
   remove_nodegroups
 
+  update_networking
+
   echo -n "￮ updating cluster configuration "
   setup_configmap
   echo "✓"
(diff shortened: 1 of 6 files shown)


manager/install.sh
```diff
@@ -365,6 +367,53 @@ function setup_istio() {
...
+  new_api_whitelist_ip_address=$(cat $CORTEX_CLUSTER_CONFIG_FILE | yq -r -c ".api_load_balancer_cidr_white_list")
+  new_operator_whitelist_ip_address=$(cat $CORTEX_CLUSTER_CONFIG_FILE | yq -r -c ".operator_load_balancer_cidr_white_list")
+
+  if [ "$prev_ssl_certificate_arn" = "$new_ssl_certificate_arn" ] && [ "$prev_api_whitelist_ip_address" = "$new_api_whitelist_ip_address" ] && [ "$prev_operator_whitelist_ip_address" = "$new_operator_whitelist_ip_address" ] ; then
+      return
+  fi
+
+  echo -n "￮ updating networking configuration "
+
+  https_index=$(kubectl get svc ingressgateway-apis -n=istio-system -o json  | jq '.spec.ports | map(.name == "https") | index(true)')
```
deliahu, line 395: nit: this line is duplicated below

manager/install.sh
```diff
@@ -365,6 +367,53 @@ function setup_istio() {
...
+
+  if [ "$new_ssl_certificate_arn" != "$prev_ssl_certificate_arn" ] ; then
+      # there is a bug where changing the certificate annotation will not cause the HTTPS listener in the NLB to update
+      # the current workaround is to delete the HTTPS listener and have it recreated with istioctl
+      if [ "$prev_ssl_certificate_arn" != "" ] ; then
+        kubectl patch svc ingressgateway-apis -n=istio-system --type=json -p="[{'op': 'remove', 'path': '/metadata/annotations/service.beta.kubernetes.io~1aws-load-balancer-ssl-cert'}]" >/dev/null
```
deliahu, line 401: reminder to ask you about the `~1` :)

manager/install.sh
```diff
@@ -365,6 +367,53 @@ function setup_istio() {
...
+      fi
+      https_index=$(kubectl get svc ingressgateway-apis -n=istio-system -o json  | jq '.spec.ports | map(.name == "https") | index(true)')
+      if [ "$https_index" != "null" ] ; then
+        kubectl patch svc ingressgateway-apis -n=istio-system --type=json -p="[{'op': 'remove', 'path': '/spec/ports/$https_index'}]" >/dev/null
```
deliahu, line 405: Should these be combined into a single patch command, or does it not matter?

pkg/types/clusterconfig/cluster_config.go
```diff
@@ -1029,43 +1031,54 @@ func (cc *Config) validate(awsClient *aws.Client) error {
...
 	if h1 != h2 {
-		return ErrorConfigCannotBeChangedOnConfigure()
+		return nil, ErrorConfigCannotBeChangedOnConfigure()
 	}
 
-	return nil
+	return fieldsToUpdate, nil
+}
+
+func emptyUpdatableFields(clusterConfig *Config) {
```
deliahu, line 1077: nit: `emptyUpdatableFields` -> `clearUpdatableFields`

manager/install.sh
vishalbollu: They should be in separate patch commands because the condition for running the dropping them is different.

manager/install.sh
vishalbollu: I recently learnt this how to escape `/` in json pointers

manager/install.sh
vishalbollu: removed

deliahu (approved):

PR #2309, opened 2021-07-02
Fix cluster-autoscaler-priority expander

RobertLucian:
This is affecting the priority of the node groups. Instead of them being based on the `priority` field inside the node groups section, they were based on the order of the node groups.

---

checklist:

### DIFF
--- a/manager/manifests/cluster-autoscaler.yaml.j2
+++ b/manager/manifests/cluster-autoscaler.yaml.j2
@@ -137,13 +137,25 @@ metadata:
   namespace: kube-system
 data:
   priorities: |-
-  {% for ng in config['node_groups']|reverse %}
-    {{ (loop.index0+1) * 10 }}:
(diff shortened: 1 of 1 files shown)


manager/manifests/cluster-autoscaler.yaml.j2
```diff
@@ -137,13 +137,25 @@ metadata:
...
+      {% if ng['priority'] == p %}
+        {%- if found.update({'priority':True}) %} {%- endif %}
+      {% endif %}
+    {% endfor %}
+    {% if found['priority'] %}
+    {{ 101-p }}:
+    {% endif %}
+    {% for ng in config['node_groups'] %}
+      {% if ng['priority'] == p %}
+      {% if ng['spot'] %}
```
deliahu, line 152: nit: indent this

deliahu (approved):

PR #2324, opened 2021-07-08
Make --print-config display yaml that is consumable by cortex configure

vishalbollu:
Refactor cluster configuration so to make it easy to allow --print-config to print a valid cluster configuration file that can be consumed by other cortex commands.

---

checklist:


deliahu (approved):

RobertLucian (approved):
Elegant implementation!

pkg/crds/controllers/batch/batchjob_controller_helpers.go
```diff
@@ -529,7 +529,7 @@ func (r *BatchJobReconciler) updateStatus(ctx context.Context, batchJob *batch.B
 func (r *BatchJobReconciler) checkWorkersOOM(ctx context.Context, batchJob *batch.BatchJob) (bool, error) {
 	workerJobPods := kcore.PodList{}
 	if err := r.List(ctx, &workerJobPods,
-		client.InNamespace(r.ClusterConfig.Namespace),
+		client.InNamespace(consts.DefaultNamespace),
```
miguelvr, line 532: isn't the cluster config more reliable?

miguelvr: oh, I see you removed the namespace from the cluster config, nvm!

PR #2325, opened 2021-07-08
Error when a nodegroup name with a dash - suffix is provided

vishalbollu:
Nodegroup names ending with `-` spin up successfully but instances in these nodegroups never join the cluster. It also causes errors when trying to deleting a nodegroup with a `-` suffix.

---

checklist:

### DIFF
--- a/pkg/lib/configreader/string.go
+++ b/pkg/lib/configreader/string.go
@@ -43,6 +43,7 @@ type StringValidation struct {
 	Prefix                               string
 	Suffix                               string
 	InvalidPrefixes                      []string
+	InvalidSuffixes                      []string
 	AllowedPrefixes                      []string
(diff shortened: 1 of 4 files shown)


pkg/lib/configreader/string.go
```diff
@@ -43,6 +43,7 @@ type StringValidation struct {
 	Prefix                               string
 	Suffix                               string
 	InvalidPrefixes                      []string
+	InvalidSuffixes                      []string
```
deliahu, line 46: nit: also add `InvalidSuffixes` to `pkg/lib/configreader/string_ptr.go` (you can search `InvalidPrefixes` and pattern-match)

deliahu (approved):

PR #2327, opened 2021-07-09
Improve error reporting

miguelvr:
checklist:


cmd/dequeuer/main.go
```diff
@@ -242,21 +242,12 @@ func main() {
 }
 
 func exit(log *zap.SugaredLogger, err error, wrapStrs ...string) {
```
deliahu, line 244: Is it potentially misleading that this function can be called without the program then exiting (since it's called `exit()`)? This would happen if `err` is nil or it it's not nil, but is not printed.

We could consider renaming this function, or exiting whether or not an error is passed in (e.g. exit with code 0 if no error is passed in, exit with code 1 if it's a no-print error)

miguelvr: It exits with log.falal

deliahu: Yes, but only `if err != nil && !errors.IsNoPrint(err)`

miguelvr: fixed

cmd/autoscaler/main.go
```diff
@@ -282,6 +282,10 @@ func informerFilter(listOptions *kmeta.ListOptions) {
 }
 
 func exit(log *zap.SugaredLogger, err error, wrapStrs ...string) {
+	if err == nil {
+		os.Exit(0)
+	}
+
 	for _, str := range wrapStrs {
 		err = errors.Wrap(err, str)
 	}
```
deliahu, line 291: This function will still not exit if `errors.IsNoPrint(err)` is true. Also, the `if err != nil` check below can be removed (since we check it above)

miguelvr: solved

cmd/activator/main.go
```diff
@@ -183,17 +183,17 @@ func informerFilter(listOptions *kmeta.ListOptions) {
...
 	}
 
-	if err != nil && !errors.IsNoTelemetry(err) {
-		telemetry.Error(err)
+	for _, str := range wrapStrs {
+		err = errors.Wrap(err, str)
 	}
 
+	telemetry.Error(err)
 	if err != nil && !errors.IsNoPrint(err) {
```
deliahu, line 195: Should we remove `err != nil &&` since it was checked above?

deliahu (approved):

PR #2329, opened 2021-07-09
Forward headers to AsyncAPI

### LINKED ISSUE #2296: Forward gateway request headers to AsyncAPI
### Description

User headers are not being propagated to the AsyncAPI workload from the Gateway.

### Motivation

Headers are useful in various scenarios such as authorization. We should propagate user provided headers

### Context

As requested by `Afeef Janjua` on Slack.

miguelvr:
closes #2296 

---

checklist:


pkg/async-gateway/service.go
```diff
@@ -52,25 +55,42 @@ func NewService(clusterUID, apiName string, queue Queue, storage Storage, logger
 }
 
 // CreateWorkload enqueues an async workload request and uploads the request payload to S3
-func (s *service) CreateWorkload(id string, payload io.Reader, contentType string) (string, error) {
+func (s *service) CreateWorkload(id string, payload io.Reader, headers http.Header) (string, error) {
+	contentType := headers.Get("Content-Type")
+	if contentType == "" {
+		return "", errors.ErrorUnexpected("missing content-type in headers")
+	}
+	headers.Del("Content-Type")
```
deliahu, line 63: Since we pass all the headers through now, should we consider preserving the "Content-Type" header here so it gets uploaded with all the others, and then not setting it explicitly in the dequeuer (i.e. deleting line 189 in `pkg/dequeuer/async_handler.go`)? Theoretically we might not even need to set the content type on the s3 object, but I think that is still nice, and there's no reason not to do it.

deliahu: Also, perhaps now we could allow the user to not specify the content type header? We can discuss this on Monday

pkg/async-gateway/service.go
```diff
@@ -52,25 +55,42 @@ func NewService(clusterUID, apiName string, queue Queue, storage Storage, logger
 }
 
 // CreateWorkload enqueues an async workload request and uploads the request payload to S3
-func (s *service) CreateWorkload(id string, payload io.Reader, contentType string) (string, error) {
+func (s *service) CreateWorkload(id string, payload io.Reader, headers http.Header) (string, error) {
+	contentType := headers.Get("Content-Type")
+	if contentType == "" {
```
RobertLucian, line 60: Since the `Content-Type` header is checked twice for its presence, what do you think of removing it from the calling function?

pkg/async-gateway/service.go
RobertLucian: I agree with this. I think making sure that the `Content-Type` is present in the headers and removing it from the `userPayload` struct so that the payload is only meant for the _payload_ makes sense to me.

---

Not asking the users to specify the content type can lead to a less opinionated implementation. Also, from [RFC 7231, section 3.1.1.5](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5)
```
   A sender that generates a message containing a payload body SHOULD
   generate a Content-Type header field in that message unless the
   intended media type of the enclosed representation is unknown to the
   sender.  If a Content-Type header field is not present, the recipient
   MAY either assume a media type of "application/octet-stream"
   ([RFC2046], Section 4.5.1) or examine the data to determine its type.
```
What that says is that if the sender doesn't know what the content type is, then it's okay if it doesn't add it. So we shouldn't enforce it.

pkg/async-gateway/service.go
RobertLucian: *Assuming the `Content-Type` will still be around*

pkg/dequeuer/async_handler.go
```diff
@@ -170,12 +179,13 @@ func (h *AsyncMessageHandler) deletePayload(requestID string) {
...
 
-func (h *AsyncMessageHandler) submitRequest(payload *userPayload, requestID string) (interface{}, error) {
+func (h *AsyncMessageHandler) submitRequest(payload *userPayload, headers http.Header, requestID string) (interface{}, error) {
 	req, err := http.NewRequest(http.MethodPost, h.config.TargetURL, payload.Body)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 
+	req.Header = headers
 	req.Header.Set("Content-Type", payload.ContentType)
```
RobertLucian, line 189: This one would be dropped in either (when relying on the headers file alone or if we no longer enforce it) case I think.

deliahu (approved):

pkg/dequeuer/async_handler.go
```diff
@@ -106,7 +106,16 @@ func (h *AsyncMessageHandler) handleMessage(requestID string) error {
...
+	headers, err := h.getHeaders(requestID)
+	if err != nil {
+		updateStatusErr := h.updateStatus(requestID, async.StatusFailed)
+		if updateStatusErr != nil {
+			h.log.Errorw("failed to update status after failure to get headers", "id", requestID, "error", updateStatusErr)
+		}
+		return errors.Wrap(err, "failed to get payload")
+	}
+
+	result, err := h.submitRequest(payload, headers, requestID)
```
RobertLucian, line 118: Can't comment on `func (h *AsyncMessageHandler) getPayload(requestID string) (*userPayload, error)` line, so I'm adding this comment here. I think we should remove the `ContentType` field from the `userPayload` struct and not set it to a default value if we don't find it.

We shouldn't make assumptions about what the content type is if it's not set.

What do you say?

deliahu: It seems to me like `payload.ContentType` is not being used currently, so I think we can just delete it from the struct and it won't have an impact

PR #2335, opened 2021-07-13
Add tests to check if autoscaler respects min and max replicas

miguelvr:
checklist:

### DIFF
--- a/pkg/autoscaler/autoscaler_test.go
+++ b/pkg/autoscaler/autoscaler_test.go
@@ -42,6 +42,7 @@ func newLogger(t *testing.T) *zap.SugaredLogger {
 }
 
 func TestAutoscaler_Awake(t *testing.T) {
+	t.Parallel()
 	log := newLogger(t)
 
 	mux := sync.RWMutex{}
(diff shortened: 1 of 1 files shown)


pkg/autoscaler/autoscaler_test.go
```diff
@@ -113,3 +114,141 @@ func TestAutoscaler_Awake(t *testing.T) {
...
+				require.NoError(t, err)
+			}
+		}
+	}()
+
+	require.Never(t, func() bool {
+		mux.RLock()
+		defer mux.RUnlock()
+		return latestRequest < minReplicas
+	}, 3*time.Second, time.Second)
```
RobertLucian, line 184: What do you say about running this for longer?

pkg/autoscaler/autoscaler_test.go
```diff
@@ -113,3 +114,141 @@ func TestAutoscaler_Awake(t *testing.T) {
...
+	mux := sync.RWMutex{}
+	var latestRequest int32
+
+	minReplicas := int32(5)
+	maxReplicas := int32(10)
+
+	scalerMock := &ScalerFunc{
+		ScaleFunc: func(apiName string, request int32) error {
+			mux.Lock()
+			defer mux.Unlock()
```
RobertLucian, line 131: Robert: is this required for the test?

pkg/autoscaler/autoscaler_test.go
```diff
@@ -113,3 +114,141 @@ func TestAutoscaler_Awake(t *testing.T) {
...
+		GetInFlightRequestsFunc: func(apiName string, window time.Duration) (*float64, error) {
+			return pointer.Float64(0), nil
+		},
+		GetAutoscalingSpecFunc: func(apiName string) (*userconfig.Autoscaling, error) {
+			return &userconfig.Autoscaling{
+				MinReplicas:    minReplicas,
+				MaxReplicas:    maxReplicas,
+				InitReplicas:   minReplicas,
+				TargetInFlight: pointer.Float64(1),
+				Window:         time.Second,
```
RobertLucian, line 145: Is it okay if the stabilization periods are set to zero? Same for downscale and upscale factors? This would also apply for the other test.

pkg/autoscaler/autoscaler_test.go
```diff
@@ -113,3 +114,141 @@ func TestAutoscaler_Awake(t *testing.T) {
...
+			}, nil
+		},
+		CurrentReplicasFunc: func(apiName string) (int32, error) {
+			return minReplicas + 1, nil
```
RobertLucian, line 149: Where does the `+1` come from?

pkg/autoscaler/autoscaler_test.go
miguelvr: it doesn't really matter, because the stabilization periods are zero. we want to keep unit tests as short as possible

pkg/autoscaler/autoscaler_test.go
miguelvr: yes because there are 2 goroutines accessing the variable

pkg/autoscaler/autoscaler_test.go
miguelvr: yes, for the stabilization periods. Not sure about the downscale and upscale factors

pkg/autoscaler/autoscaler_test.go
miguelvr: This is to make sure that the value is different from the minReplicas value. This way a scale operation will be triggered

pkg/autoscaler/autoscaler_test.go
miguelvr: reduced the interval each autoscaleFn runs to have it run longer instead

pkg/autoscaler/autoscaler_test.go
miguelvr: updated the scale factors to the default ones

RobertLucian (approved):

PR #2342, opened 2021-07-14
Fix IstioD causing the Cortex cluster to no longer be reachable

### LINKED ISSUE #2332: IstioD is using too many resources when lots of pods are being added/churned
#### Description

Find a way to disable the monitoring of k8s resources so that istiod stops using so many resources.

#### Consequence

When istiod uses too much memory, it triggers an OOM event on the other pods collocated on the same instance such as the ingress gateways, the cluster autoscaler, the operator, etc. Or it can make k8s evict the other pods which is equally bad. This can render the cluster unusable.

RobertLucian:
Fixes #2332

---

checklist:


miguelvr (commented):
AFAIK there are a more places referring to prometheus in the default namespace

From the top of my head, look for the autoscaler, activator, cluster health commands. But it is better to do a wide search for anything using the word "default"

manager/install.sh
```diff
@@ -35,6 +35,10 @@ function main() {
 function cluster_up() {
   create_eks
 
+  echo -n "￮ creating namespaces "
+  setup_namespaces
```
deliahu, line 39: nit: move this right before `setup_configmap` and remove the `echo`s

manager/manifests/istio.yaml.j2
```diff
@@ -26,7 +30,23 @@ spec:
         resources:
           requests:
             cpu: 100m  # default is 500m
-            memory: 200Mi  # default is 2048Mi == 2Gi
+            memory: 700Mi  # default is 2048Mi == 2Gi
+        hpaSpec:
+          minReplicas: 1
+          maxReplicas: 5  # edit autoscaleEnabled in values if increasing this
```
deliahu, line 36: nit: remove this comment (and the others like it)

deliahu (approved):

pkg/health/health.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/cortexlabs/cortex/pkg/lib/parallel"
 	"github.com/cortexlabs/cortex/pkg/types/clusterconfig"
 	kapps "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
```
miguelvr, line 33: kcore

miguelvr (approved):

PR #2347, opened 2021-07-15
Ensure cluster autoscaler availability

### LINKED ISSUE #2346: Ensure the cluster autoscaler has the highest pod priority
#### Description

Required to ensure that if a pod gets evicted due to low resource availability on the node, the cluster autoscaler can provision another node for the operator node group. When the node's kubelet notices the low resource availability (memory), it taints the node with the `node.kubernetes.io/memory-pressure:NoSchedule` taint, which will effectively trigger the autoscaler to add another node for the other pending pods.

Whereas if it doesn't have the highest priority, there's a chance of the cluster autoscaler pod to get evicted, which will leave the cluster in a permanently broken state - it is possible that most of the pods won't be able to start on the nodes of the operator node group because they will all be demanding more memory because:

a). They would all use more memory on startup than usual (given the node's already dwindling resources)  
b). The pods may be requiring more memory due to the cluster's big size.

RobertLucian:
Closes #2346

---

checklist:

### DIFF
--- a/manager/manifests/cluster-autoscaler.yaml.j2
+++ b/manager/manifests/cluster-autoscaler.yaml.j2
@@ -182,7 +182,7 @@ spec:
           resources:
             limits:
               cpu: 300m
-              memory: 1Gi
+              memory: 1.5Gi
             requests:
               cpu: 100m
               memory: 200Mi
--- a/manager/manifests/cluster-autoscaler.yaml.j2
+++ b/manager/manifests/cluster-autoscaler.yaml.j2
@@ -176,6 +176,7 @@ spec:
         app: cluster-autoscaler
     spec:
       serviceAccountName: cluster-autoscaler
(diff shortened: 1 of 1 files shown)


deliahu (approved):

manager/manifests/cluster-autoscaler.yaml.j2
```diff
@@ -176,13 +176,14 @@ spec:
...
       serviceAccountName: cluster-autoscaler
+      priorityClassName: system-cluster-critical
       containers:
         - image: {{ config['image_cluster_autoscaler'] }}
           name: cluster-autoscaler
           resources:
             limits:
               cpu: 300m
-              memory: 1Gi
+              memory: 1.5Gi
```
deliahu, line 186: let's remove the limit

PR #2350, opened 2021-07-16
Fix autoscaler behaviour when encountering zero-valued recommendations

miguelvr:
checklist:


pkg/autoscaler/autoscaler.go
```diff
@@ -44,19 +44,17 @@ type Scaler interface {
...
 
 func New(logger *zap.SugaredLogger) *Autoscaler {
 	return &Autoscaler{
-		logger:              logger,
-		crons:               make(map[string]cron.Cron),
-		scalers:             make(map[userconfig.Kind]Scaler),
-		lastAwakenTimestamp: make(map[string]time.Time),
+		logger:  logger,
+		crons:   make(map[string]cron.Cron),
+		scalers: make(map[userconfig.Kind]Scaler),
```
deliahu, line 57: Do we need to make an empty map for `recs` here?

pkg/autoscaler/autoscaler.go
```diff
@@ -240,25 +229,20 @@ func (a *Autoscaler) autoscaleFn(api userconfig.Resource) (func() error, error)
 
 		if request < currentReplicas {
 			downscaleStabilizationFloor = recs.maxSince(autoscalingSpec.DownscaleStabilizationPeriod)
+			if downscaleStabilizationFloor != nil {
+				downscaleStabilizationFloor = pointer.Int32(libmath.MaxInt32(*downscaleStabilizationFloor, 1))
```
deliahu, line 233: It there a reason that this should not be the opposite of the one below? e.g.

```go
downscaleStabilizationFloor = pointer.Int32(libmath.MinInt32(*downscaleStabilizationFloor, currentReplicas))
```

we can discuss this live

pkg/autoscaler/autoscaler.go
```diff
@@ -227,6 +214,8 @@ func (a *Autoscaler) autoscaleFn(api userconfig.Resource) (func() error, error)
 			recommendation = autoscalingSpec.MaxReplicas
 		}
 
+		recs := a.recs[api.Name]
```
deliahu, line 217: nit: Is this line necessary or just for convenience? Also, if just for convenience, would it make sense to move it above this function (e.g. next to `a.recs[api.Name] = newRecommendations()`)?

deliahu (approved):

PR #2356, opened 2021-07-19
Add go cache to circle ci lint and test workflows

miguelvr:


.circleci/config.yml
```diff
@@ -81,6 +87,10 @@ jobs:
       - run:
           name: Lint
           command: make lint
+      - save_cache:
```
RobertLucian, line 90: Do you know if there are any storage limits on this?

RobertLucian (approved):
Nice!

.circleci/config.yml
miguelvr: 500Mb

PR #2366, opened 2021-07-22
RealtimeAPI CRD

miguelvr:
checklist:


pkg/crds/controllers/serverless/realtimeapi_controller_helpers.go
```diff
@@ -0,0 +1,571 @@
...
+
+func (r *RealtimeAPIReconciler) desiredContainers(api serverless.RealtimeAPI) ([]kcore.Container, []kcore.Volume) {
+	containers, volumes := r.userContainers(api)
+	proxyContainer, proxyVolume := r.proxyContainer(api)
+
+	containers = append(containers, proxyContainer)
+	volumes = append(volumes, proxyVolume)
+
+	return containers, volumes
+}
```
RobertLucian, line 315: We should use the `workloads.RealtimeContainers(api)` function instead of the `userContainers` and `proxyContainer`. We don’t need duplicate code. Same code already also applies to the task/batch/async APIs, so we should use the `workloads` package. See how we’re doing it for the batch CRD for instance. You'll probably need a `serverless.RealtimeAPI` to `spec.API` translator. I believe that's how we do it for the batch CRD too.

pkg/crds/controllers/serverless/suite_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Controller Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func() {
```
RobertLucian, line 52: What does this do?

miguelvr: sets up the test environment. It is generated by kubebuilder

RobertLucian:
We should also confirm the following 2 things:

1. That updating the requested replicas for an API (through the tweaking of `min_replicas`/`max_replicas`) has the same behavior as what's currently on master.
1. That re-deploying an API (realtime) that hasn't had its API spec updated doesn't trigger a rolling-update process, as it's normally expected from the refresh command.

pkg/operator/resources/realtimeapi/api.go
```diff
@@ -17,147 +17,157 @@ limitations under the License.
...
+	// the deployment-id and spec-id components of the api-id remain unchanged
+	api.Annotations["cortex.dev/api-id"] = ""
+	_, _, _, apiID := api.GetOrCreateAPIIDs()
+	api.Annotations["cortex.dev/api-id"] = apiID
 
-	initialDeploymentTime, err := k8s.ParseInt64Label(prevVirtualService, "initialDeploymentTime")
+	err = config.K8s.Update(ctx, &api)
 	if err != nil {
-		return "", err
+		return "", errors.Wrap(err, "failed to update realtime api resource")
```
miguelvr, line 139: Why did you change this function? The Refresh operation should not change the api-id. it will change the deployment id and the controller updates the api id

pkg/crds/apis/serverless/v1alpha1/realtimeapi_types.go
```diff
@@ -0,0 +1,283 @@
...
+	buf.WriteString(s.Obj(api.Spec.UpdateStrategy))
+	specID = hash.Bytes(buf.Bytes())[:32]
+
+	apiID = api.Annotations["cortex.dev/api-id"]
+	if apiID == "" ||
+		api.Annotations["cortex.dev/deployment-id"] != deploymentID ||
+		api.Annotations["cortex.dev/spec-id"] != specID {
+		apiID = fmt.Sprintf("%s-%s-%s", spec.MonotonicallyDecreasingID(), deploymentID, specID)
+	}
+	return
```
miguelvr, line 269: Do not do empty returns. We don't use that pattern anywhere else. It is better to have the explicit return, always

miguelvr:
You removed the default annotations I had. Those were intentional and necessary. Please revert

pkg/operator/resources/realtimeapi/api.go
RobertLucian: The problem is that the refresh was not showing up in `cortex get <api-name>` as it did for the realtime api and as it currently does for the async api. It just does a "ghost" rolling update, which is just one part of the picture. We need the new `api-id` to show up in `cortex get` too.

I set the `deployment-id` instead. Also tested it and it works just like async.

Also, why not set the `api-id` here directly (as long as it's only changing the first component of the ID)? We do this too in `UpdateAPI` so what's the difference?

miguelvr: It's in UpdateAPI because it needs to be returned in the response.

For all other cases it should be handled by the controller. The controller should generate a - new api id, so it should (theoretically) work as before. If it is not working it might be a bug that needs to be addressed

RobertLucian: Right, that makes sense. As of this moment, letting the controller update the `api-id` works. It does coincide with the `api-id` of the s3 file (which we generate in the operator), which is great. Wouldn't have wanted to update the resource and then get it back as that's not efficient, plus we know all other details of the spec.

RobertLucian:
@miguelvr the Kubebuilder default annotations were removed for a good reason. With them in there, there was a decent chance that if the user had set a value to 0 (say `min_replicas` for instance, could be any other that allows 0 as a value), then the controller would step in and give it the default value, which is really bad because the operator would no longer be able to download the API spec - they would no longer match because the controller deliberately changed the spec.

And couldn't just double-comment them because they would then appear in the description of each field, which is also bad. They are definitely necessary if we weren't to have the CLI/operator and if the user were to create the RealtimeAPI resource(s) directly, yes. But until then, I'm not seeing a good reason for having them around, not as long as they also break the functionality.

The controller seems to be working well, no issues with it (as of this moment). Where are these annotations required?

miguelvr:
There's no chance the user has to set anything because the operator defaults everything. By removing them, the controller does not work properly without the operator. This is not good for different reasons, but the main one is development. I spent quite a while figuring out the correct defaulting scheme

RobertLucian:
Sure, the operator defaults, but that doesn't matter as long as it still sets the value to zero (whichever field that is). The outcome is that the functionality is broken. It just doesn't work. Tested it and it fails.

> By removing them, the controller does not work properly without the operator.

Where exactly does the controller not work properly? I have tested it and it seems to work well. Could you please provide a granular explanation as to where this is required?

>  I spent quite a while figuring out the correct defaulting scheme

Don't see why they would be required as long as we don't do require the users to create their realtimeapi deployments directly. Could you expand on this one?

miguelvr:
The controller is 100% independent from the "cortex operator" it shouldn't rely on specific defaults to be set. Therefore it needs to work by applying a Kubernetes manifest. This is how the  development of the controller should look like.

If there are no proper defaults set, the controller will try to access stuff that is expected to be set and it will lead to nil pointer dereferences. Tried this and saw it failing.

Whatever you saw failing, it needs to be fixed with the correct annotations, not by removing all the defaulting ones

miguelvr:
In the case of zeros being allowed, then a pointer should be used to avoid the defaulting.

PR #2367, opened 2021-07-23
Restart deployments depending on cluster config during cx configure

vishalbollu:
---

checklist:

### DIFF
--- a/manager/install.sh
+++ b/manager/install.sh
@@ -281,8 +281,8 @@ function start_controller_manager() {
 function restart_controller_manager() {
   echo -n "￮ restarting controller manager "
 
-  kubectl scale deployment operator-controller-manager --replicas=0 - >/dev/null
-  kubectl scale deployment operator-controller-manager --replicas=1 - >/dev/null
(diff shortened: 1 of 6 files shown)


manager/install.sh
```diff
@@ -276,6 +278,15 @@ function start_controller_manager() {
   echo "✓"
 }
 
+function restart_controller_manager() {
+  echo -n "￮ restarting controller manager "
+
+  kubectl scale deployment operator-controller-manager --replicas=0 - >/dev/null
+  kubectl scale deployment operator-controller-manager --replicas=1 - >/dev/null
```
deliahu, line 285: For restarting the operator pod, we delete the pod and let the deployment bring it back:

```bash
kubectl -n=default delete --ignore-not-found=true --grace-period=10 deployment operator >/dev/null 2>&1
```

Is there a reason you went with `kubectl scale`, and if so, should we change that one too?

deliahu (approved):

deliahu (approved):

PR #2376, opened 2021-07-28
Add concurrency to dequeuer

### LINKED ISSUE #2200: Support max_concurrency for async APIs
### Description

Currently, async APIs process one request at a time (i.e. `max_concurrency` is 1). Support the user specifying `max_concurrency` > 1, which will result in concurrent requests into the container.

### Notes

One approach is to have multiple goroutines, all pulling from the SQS queue. Another option is to have a single goroutine pulling more than one message at a time from the queue, and then distributing the messages across goroutines (care will have to be taken to avoid idle goroutines).

### Misc

* Update `async/autoscaling.md` and `async/configuration.md` accordingly (add new sections, update default values for `target_in_flight` and `max_concurrency`, etc)

* Here is the commit which removed `max_concurrency` from the async docs: https://github.com/cortexlabs/cortex/pull/2173/commits/7508c76525fd096ad48ba6fb020b43abd5c6e101

miguelvr:
closes #2200 

---

checklist:


RobertLucian (approved):
LGTM!

pkg/dequeuer/dequeuer.go
```diff
@@ -96,12 +98,37 @@ func (d *SQSDequeuer) ReceiveMessage() (*sqs.Message, error) {
...
+	for i := 0; i < numWorkers; i++ {
+		doneChs[i] = make(chan struct{})
+		go func(i int) {
+			errCh <- d.worker(messageHandler, readinessProbeFunc, doneChs[i])
+		}(i)
+	}
+
+	select {
+	case err := <-errCh:
+		return err
```
deliahu, line 115: Should we send `doneCh <- struct{}{}` for each of the `doneChs` in this case (like below), or is it not necessary since we are returning `err` here which will stop the entire program?

pkg/dequeuer/dequeuer.go
```diff
@@ -96,12 +98,37 @@ func (d *SQSDequeuer) ReceiveMessage() (*sqs.Message, error) {
...
+	case <-d.done:
+		for _, doneCh := range doneChs {
+			doneCh <- struct{}{}
+		}
+	}
+
+	return <-errCh
```
deliahu, line 122: In what case will this be reached, and if it is reached, does it make sense to receive the `errCh`? Is it possible that it will block indefinitely here?

pkg/dequeuer/dequeuer.go
miguelvr: this should be fine

vishalbollu: This abruptly terminates the goroutines but I think that should be fine for now.

If we wanted to handle the error more gracefully, we can use sync.WaitGroup.

deliahu (approved):

PR #2380, opened 2021-07-29
Use a shared async gateway between all async apis

### LINKED ISSUE #2279: Create a shared async gateway deployment which runs on the operator node group

vishalbollu:
closes #2279 

---

checklist:


miguelvr (approved):
LGTM

manager/manifests/async-gateway.yaml.j2
```diff
@@ -0,0 +1,109 @@
...
+            httpGet:
+              path: /healthz
+              port: 8888
+              scheme: HTTP
+          resources:
+            requests:
+              cpu: 100m
+              memory: 200Mi
+            limits:
+              cpu: 100m
```
deliahu, line 73: I think we could make these a little larger, let's discuss live

pkg/async-gateway/service.go
```diff
@@ -78,7 +77,8 @@ func (s *service) CreateWorkload(id string, payload io.Reader, headers http.Head
 	}
 
 	log.Debug("sending message to queue")
-	if err := s.queue.SendMessage(id, id); err != nil {
+	queue := NewSQS(queueURL, &s.session)
```
deliahu, line 80: Is this efficient / ok to do on each request?

deliahu (approved):

pkg/async-gateway/service.go
vishalbollu: Yes, this isn't making any IO calls

PR #2382, opened 2021-08-02
API cost breakdown dashboard

### LINKED ISSUE #1962: Support API Cost Breakdown
### Description

Per the discussion over [here](https://cortex-dot-dev.slack.com/archives/C01KHQRUWNT/p1615560209069100), I would like to be able to understand and calculate the cost of my various API endpoints running within my cortex cluster. 

Right now, using AWS cost tags only gives me the overall cost of the cluster, but not my true Spot, CPU, Memory, or any other costs by replica. Using a formula or estimator is difficult considering all the variables at play.

### Motivation

This has real implication for enabling teams to rapidly prototype new models and move them into production, because you can understand the cost of any individual API endpoint running within the cluster. This might allow you to focus on memory optimization, reduction in init time, speed, or anything else that helps keep your costs down. 

From a contractual R&D perspective, this also helps understand costs specific for customers or programs without spinning up new dedicated clusters just for a single API endpoint or two. While that is practical is some circumstances, it gets unwieldy at large scales and would be easier to track costs based on a single cluster.

### Additional context

Here is an example project that helps track kubernetes costs using Prometheus, it might be a great starting point.

https://github.com/kubecost/cost-model

RobertLucian:
Closes #1962.

---

checklist:

### DIFF
--- a/pkg/operator/operator/cron.go
+++ b/pkg/operator/operator/cron.go
@@ -17,9 +17,11 @@ limitations under the License.
 package operator
 
 import (
+	"context"
 	"strings"
 
 	"github.com/cortexlabs/cortex/pkg/config"
+	"github.com/cortexlabs/cortex/pkg/consts"
 	"github.com/cortexlabs/cortex/pkg/lib/aws"
 	"github.com/cortexlabs/cortex/pkg/lib/errors"
 	"github.com/cortexlabs/cortex/pkg/lib/k8s"
@@ -28,11 +30,19 @@ import (
 	"github.com/cortexlabs/cortex/pkg/lib/sets/strset"
 	"github.com/cortexlabs/cortex/pkg/lib/telemetry"
 	"github.com/cortexlabs/cortex/pkg/types/clusterconfig"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	v1 "k8s.io/api/core/v1"
+	kresource "k8s.io/apimachinery/pkg/api/resource"
 	kmeta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var operatorLogger = logging.GetLogger()
 var previousListOfEvictedPods = strset.New()
+
+var CortexCPUPodReserved = kresource.MustParse("260m")
+var CortexMemPodReserved = kresource.MustParse("350Mi")
 
 func DeleteEvictedPods() error {
 	failedPods, err := config.K8s.ListPods(&kmeta.ListOptions{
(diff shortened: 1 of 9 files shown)


pkg/operator/operator/cron.go
```diff
@@ -215,7 +241,147 @@ func clusterFixedPrice() float64 {
 		natTotalPrice = natUnitPrice * float64(len(config.ClusterConfig.AvailabilityZones))
 	}
 
-	return eksPrice + 2*(operatorInstancePrice+operatorEBSPrice) + metricsEBSPrice + 2*nlbPrice + natTotalPrice
+	fixedCosts := eksPrice + metricsEBSPrice + 2*nlbPrice + natTotalPrice
+	if !includeBaseCortexSystemNodes {
+		return fixedCosts
+	}
+
+	// not entirely correct as there could be more than just one operator instance
```
deliahu, line 249: make a ticket (or fix now) for telemetry and cluster info

pkg/operator/operator/cron.go
```diff
@@ -215,7 +241,147 @@ func clusterFixedPrice() float64 {
...
+			totalCortexSystemCosts += instancePrice
+			continue
+		}
+
+		nodeCPUQty := node.Status.Allocatable[v1.ResourceCPU]
+		nodeMemoryQty := node.Status.Allocatable[v1.ResourceMemory]
+		nodeCPUQty.Sub(CortexCPUPodReserved)
+		nodeMemoryQty.Sub(CortexMemPodReserved)
+		node.Status.Allocatable[v1.ResourceCPU] = nodeCPUQty
+		node.Status.Allocatable[v1.ResourceMemory] = nodeMemoryQty
```
deliahu, line 347: Let's go with a different approach instead of modifying the node's status

pkg/operator/operator/cron.go
```diff
@@ -215,7 +241,147 @@ func clusterFixedPrice() float64 {
...
+			if apiKind, ok := pod.Labels["apiKind"]; ok {
+				if _, okMap := totalWorkloadComputeCostsByAPIName[apiKind]; okMap {
+					totalWorkloadComputeCostsByAPIKind[apiKind] += costPerPod
+				} else {
+					totalWorkloadComputeCostsByAPIKind[apiKind] = costPerPod
+				}
+			}
+		}
+	}
+
```
deliahu, line 373: re-normalize to total cost

pkg/operator/resources/validations.go
```diff
@@ -124,7 +125,8 @@ FluentBit 150
 NodeExporter 200 (it has two containers)
 Reserved (300 + 300 + 200) see generate_eks.py for details
```
deliahu, line 126: move this comment above `CortexMemK8sReserved` (also above)

deliahu (approved):

PR #2398, opened 2021-11-11
Update versions

deliahu:


dev/load.go
```diff
@@ -109,7 +109,7 @@ func runConstantRequestsPerSecond(url string, jsonBytes []byte) {
 	ticker := time.NewTicker(_requestInterval)
 	done := make(chan bool)
 
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 1)
```
RobertLucian, line 112: How does this affect the `load.go` app?

images/manager/Dockerfile
```diff
@@ -15,21 +15,21 @@ RUN pip install --upgrade pip && \
 
 RUN apk add --no-cache bash curl gettext jq openssl
 
-RUN curl --location "https://github.com/weaveworks/eksctl/releases/download/0.51.0/eksctl_$(uname -s)_amd64.tar.gz" | tar xz -C /tmp && \
+RUN curl --location "https://github.com/weaveworks/eksctl/releases/download/v0.67.0/eksctl_$(uname -s)_amd64.tar.gz" | tar xz -C /tmp && \
```
RobertLucian, line 18: To confirm this, they changed `{major.minor.patch}` to `v{major.minor.patch}`, right?

manager/install.sh
```diff
@@ -378,7 +378,7 @@ function remove_nodegroups() {
 
   echo "￮ removing nodegroup(s) from the cluster ..."
   python generate_eks.py $CORTEX_CLUSTER_CONFIG_FILE manifests/ami.json --remove-eks-node-groups="$eks_nodegroup_names" > /workspace/nodegroups.yaml
-  eksctl delete nodegroup --timeout=$EKSCTL_NODEGROUP_TIMEOUT --approve -f /workspace/nodegroups.yaml
+  eksctl delete nodegroup --timeout=$EKSCTL_NODEGROUP_TIMEOUT --drain=false --approve -f /workspace/nodegroups.yaml
```
RobertLucian, line 381: Has this been tested with an API that was running on node group that was about to get removed? I believe the drain is important.

images/manager/Dockerfile
deliahu: Yes that is correct

dev/load.go
deliahu: This should have no effect, but it was necessary to satisfy `go vet` on go 1.17

manager/install.sh
deliahu: Yes, `--drain=false` is preferred. I was having issues with it on `eksctl` version 0.73, but since I've downgraded it to v0.67 it might work now! I'll give it a try

deliahu: Yes it's working as expected now, so I'll revert it, thanks for bringing it up! We'll just have to remember to test this again next time we upgrade `eksctl`

dev/load.go
RobertLucian: Got it.

images/manager/Dockerfile
RobertLucian: Got it.

manager/install.sh
RobertLucian: I see. Thanks for the confirmation on this!

RobertLucian (approved):
LGTM!

PR #2402, opened 2021-12-01
Wait for zero in-flight requests before terminating realtime proxy

deliahu:


miguelvr (commented):
I think we can achieve the same without adding an endpoint. You can simply wait for in-flight requests when the proxy gets the `SIGTERM` signal and before calling `server.Shutdown()`

miguelvr:
> I think we can achieve the same without adding an endpoint. You can simply wait for in-flight requests when the proxy gets the `SIGTERM` signal and before calling `server.Shutdown()`

Ref to the code:
https://github.com/cortexlabs/cortex/blob/16ffd084fc2fe1448a2f2bd6cad14861562b76ee/cmd/proxy/main.go#L172-L189

deliahu:
@miguelvr yeah, that would be cleaner!

I just tried adding it there, but it didn't work as we expected. So I dug into it, and realized that we weren't handling the TERM signal (just INT). After listening for TERM too, checking the breaker's in-flight count was no longer necessary, since `server.Shutdown()` seems to handle this automatically.

What are your thoughts on the diff now?

miguelvr (approved):

tfriedel:
Thanks for the PR! I built the images and tried my test (route which sleeps 1 sec). Contrary to the previous verion, deleting a pod did not result in 503s. Good job!

deliahu:
@tfriedel awesome, thanks for bringing this to our attention, and for confirming that the fix is working as intended!

I'm working on a separate PR now to support custom preStop commands.

RobertLucian (approved):

PR #2443, opened 2022-09-09
Send all VPC CNI logs to /dev/null by default

RobertLucian:
Without this PR, the following lines would constantly get printed on screen when the `cortex` CLI is run:

```
2022-09-09 16:59:13.481063827 +0000 UTC m=+0.039526031 write error: can't make directories for new logfile: mkdir /host: permission denied
2022-09-09 16:59:13.481448123 +0000 UTC m=+0.039910297 write error: can't make directories for new logfile: mkdir /host: permission denied
```

---

checklist:

### DIFF
--- a/python/client/cortex/binary/__init__.py
+++ b/python/client/cortex/binary/__init__.py
@@ -25,7 +25,10 @@ def run():
     Runs the CLI from terminal.
     """
     try:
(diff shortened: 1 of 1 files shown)


jshahbazi:
Just a heads-up: I'm still running into this in v0.42.1 on macOS.  Setting the env variable as per the PR fixes it.

